PR #2, opened 2020-10-22
WIP: Data catalog

tumido:
:boom: :bomb: **This is WIP. Do not merge.** :bomb: :boom:

I've worked on `kustomize`-ing the Data Catalog for Internal DH on: https://github.com/AICoE/idh-manifests/pull/29, https://github.com/AICoE/idh-manifests/pull/27 

This PR is an attempt to convert those manifests to a scheme consumable via `kfdef`. This is a custom, additional, component to ODH and the aim is to have it implemented here and usable for operate-first, before it lands in ODH upstream (that can take time, but we need to have it usable right now.

To-be-done:
- Debug the manifests
- Ensure credentials exchange and connections between Hue, Thriftserver and Hive and their databases
- Assess database unification, or usage of a provided service via operator, instead of a custom DM manifests for `mysql` and `postgresql`.
- Check against https://gitlab.com/opendatahub/opendatahub-operator/-/tree/master/roles/

PR #3, opened 2020-10-23
Docs for overrides

### LINKED ISSUE #1: Update Readme to include Override Procedure
As per the discussions in this issue thread [here](https://github.com/operate-first/continuous-deployment/issues/19) we should document / summarize the important bits of the discussion. Mainly we want to document the override process and include instructions on how one can override changes. We should also link to the examples presented and Vasek's blog post for further reading.

tumido:
Resolves #1 and does a bit of cleanup in the repo.
Related https://github.com/operate-first/continuous-deployment/pull/56

There's no need to repeat the `odh` everywhere:

- Move `odh-operator` to `operator`
- Move `odh-operator-kfdef` to `kfdef` and split per application
- Add a new `components` folder for additional staging ODH components like #2


HumairAK:
So I'm assuming this will follow the following structure as per your comment in chat: 

```
odh
├── components
│   └── data-catalog
├── kfdef
│   ├── jupyterhub
│   │   ├── kfdef.yaml
│   │   └── overrides
│   └── superset
│       ├── kfdef.yaml
│       └── overrides
└── operator

```

My issue with this is that overrides will go under "kfdef", and there is no clear distinction between "components" and the components under kfdef folder. I'm guessing the stuff that goes under "components" will be the full blown manifests to deploy a particular component that is not in odh (yet) ?

README.md
```diff
@@ -1,3 +1,75 @@
...
+        └── kustomization.yaml
+```
+
+### Operator
+
+Folder `operator` hosts manifests for ODH operator deployment. When deployed, it creates a subscription to this operator.
+
+### Kfdef
+
+`kfdef` folder hosts `kfdef.yaml` definitions for our deployments. For easier maintenance we prefer to deploy ODH components into separate namespaces. Each namespace has it's own sub-folder in the `kfdef`.
```
HumairAK, line 37: > Each namespace has it's own sub-folder in the kfdef. 

The wording is a bit confusing here, maybe mention that each override/kfdef component folder should correspond to an individual namespace deployment, atm it seems to imply you have namespaces named "jupyterhub" "superset" etc.

tumido:
@HumairAK what if we change it the other way around: No `components` folder and no `kfdef/<component>/overrides` folder. Just a plain simple `kfdef/<component>/manifests` folder and that's it.

Within each (namespaced) `kfdef` folder we can have a simple manifests folder, which can either hold new "full" manifests or it can contain the overrides. Or it's not present at all if we don't want to override anything.

```
├── kfdef
│   ├── datacatalog
│   │   ├── kfdef.yaml
│   │   ├── kustomization.yaml
│   │   └── manifests
│   │       └── HERE GOES FULL MANIFESTS...
│   ├── jupyterhub
│   │   ├── kfdef.yaml
│   │   ├── kustomization.yaml
│   │   └── manifests
│   │       └── HERE GOES OVERRIDES...
│   └── superset
│       ├── kfdef.yaml
│       └── kustomization.yaml
└── operator
    └── base
        ├── kustomization.yaml
        ├── operator_group.yaml
        └── subscription.yaml
```

README.md
tumido: What about:

```
Each sub-folder in the `kfdef` folder hosts a deployment configuration for a single component or a closely related subset of components. When deployed, each of these configurations (sub-folders) can live in a separate namespace.
```

Is that better?

HumairAK:
Yeah this works for me

martinpovolny:
so the full blown manifests follow the same structure as the customizations?

HumairAK:
Yeah, what if instead of kfdef/jupyterhub/manifests we change it to kfdef/jupyterhub/overrides to further make it clear when we are overriding

martinpovolny:
Actually, I like the proposed structure and do not mind that the 2 are not different.

Just pointing out that this was mentioned in https://github.com/operate-first/odh/pull/3#issuecomment-715328206 and is not addressed in the latest suggestion.

But I do not see it as a flaw.

README.md
HumairAK: yes much better!

martinpovolny (approved):

4n4nd:
We will update the kfdef structure to the following:
```
odh
├ 
│  
├── kfdef
|   ├── kfdef.yaml
│   ├── jupyterhub
│   │   └── overrides
│   └── superset
│       └── overrides
└── operator
```
This is because we are keeping all the odh components in the same namespace for now.

tumido:
@4n4nd we've decided to split the components into different namespaces (see the updates manifests and https://github.com/operate-first/continuous-deployment/pull/56 ). It's like that for easier management, monitoring and debugging. Deploying the full ODK + Kubeflow into a single namespace is unmaintainable and will break through any reasonable resource quota anyways...

4n4nd:
@tumido this was recommended by upstream ODH as it might break some integrations if we divide some of the components into separate namespaces.

tumido:
@4n4nd Thank you for tagging me in that thread, I was not aware of it. Actually I'm reading it differently:

Vasek said:
Sounds good..if you see immediate things to break up, we'd be happy to discuss/help
Marcel replied:
thats the plan

And I agree with it. Let's split things up into manageable chunks, and see what it'll break and work on those pain points.

After all, I don't see any real user, which dedicated a cluster to ODH only (since AI can consume a lot of resources), to spin everything up in a single namespace. That's not a real world usecase.

PR #7, opened 2020-10-27
Update namespaces to use opf- prefix

HumairAK:
The suggestion here is that the default namespaces for components  use the `opf-` prefix with the format opf-< component >


anishasthana (approved):
I like it.

PR #23, opened 2020-11-12
Add opf-monitoring namespace with Prometheus instance

4n4nd:
Will need to add a route for public access
This will read metrics from opf-jupyterhub prometheus using remote read


anishasthana:
@4n4nd can't you add that in the same PR?
also, can you fix the precommit checks?

4n4nd:
> @4n4nd can't you add that in the same PR?

Don't want to make it public rn

4n4nd:
@anishasthana Added the route and made pre-commit happy

HumairAK (approved):

PR #40, opened 2020-11-25
Add superset pvc override

### LINKED ISSUE #35: Superset PVC error on moc
Superset is running into the following error on MOC: 
```
running "VolumeBinding" filter plugin for pod "superset-1-f9zll": pod has unbound immediate PersistentVolumeClaims
```
The PVC shows the following error: 
```
superset-data:
no persistent volumes available for this claim and no storage class is set
```

I'm guessing this is related to #22 - where JH had similar issues. But `superset` component doesn't seem to give the same method to change storage class via a parameter. We should try overriding this for now. If it works, suggest upstream to add a parameter to specify storage class.

HumairAK:
Fixes #35

### DIFF
--- a/kfdef/superset/kfdef.yaml
+++ b/kfdef/superset/kfdef.yaml
@@ -21,7 +21,14 @@ spec:
           name: manifests
           path: superset
       name: superset
+    - kustomizeConfig:
+        repoRef:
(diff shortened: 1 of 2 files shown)


hemajv (approved):
:ship:

PR #51, opened 2020-12-03
[JH] add notebook-images to jh kustomization

4n4nd:
This will also sync all the notebook images along with the jupyterhub kfdef in the [argocd app](https://github.com/operate-first/argocd-apps/blob/main/odh/jupyterhub.yaml)

### DIFF
--- a/odh/base/jupyterhub/kustomization.yaml
+++ b/odh/base/jupyterhub/kustomization.yaml
@@ -4,3 +4,4 @@ kind: Kustomization
 namespace: opf-jupyterhub
 resources:
   - kfdef.yaml
+  - ./notebook-images


4n4nd:
@tumido do you think it's is a good idea to add the notebook-images to this kustomization?

martinpovolny:
I think that it's reasonable to assume that the notebooks are included here.

Actually, I think it should be considered a bug that they are not included.

tumido (approved):

tumido:
I think this PR slipped my radar after the restructure. Yes, it make sense to have them included. They are `ImageStream`s after all, so it's not dependent of ODH per se. I would say we can deploy them outside of ODH (outside of `kfdef`).

tumido:
I need one more github "approve" review (not a chatops - sesheta one) to unblock the merge button. :smile:

durandom:
> I need one more github "approve" review (not a chatops - sesheta one) to unblock the merge button. 😄

Is this a bug or a feature? cc @goern @harshad16 
We require 2 approvals, why does @martinpovolny approve not count towards it?

durandom (approved):

tumido:
@durandom @martinpovolny's review doesn't count because it was not a "GitHub" review (the green tick at the "Reviewers" box), but just a Prow approval. And GitHub is the one enforcing the 2 reviewers rule here.

We have a war of bots starting in here...

harshad16:
> Is this a bug or a feature?
It is not a feature from the prow side, prow just looks for the approved label from an approver in OWNERS file. 
and status checks successes. This should be enough to get the pr merged. 
It would be hard now to trace and see why that happened, but looking from the whole PR discussion it feels like, martin's approval was the one that created the `approved` label. 
prow would be waiting for its pool, as it prow merge based on its pool set. 

> We require 2 approvals, why does @martinpovolny approve not count towards it?
the maintainer of the repo has selected to add branch protection of `2 approvals`, so Github is doing its job of waiting. 
we had a request for this https://github.com/thoth-station/thoth-application/issues/527, unfortunately for others, it didn't stop prow from merging with one approval. https://github.com/AICoE/aicoe-cd/pull/198#issuecomment-742066380
If prow merged stopped due to GitHub block, that is definitely something which we would like to check and verify. As then we would be able to point others, who actually want 2 approvals. 

anyhow as tom explained that prow doesn't have a feature to mark pull requests with `GitHub approved(the green tick)`, so GitHub didn't get the second approved.

Issue #10, opened 2020-10-28
Deploy all components added via kfdef #9  on MOC using same method as the one employed in #60


HumairAK:
Add KfDefs and deploy all components added via kfdef #9 on MOC using same method as the one employed [here](https://github.com/operate-first/continuous-deployment/issues/60)

- [ ] ~~ai-library~~
- [ ] ~~airflow~~
- [ ]  seldon
- [x] datactalog
- [x]  radanalyticsio
- [x] [grafana](https://grafana-route-opf-monitoring.apps.cnv.massopen.cloud/)
- [x] [jupyterhub](https://jupyterhub-opf-jupyterhub.apps.cnv.massopen.cloud/)
- [x]  kafka 
- [x]  [argo](http://argo-portal-opf-argo.apps.cnv.massopen.cloud/)
- [x]  [prometheus](http://prometheus-opf-monitoring.apps.cnv.massopen.cloud/)
- [x]  [superset](http://superset-opf-superset.apps.cnv.massopen.cloud/login/)
- [x] [dashboard](https://odh-dashboard-opf-dashboard.apps.cnv.massopen.cloud/)

goern:
@durandom @HumairAK 

just to clarify: you are deploying https://github.com/opendatahub-io/odh-manifests/tree/v0.8.0 and all the services/features deploys will be supported and accessible (with whatever-SLA)

HumairAK:
that is the end goal yes

HumairAK:
For this issue I'm removing ai-library and airflow as part of tasks as we're have no current use case for it and it's not exactly clear if they are in a usable state. We add more issues for those particular components separately should the need arise in the future.

HumairAK:
I think the apps we have right now are sufficient. If the remaining ones are required, an issue can be made and we'll accommodate.

tumido:
We still have another issue pending for the Kubeflow Pipelines, so in a way we're progressing on this

https://github.com/operate-first/support/issues/6

PR #53, opened 2020-12-04
Added kafka component

HumairAK:
Add Kafka: 

related: [this](https://github.com/operate-first/odh-moc-support/issues/27) and #10 

The override is the same as the one in odh-manifests, but I suspect we want to be able to tune it further, so I include it here to give us that flexibility.


4n4nd (approved):
:+1:

PR #54, opened 2020-12-04
Added structure for topic creation with docs

HumairAK:
Related: [this](https://github.com/operate-first/odh-moc-support/issues/27) and #10

### DIFF
--- a/docs/add_kafka_topics.md
+++ b/docs/add_kafka_topics.md
@@ -1,0 +1,34 @@
+# Add Kafka Topics
+
+To add Kafka topics, find your target environment under `overlays` and create a new `KafkaTopic` resource within the
+`kafka` sub directory. For example if you would like to add your `KafkaTopic` on `MOC` then first create a topic like
+the one below:
+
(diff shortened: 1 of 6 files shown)


docs/add_kafka_topics.md
```diff
@@ -0,0 +1,32 @@
...
+  name: orders
+  labels:
+    strimzi.io/cluster: odh-message-bus
+spec:
+  partitions: 2
+  replicas: 3
+```
+
+The label `strimzi.io/cluster` should have the value `odh-message-bus`. The recommended partition size is `2` and
+replica count is `3`. Pick a suitable name, ensure that it's unique in the `topics` folder.
```
maulikjs, line 22: You might also want to mention here that the replica count should be lower than/equal to the number of Kafka brokers

HumairAK: updated, thanks!

maulikjs:
Looks good now

anishasthana (approved):
:ship:

PR #46, opened 2020-12-03
Restructure to a per-downstream overlays

tumido:
Resolves: https://github.com/operate-first/odh/issues/36

As per offline discussions according to the Apps-of-apps evolution here's yet another restructure of this repo reflecting the WG outcome. 

@HumairAK is already working on a ADR for this, however it's easier to base further changes on the updated structure than redo the work multiple times. It may be a bit eary for the PR now, but here it is anyways.

/cc @HumairAK @anishasthana @durandom

PR #56, opened 2020-12-07
Add JH monitoring back in and fix kustomize build error for jh

HumairAK:
Was accidentally removed in #46 
This also happened to break the `kustomize build` for JH because the overlays index had changed.


4n4nd (approved):
👍

Issue #37, opened 2020-11-25
Add authentication via Openshift to Grafana


HumairAK:
The default access should be read only for users.

HumairAK:
I have a prototype example [here](https://github.com/HumairAK/odh/tree/add_grafana_auth_dex/kfdef/monitoring) that uses dex to add auth to grafana. We can assign grafana roles based on openshift groups using this method. Downside is we have to deploy a separate instance of dex to do this.

anishasthana:
Doesn't really seem like there are any downsides to me. This way our monitoring graphs and stuff can be open for anyone to look at, but only we have edit rights on stuff. An additional dex server isn't _that_ bad

tumido:
TBH, we'll need DEX for Argo as well and maybe for other components. Think, instead of deploying DEX along with every component, we should rather leverage the ArgoCD's DEX or have one common DEX deployment with multiple static clients. WDYT?

https://github.com/operate-first/apps/issues/39

HumairAK:
I have been thinking the exact same thing @tumido  -- we should try to use one instance. ArgoCD has some integrations with dex so let's wait on moving that one to use the external one. But we should start by having grafana/argo and others using this central one then moving argocd there as well.

HumairAK:
I'm guessing we can use dex with superset too.

tumido:
I can take care of it. :+1: I'll file an issue for myself. I have a POC ready for Argo, you have one for Grafana, so I combine it together... WDYT? :slightly_smiling_face:

tumido:
Yeah, we can hook Superset to it and I'm sure Hue as well. It's more user friendly than Openshift itself. :smile:

HumairAK:
Yeah go for it, if you have any questions about the grafana stuff feel free to ask.

HumairAK:
Can we add it under operate-first/apps under a separate folder `dex` that's outside of odh.

HumairAK:
more authorization related, but is relevant I think:

if we can, we should also see about limiting access to editing dashboards. [Discussion here](https://github.com/operate-first/apps/issues/112#issuecomment-762468104)

HumairAK:
PR by @4n4nd to add `role_attribute_strict` attribute to grafana operator: https://github.com/grafana-operator/grafana-operator/pull/499

PR #67, opened 2020-12-08
Add Loki Stack using Observatorium

4n4nd:
The Observatorium CR consists a lot of thanos configuration because it is required, but all of thanos stuff has been scaled down.

### DIFF
--- a/observatorium/grafana-data-sources/kustomization.yaml
+++ b/observatorium/grafana-data-sources/kustomization.yaml
@@ -1,0 +1,5 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - opf-test-loki-source.yaml
--- a/observatorium/grafana.yaml
+++ b/observatorium/grafana.yaml
@@ -1,0 +1,34 @@
+---
+apiVersion: integreatly.org/v1alpha1
+kind: Grafana
+metadata:
+    annotations:
+        argocd.argoproj.io/sync-wave: "1"
+    name: grafana
+spec:
+    baseImage: quay.io/app-sre/grafana:7.1.1
+    config:
+        security:
+            admin_user: "root"
+            admin_password: "secret"
+        users:
+            viewers_can_edit: true
+            allow_sign_up: false
+            auto_assign_org: true
+            auto_assign_org_role: Viewer
+        auth:
+            disable_login_form: true
+        auth.anonymous:
+            enabled: true
+            org_role: Viewer
+        log:
+            level: warn
+            mode: console
+    dashboardLabelSelector:
+    -   matchExpressions:
+        -   key: app
+            operator: In
+            values:
+            - grafana
+    ingress:
+        enabled: true
--- a/observatorium/observatorium.yaml
+++ b/observatorium/observatorium.yaml
@@ -1,0 +1,151 @@
+---
+apiVersion: core.observatorium.io/v1alpha1
+kind: Observatorium
+metadata:
+  labels:
+    app.kubernetes.io/instance: opf-observatorium
+    app.kubernetes.io/name: observatorium-cr
+    app.kubernetes.io/part-of: observatorium
+  name: opf-observatorium
+spec:
+  # Observatorium API is scaled down
+  api:
+    image: quay.io/observatorium/observatorium:master-2020-09-08-v0.1.1-142-g61a908f
+    rbac:
+      roleBindings:
+      - name: test
+        roles:
+        - read-write
+        subjects:
+        - kind: user
+          name: admin@example.com
+      roles:
+      - name: read-write
+        permissions:
+        - read
+        - write
+        resources:
+        - logs
+        - metrics
+        tenants:
+        - test
+    replicas: 0
+    tenants:
+    - id: 1610b0c3-c509-4592-a256-a1871353dbfa
+      name: test
+      oidc:
+        clientID: test
+        clientSecret: ZXhhbXBsZS1hcHAtc2VjcmV0
+        issuerURL: http://dex.dex.svc.cluster.local:5556/dex
+        usernameClaim: email
+    tls:
+      caKey: ca.pem
+      certKey: cert.pem
+      configMapName: observatorium-opf-tls
+      keyKey: key.pem
+      secretName: observatorium-opf-tls
+      serverName: observatorium-opf-observatorium-api.opf-monitoring.svc.cluster.local
+    version: master-2020-09-08-v0.1.1-142-g61a908f
+  apiQuery:
+    replicas: 0
+    image: quay.io/thanos/thanos:master-2020-08-12-70f89d83
+    version: master-2020-08-12-70f89d83
+  compact:
+    enableDownsampling: false
+    image: quay.io/thanos/thanos:master-2020-08-12-70f89d83
+    replicas: 0
+    retentionResolution1h: 1s
+    retentionResolution5m: 1s
+    retentionResolutionRaw: 14d
+    version: master-2020-08-12-70f89d83
+    volumeClaimTemplate:
+      spec:
+        accessModes:
+        - ReadWriteOnce
+        resources:
+          requests:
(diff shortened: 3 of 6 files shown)


observatorium/grafana-data-sources/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - opf-test-loki-source.yaml
```
anishasthana, line 5: Should we remove references to `test` from all the files?

observatorium/grafana.yaml
```diff
@@ -0,0 +1,34 @@
...
+kind: Grafana
+metadata:
+    annotations:
+        argocd.argoproj.io/sync-wave: "1"
+    name: grafana
+spec:
+    baseImage: quay.io/app-sre/grafana:7.1.1
+    config:
+        security:
+            admin_user: "root"
```
anishasthana, line 12: can we move this to a secret? https://github.com/AICoE/internal-data-hub/pull/17

anishasthana (changes requested):

observatorium/grafana-data-sources/kustomization.yaml
4n4nd: This is a test datasource that can be used as an example to create others

observatorium/grafana.yaml
4n4nd: It doesn't really matter since the login page is disabled you cannot log in with any credentials

observatorium/grafana-data-sources/kustomization.yaml
4n4nd: This was copied from https://github.com/operate-first/apps/blob/master/odh/base/monitoring/grafana.yaml

observatorium/grafana.yaml
4n4nd: This was copied from https://github.com/operate-first/apps/blob/master/odh/base/monitoring/grafana.yaml

observatorium/grafana-data-sources/kustomization.yaml
4n4nd: Also this is for the org-id `'opf-test'`

HumairAK:
So the reason for having a separate grafana atm is because the opf-monitoring is behind a proxy and we have a use-case requiring it to be public. This will change later as the intention is to make the grafana deployed in opf-monitoring namespace to be publicly accessible as read-only, when that happens we should get rid of this grafana and use the data source with our one opf-monitoring grafana instance. Since this is a blocker for others, let's get this grafana in for now as a quick fix. 

Track grafana auth work here #37

HumairAK (approved):

observatorium/observatorium.yaml
```diff
@@ -0,0 +1,151 @@
...
+    image: quay.io/thanos/thanos:master-2020-08-12-70f89d83
+    version: master-2020-08-12-70f89d83
+  compact:
+    enableDownsampling: false
+    image: quay.io/thanos/thanos:master-2020-08-12-70f89d83
+    replicas: 0
+    retentionResolution1h: 1s
+    retentionResolution5m: 1s
+    retentionResolutionRaw: 14d
+    version: master-2020-08-12-70f89d83
```
tumido, line 60: The version should be abstracted away. It's tied to an image tag I suppose, so it should be passed from a single variable/annotation whatever from `kustomization.yaml` along with the image name.

observatorium/observatorium.yaml
```diff
@@ -0,0 +1,151 @@
...
+          requests:
+            storage: 250Mi
+  objectStorageConfig:
+    loki:
+      accessKeyIdKey: aws_access_key_id
+      bucketsKey: buckets
+      endpointKey: endpoint
+      regionKey: region
+      secretAccessKeyKey: aws_secret_access_key
+      secretName: loki-objectstorage
```
tumido, line 95: This secret is missing from the manifests completely.

observatorium/observatorium.yaml
```diff
@@ -0,0 +1,151 @@
...
+    thanos:
+      key: thanos.yaml
+      name: thanos-objectstorage
```
tumido, line 98: What does this refer to? A config map? Why it's not part of this PR?

tumido (commented):
I think somebody was too fast to get this merged without a proper review. If this is observatorium `base`, it should be in `base` folder. If it's a POC, it shouldn't be here in production manifests at all and something else, usable, should be written from scratch here.

These manifests are not even deployable.

Can we revisit this PR? @4n4nd @HumairAK @durandom

durandom:
> I think somebody was too fast to get this merged without a proper review. If this is observatorium `base`, it should be in `base` folder. If it's a POC, it shouldn't be here in production manifests at all and something else, usable, should be written from scratch here.
> 
> These manifests are not even deployable.
> 
> Can we revisit this PR? @4n4nd @HumairAK @durandom

Agreed. And hopefully we'll have a proper CI pipeline to test-deploy stuff before it's merged. 

@goern can we configure the bots so, that if @anishasthana requests changes, but @HumairAK approves, it wont be merged? 

@4n4nd @tumido should we "revert" this PR (by clicking the revert button), or should we address the comments in a follow up PR?

@4n4nd also, please get the observatorium folks (Bartek) to review.

durandom:
I think reverting will be cleaner: https://github.com/operate-first/apps/pull/71

observatorium/observatorium.yaml
```diff
@@ -0,0 +1,151 @@
...
+        - ReadWriteOnce
+        resources:
+          requests:
+            storage: 1Gi
+  hashrings:
+  - hashring: default
+    tenants: []
+  # Only Loki Components are scaled up
+  loki:
+    image: quay.io/4n4nd/loki:1.6.1
```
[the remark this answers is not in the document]
4n4nd: I can update that :+1

observatorium/observatorium.yaml
```diff
@@ -0,0 +1,151 @@
...
+        issuerURL: http://dex.dex.svc.cluster.local:5556/dex
+        usernameClaim: email
+    tls:
+      caKey: ca.pem
+      certKey: cert.pem
+      configMapName: observatorium-opf-tls
+      keyKey: key.pem
+      secretName: observatorium-opf-tls
+      serverName: observatorium-opf-observatorium-api.opf-monitoring.svc.cluster.local
+    version: master-2020-09-08-v0.1.1-142-g61a908f
```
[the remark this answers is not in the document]
4n4nd: We are not using  that yet, it has been scaled down. The only reason it is in the template is because these are required values in the Resource.

observatorium/observatorium.yaml
4n4nd: This is a secret which has been deployed manually

observatorium/observatorium.yaml
4n4nd: This is a secret that has been deployed manually

observatorium/observatorium.yaml
4n4nd: these are just dummy values and not actually being used

4n4nd:
> I think somebody was too fast to get this merged without a proper review. If this is observatorium `base`, it should be in `base` folder. If it's a POC, it shouldn't be here in production manifests at all and something else, usable, should be written from scratch here.

Oh my bad, I can move it to a `base` directory

> These manifests are not even deployable.
> 
> Can we revisit this PR? @4n4nd @HumairAK @durandom

Only the [loki items](https://github.com/operate-first/apps/blob/master/observatorium/observatorium.yaml#L71-L99) are actually working for us, everything else has been scaled down (Observatorium and Thanos are using dummy values). We are using the Observatorium Operator to deploy Loki for us, but we are not actually using the Observatorium API to interact with Loki.

tumido:
@4n4nd let's revert it and review this properly.

PR #68, opened 2020-12-08
Add github alertmanager receiver

hemajv:
Adding [GitHub Alertmanager](https://github.com/m-lab/alertmanager-github-receiver) for incident management. The GitHub alertmanager receiver will create new issues in a given repository for triggered Prometheus alerts. See https://github.com/operate-first/SRE/issues/19 for more details.

### DIFF
--- a/odh/base/monitoring/github-receiver-service.yaml
+++ b/odh/base/monitoring/github-receiver-service.yaml
@@ -6,7 +6,6 @@ metadata:
 spec:
   ports:
     - port: 9393
-      protocol: TCP
       targetPort: receiver
   selector:
     # Pods with labels matching this key/value pair will be publically
--- a/odh/base/monitoring/github-receiver-service.yaml
+++ b/odh/base/monitoring/github-receiver-service.yaml
@@ -1,0 +1,15 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: github-receiver-service
+spec:
+  ports:
+    - port: 9393
+      protocol: TCP
+      targetPort: receiver
+  selector:
+    # Pods with labels matching this key/value pair will be publically
+    # accessible through the service IP and port.
+    run: github-receiver
+  type: ClusterIP
--- a/odh/base/monitoring/github-receiver.yaml
+++ b/odh/base/monitoring/github-receiver.yaml
@@ -1,0 +1,37 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: github-receiver
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      run: github-receiver
+  template:
+    metadata:
+      labels:
+        run: github-receiver
+      annotations:
+        prometheus.io/scrape: 'true'
+        prometheus.io/port: '9990'
+    spec:
+      containers:
+        - name: github-receiver
+          image: quay.io/operate-first/alertmanager-github-receiver:v0.10
+          env:
+            - name: GITHUB_AUTH_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: github-secrets
+                  key: auth-token
+          args: ["--authtoken=$(GITHUB_AUTH_TOKEN)",
+                 "--org=operate-first",
+                 "--repo=SRE",
+                 "--label=in progress",
+                 "--enable-inmemory=false"]
+          ports:
+            - containerPort: 9393
+              name: receiver
+            - containerPort: 9990
+              name: metrics
--- a/odh/base/monitoring/kustomization.yaml
+++ b/odh/base/monitoring/kustomization.yaml
@@ -4,8 +4,5 @@ kind: Kustomization
 namespace: opf-monitoring
 resources:
   - kfdef.yaml
-  - prometheus.yaml
-  - prometheus-route.yaml
-  - grafana.yaml
   - github-receiver.yaml
   - github-receiver-service.yaml
--- a/odh/base/monitoring/kustomization.yaml
+++ b/odh/base/monitoring/kustomization.yaml
@@ -4,3 +4,8 @@ kind: Kustomization
 namespace: opf-monitoring
 resources:
   - kfdef.yaml
+  - prometheus.yaml
+  - prometheus-route.yaml
+  - grafana.yaml
+  - github-receiver.yaml
+  - github-receiver-service.yaml
--- a/odh/overlays/crc/monitoring/example-github-secret.yaml
+++ b/odh/overlays/crc/monitoring/example-github-secret.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: github-secrets
+type: Opaque
+data:
+  auth-token: GITHUB-AUTH-TOKEN
--- a/odh/overlays/crc/monitoring/kustomization.yaml
+++ b/odh/overlays/crc/monitoring/kustomization.yaml
@@ -1,6 +1,7 @@
 ---
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-
+namespace: opf-monitoring
 resources:
   - ../../../base/monitoring
+  - example-github-secret.yaml
--- a/odh/overlays/moc/monitoring/.sops.yaml
+++ b/odh/overlays/moc/monitoring/.sops.yaml
@@ -1,0 +1,3 @@
+creation_rules:
+  - encrypted_regex: '^(data|stringData)$'
+    pgp: '0508677DD04952D06A943D5B4DC4116D360E3276'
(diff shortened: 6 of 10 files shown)


odh/base/monitoring/github-receiver.yaml
```diff
@@ -0,0 +1,33 @@
...
+    metadata:
+      labels:
+        run: github-receiver
+      annotations:
+        prometheus.io/scrape: 'true'
+        prometheus.io/port: '9990'
+    spec:
+      containers:
+        - name: github-receiver
+          image: quay.io/4n4nd/alertmanager-github-receiver:v0.10
```
HumairAK, line 21: can we have this added to the operate-first quay repo [here](https://quay.io/organization/operate-first?tab=repos)?

odh/base/monitoring/github-receiver-service.yaml
```diff
@@ -0,0 +1,17 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: github-receiver-service
+  namespace: default
```
HumairAK, line 6: can we remove this namespace -- we overwrite it anyway in the root kustomization

odh/base/monitoring/github-receiver.yaml
hemajv: yes will do :+1:

odh/base/monitoring/github-receiver.yaml
```diff
@@ -0,0 +1,33 @@
...
+          env:
+            - name: GITHUB_AUTH_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: github-secrets
+                  key: auth-token
+          args: ["--authtoken=$(GITHUB_AUTH_TOKEN)", "--org=operate-first", "--repo=SRE", "--label=in progress", "--enable-inmemory=false"]
```
tumido, line 28: Can this be converted to a list in format:
```
- A
- B
- C
```

It's much easier to follow if there are more than few items in array.

odh/base/monitoring/github-receiver-service.yaml
```diff
@@ -0,0 +1,16 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: github-receiver-service
+spec:
+  ports:
+    - port: 9393
+      protocol: TCP
+      targetPort: 9393
```
tumido, line 10: Instead of this, you can refer to the port by its `name` directly (the name, your setting in `containerPorts` in your deployment spec). It's less error prone having the port number and protocol defined only once.

odh/base/monitoring/github-receiver-service.yaml
```diff
@@ -0,0 +1,16 @@
...
+  selector:
+    # Pods with labels matching this key/value pair will be publically
+    # accessible through the service IP and port.
+    run: github-receiver
+  sessionAffinity: None
```
tumido, line 15: This is the default. No need to specify it.

odh/base/monitoring/github-receiver.yaml
```diff
@@ -0,0 +1,33 @@
...
+        prometheus.io/port: '9990'
+    spec:
+      containers:
+        - name: github-receiver
+          image: quay.io/4n4nd/alertmanager-github-receiver:v0.10
+          env:
+            - name: GITHUB_AUTH_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: github-secrets
```
tumido, line 26: Do we have the secret available anywhere? I'm missing it from the manifests here.

tumido (dismissed):

odh/base/monitoring/github-receiver.yaml
hemajv: I haven't added the secret yet. Thought I can update it once we actually start setting up all our Prometheus alerts?

HumairAK: I don't think we can upload this secret, we don't have a secret management solution set up yet -- this needs to definitely be a to-do and come with an issue attached, for now we'll just have to settle with having to manually deploy the secret before deploying it to prod. 

We could just upload the secret encrypted with some gpg key (we haven't made one yet for operate-first, though we should) and not include in the kustomization.

4n4nd: let's use @tumido's suggestion [here](https://github.com/operate-first/apps/pull/72#discussion_r539357449) and add a dummy/stub secret with this and not include it in the kustomization.yaml

HumairAK: I'm fine with that, but we should have an issue attached to this so that we don't forget about it when we go back and add all our secrets.

HumairAK: Let's get a sops yaml and ksops encrypted secret instead as per our offline discussion, but not include it in the kustomization just yet. And link this secret [here](https://github.com/operate-first/apps/issues/77)

hemajv: added the ksops encrypted secret: https://github.com/operate-first/apps/pull/68/commits/f88928116696bd0d12554fe7d10425ef71f7e07e

hemajv:
As discussed, added the ksops encrypted secret for the github (sesheta) auth token: https://github.com/operate-first/apps/commit/f88928116696bd0d12554fe7d10425ef71f7e07e
cc @HumairAK @HumairAK @4n4nd

odh/base/monitoring/github-receiver-service.yaml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: github-receiver-service
+spec:
+  ports:
+    - name: receiver
```
HumairAK, line 8: don't you need to add a port number here?

odh/base/monitoring/kustomization.yaml
```diff
@@ -4,3 +4,8 @@ kind: Kustomization
 namespace: opf-monitoring
 resources:
   - kfdef.yaml
+  - prometheus.yaml
+  - prometheus-route.yaml
+  - grafana.yaml
```
HumairAK, line 9: these 3 are now created via the kfdefs, so they shouldn't be here (the prometheus/grafana instancess, and prometheus route)

odh/base/monitoring/github-receiver-service.yaml
hemajv: As per @tumido suggestion [here](https://github.com/operate-first/apps/pull/68#discussion_r539196386), I'm referring to the port `name` directly (the name specified in `containerPorts` in the deployment spec)

HumairAK:
Secret itself fine to me, I do think it may be worth adding a dummy secret for the other environments (i.e. crc/quicklab), doesn't need to be encrypted, so the deployment itself works.

odh/base/monitoring/github-receiver-service.yaml
HumairAK: He was probably referring to the `targetPort`, the service port itself requires a `port` field, which is the outgoing port. We can wait for @tumido to clarify.

tumido: Yes, that's what I meant - the `targetPort` Sorry for the confusion, I've inlined the comment on the `targetPort` line and it didn't cross my mind that it can be interpreted as for the whole port object. :smile: 

The `port` property is needed in there, `protocol` is TCP by default and `targetPort` should be either nuber but rather the exported pod's `containerPort` port name.

Reference:
https://github.com/kubernetes/kubernetes/blob/release-1.0/docs/user-guide/services.md#defining-a-service

hemajv: @tumido @HumairAK fixed :+1:

4n4nd (approved):
LGTM :+1:

odh/overlays/crc/monitoring/example-github-secret.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
HumairAK, line 1: this secret you can add in the kustomization, since it doesn't require sops, be sure to add the namespace field to the kustomization (otherwise the kustomize build won't add a namespace field to the secret, we'll be doing this to all overlays as part of another issue as well, but you can do it for just the monitoring folder in crc/quicklab, see #99 ) 

you can do the same for the quicklab secret

odh/base/monitoring/github-receiver-service.yaml
```diff
@@ -0,0 +1,15 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: github-receiver-service
+spec:
+  ports:
+    - port: 9393
+      protocol: TCP
```
HumairAK, line 9: TCP is already default, I don't think you need to specify it just the `port` and `targetport` are sufficient

odh/overlays/crc/monitoring/example-github-secret.yaml
hemajv: ack, updated :+1:

HumairAK (approved):
LGTM

PR #71, opened 2020-12-09
Revert "Add Loki Stack using Observatorium"

durandom:
Reverts operate-first/apps#67


durandom:
I'm undecided 🤣 A revert PR also looks ugly

tumido:
I think revert is cleaner.

It's easier to implement it from scratch properly, than correcting all mistakes and possibly missing something. This way we can iterate the review process over a single PR and get something clean, rather than leaving a tech debt in the repo tree and cleaning it up. :slightly_smiling_face:

tumido:
cc @4n4nd
/approve

HumairAK:
This has already happened but I think going forward we should save reverts for changes to services that are deployed live. Since this was an app that is just being onboarded with no one using it, I think it would have been sufficient to just do a follow up PR fixing the changes. Reverts clog up the history and make it difficult to follow what happened, I think we should avoid them whenever possible.

tumido:
From a review and code sanity point of view, I disagree. It's really hard to point out issues in already merged code and have inline discussions on it. You may not catch up everything in a follow up PR. You forget/skip a line here and there and it's not really that helpful... PR reviews are here because we want to maintain some standards and quality, if something slips under our nose, the easiest thing is to roll back to the last know working configuration (to revert the PR merge) and start over.

It's not a good practice to avoid reverts in my eyes. In big repos and projects reverts are totally normal thing to do. I'm speaking from experience, I've got PRs reverted, heck I even reverted my own PRs. It's the same as rollback deployments, etc. You're not trying to fix a deployed image it's not performing. You roll back. If you break your database, you roll back to the previous snapshot... It's not a bad thing by any means.

I don't see a reason why it clogs up the history, can you explain that bit please?

I think you have 2 options here - either let the master branch be broken, until a follow up PR is merged (which can take long time or even multiple PRs) or revert, "rollback" the repo to a maintainable state and work on the changes again as we did this time. It really depends if you want to have master stable or in a broken state all the time. Because once you let the mentality "we've merged something that broke the build, heck we'll fix it next time" sink in, you'd end you in the broken master state for most of the time. 

When you compare the sizes of https://github.com/operate-first/apps/pull/67 and https://github.com/operate-first/apps/pull/72, we've already trimmed away 50 lines of the code, while the new PR also includes the resouces missing from the first iteration. That's **20%** code size reduction while providing better functionality. The code is cleaner, much more maintainable, uses better images and many more... This comparison of the PRs side by side brings clarity and metrics that we're reviewing it properly. You won't ever have that in follow up PRs. You can't measure how much better we've got it this time, unless you really do the in-depth performance analysis or whatever.

A simple revert made the improved code quality really stand out and shine. I consider that as a win.

martinpovolny:
If there's a quick fix immediately available then a follow-up PR that fixes the problem is ok and no reverting is needed.

If the problem is bigger, then a project should try to avoid being in a broken state and revert is the fastest and cleanest option to the last known working state.

So in this particular case, a revert was the best solution in my opinion.

It makes little difference in such a small repo with such a small number of committers. But when things get bigger it becomes a challenge to maintain consistency. Believe me, I know what I am speaking about https://github.com/ManageIQ/manageiq-ui-classic/graphs/contributors ;-)

HumairAK:
Hrmm I think you've made some really solid points @tumido -- especialy being able to do the inline comments. I'm going to go back on my earlier stance and agree with you guys here.

PR #72, opened 2020-12-09
Add the Observatorium CR to deploy Loki 

4n4nd:
Add Grafana deployment using ODH 
Add example Grafana Datasource for Loki 
Trim down the Observatorium CR to just loki and required component values
This disables the Observatorium API and we can interact with Loki directly 
Later we can update it to deploy Observatorium and API and Thanos

### DIFF
--- a/observatorium/base/instance/example-loki-objectstorage-secret.yaml
+++ b/observatorium/base/instance/example-loki-objectstorage-secret.yaml
@@ -1,0 +1,8 @@
+---
+kind: Secret
+apiVersion: v1
+metadata:
+  name: loki-objectstorage
+stringData:
+  endpoint: http://ACCESS_KEY_ID:ACCESS_SECRET_KEY@S3_HOST:S3_PORT/BUCKET_NAME
+type: Opaque
--- a/observatorium/base/instance/grafana-data-sources/kustomization.yaml
+++ b/observatorium/base/instance/grafana-data-sources/kustomization.yaml
@@ -1,0 +1,5 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - opf-test-loki-source.yaml
--- a/observatorium/base/instance/grafana-data-sources/opf-test-loki-source.yaml
+++ b/observatorium/base/instance/grafana-data-sources/opf-test-loki-source.yaml
@@ -1,0 +1,18 @@
+---
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDataSource
+metadata:
+    name: loki-opf-example
+spec:
+    name: loki-opf-example
+    datasources:
+        - name: loki-opf-example
+          type: loki
+          access: proxy
+          url: http://opf-observatorium-loki-query-frontend-http.opf-observatorium.svc.cluster.local:3100
+          version: 1
+          editable: false
+          jsonData:
+              httpHeaderName1: 'X-Scope-OrgID'
+          secureJsonData:
+              httpHeaderValue1: 'opf-example'
--- a/observatorium/base/instance/grafana.yaml
+++ b/observatorium/base/instance/grafana.yaml
@@ -1,0 +1,29 @@
+---
+apiVersion: integreatly.org/v1alpha1
+kind: Grafana
+metadata:
+    name: grafana
+spec:
+    baseImage: quay.io/app-sre/grafana:7.1.1
+    config:
+        users:
+            viewers_can_edit: true
+            allow_sign_up: false
+            auto_assign_org: true
+            auto_assign_org_role: Viewer
+        auth:
+            disable_login_form: true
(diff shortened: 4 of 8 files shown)


4n4nd:
@bwplotka could you please review this?

observatorium/base/observatorium.yaml
```diff
@@ -0,0 +1,97 @@
+---
+apiVersion: core.observatorium.io/v1alpha1
+kind: Observatorium
+metadata:
+  labels:
+    app.kubernetes.io/instance: opf-observatorium
+    app.kubernetes.io/name: observatorium-cr
+    app.kubernetes.io/part-of: observatorium
```
tumido, line 8: What about moving (some of) this to `commonLabels` in `kustomization.yaml`, so it gets applied to the rest of the resources as well.

observatorium/base/observatorium.yaml
```diff
@@ -0,0 +1,97 @@
...
+          requests:
+            storage: 250Mi
+  objectStorageConfig:
+    loki:
+      accessKeyIdKey: aws_access_key_id
+      bucketsKey: buckets
+      endpointKey: endpoint
+      regionKey: region
+      secretAccessKeyKey: aws_secret_access_key
+      secretName: loki-objectstorage
```
tumido, line 35: Can we have a "stub", "dummy" secret as part of the base manifests? With empty values for example, maybe excluded from the `kustomization.yaml` and so on. So we just know what values we have to fill in to make it work, and we don't have to dig into different manifests...

tumido (commented):
hey @4n4nd this is much better! :+1: I like it. Seems most of the issues got fixed. I have just a couple of nitpicks maybe. :wink:

observatorium/base/observatorium.yaml
4n4nd: good idea :+1:

4n4nd: Added an example secret for loki (https://github.com/operate-first/apps/blob/4feded3993df466023a62c232c7f455831efcff6/observatorium/base/example-loki-objectstorage-secret.yaml)

tumido: can you please list the rest of the keys in there as well? The `aws_secret_access_key` `aws_access_key_id` etc? So the picture is complete?

4n4nd: These keys are supposed to work according to the observatorium documentation, but only the `endpoint` works for now.
I have an issue open upstream for it. (https://github.com/observatorium/operator/issues/31)

tumido: let's replace them with stub values in the manifest and link to the issue then:

```
objectStorageConfig:
    loki:
      # FIXME: Most of the following values are not used, yet they are required to be defined 
      # See: https://github.com/observatorium/operator/issues/31
      accessKeyIdKey: ""
      bucketsKey: ""
      endpointKey: endpoint
      regionKey: ""
      secretAccessKeyKey: ""
      secretName: loki-objectstorage
```

or something along those lines. So we don't have to wonder where the values come from and the awareness of such problem is shared and tracked. :wink:

4n4nd: updated :+1:

observatorium/base/observatorium.yaml
```diff
@@ -0,0 +1,98 @@
...
+      roles:
+      - name: dummy
+        permissions:
+        - read
+        - write
+        resources:
+        - logs
+        - metrics
+        tenants:
+        - test
```
tumido, line 59: Just a quick question. Is this really necessary?

Wouldn't something like be sufficient?

```yaml
rbac: {}
tenants: []
```
I haven't tried it out, it's more of a question. We certainly don't want fixed rbac like this in here, so I'm looking for ways how to avoid it.

4n4nd: It is a required parameter in the Observatorium resource

tumido: Ok, so.. I've cloned your PR and deployed it myself. This works just fine, please use this:

```suggestion
    rbac:
      roleBindings: []
      roles: []
    replicas: 0
    tenants: []
```

tumido (changes requested):

tumido:
@4n4nd I think it would be also nice if you'd include the operator spec in here. Because as of now, your manifests are deployable to MOC `prod` only. That doesn't seem like a good practice to me.

https://gitlab.com/open-infrastructure-labs/moc-cnv-sandbox/-/blob/master/manifests/operators/observatorium/kustomization.yaml

Maybe structure it as or something:
```
observatorium
└── base
    ├── instance
    └── operator
```

4n4nd:
@tumido made the changes :+1:

tumido (approved):
@4n4nd I like it a lot! :+1: This is nice and polished. Good job! :tada:

Issue #57, opened 2020-12-07
Test out giving a single prom instance rbac to multi namespaces 


HumairAK:
Since the odh-dashboard requires a single component instance installed via kfdef to pick up it's route, it would be ideal if we could have only one odh deployed prometheus instance via one odh deployed prometheus operator. 

This issue is to test out if we can have service/pod monitors all deployed in a single namespace like `opf-monitoring` then give rbac to a single prometheus instance to other namespaces like `opf-jupyterhub` and `opf-argo` and monitor them from a single source.

4n4nd:
Test results:
1. When the servicemonitors are in the same namespace as prometheus, with the namespace for the target service specified, the prometheus SA just needs to be able to `get/list/watch services`  from the target namespace and it is able to monitor services across namespaces.

2. When the servicemonitors are not in the same namespace as prometheus, then the prometheus needs to be able to `get namespaces` (cluster role) as well as `get/list/watch servicemonitors/podmonitors/services/pods/endpoints` in the target namespaces.
If we apply this role cluster wide, it would look like:
```yaml

---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: operatefirst-monitoring-cluster
rules:
- apiGroups:
  - ""
  resources:
  - namespaces
  verbs:
  - get
- apiGroups:
  - ""
  resources:
  - services
  - endpoints
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - monitoring.coreos.com
  resources:
  - servicemonitors
  - podmonitors
  verbs:
  - get
  - list
  - watch
```
Method 2 would also fix #60, i.e. we would have permissions to access cluster metrics enabling better monitoring for our pods/deployments.

tumido:
So in other words... we found the solution! :slightly_smiling_face: Do I understand it correctly? :tada:  I like the method 2, since it keeps the monitors and services it monitors together.

Does the method 2 also work without a cluster role? I mean, if you use a `Role` and `RoleBinding` in each target namespace giving access to the prometheus SA from the monitoring namespace.

4n4nd:
> So in other words... we found the solution! :slightly_smiling_face: Do I understand it correctly? :tada:  I like the method 2, since it keeps the monitors and services it monitors together.
> 
> 
> 
> Does the method 2 also work without a cluster role? I mean, if you use a `Role` and `RoleBinding` in each target namespace giving access to the prometheus SA from the monitoring namespace.

[...]

PR #78, opened 2020-12-10
Add opf-observatorium loki monitoring

### LINKED ISSUE #75: Add monitoring for the loki deployment
Add service monitor for loki
Add new Prometheus deployment for opf-observatorium namespace

4n4nd:
Add Prometheus and Prometheus Operator deployment  using ODH
Add ServiceMonitor for Loki
Update opf-monitoring Prometheus to monitor opf-observatorium prometheus
Update opf-monitoring metric retention to 7d from default (24H)

Closes #75

### DIFF
--- a/observatorium/base/instance/monitoring/prometheus.yaml
+++ b/observatorium/base/instance/monitoring/prometheus.yaml
@@ -10,6 +10,6 @@ spec:
   serviceAccountName: prometheus-k8s
   securityContext: {}
   serviceMonitorSelector:
-      matchLabels:
-        observatorium-component: monitoring
+    matchLabels:
+      observatorium-component: monitoring
   ruleSelector: {}
--- a/observatorium/base/instance/monitoring/prometheus.yaml
+++ b/observatorium/base/instance/monitoring/prometheus.yaml
@@ -1,0 +1,15 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: Prometheus
+metadata:
+  name: prometheus
+  labels:
+    prometheus: k8s
+spec:
+  replicas: 1
+  serviceAccountName: prometheus-k8s
+  securityContext: {}
+  serviceMonitorSelector:
+      matchLabels:
+        observatorium-component: monitoring
+  ruleSelector: {}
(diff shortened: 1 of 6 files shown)


observatorium/base/instance/monitoring/prometheus.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: prometheus
+  labels:
+    prometheus: k8s
+spec:
+  replicas: 1
+  serviceAccountName: prometheus-k8s
+  securityContext: {}
+  serviceMonitorSelector:
+    matchLabels:
```
HumairAK, line 13: Don't you need to add this label to the service monitor?

HumairAK:
So this looks fine to me, seems to be imitating the setup we use for how we're monitoring JH. 

I wonder how this squares with [this](https://github.com/operate-first/apps/issues/57) investigation. If it's successful then we technically don't need to be adding another prometheus instance here. On the other hand we could just add this for now and remove it later since we'll have to do it for other services like JH anyway. 

Thoughts? @anishasthana  @tumido

HumairAK (approved):

4n4nd:
> On the other hand we could just add this for now and remove it later since we'll have to do it for other services like JH anyway.

I think this is the way to go for now. I am not too optimistic about #57 tbh, but i still need to do more testing.

anishasthana:
I'd like to confirm the results of #57  one way or the other. I'd vote for holding off on merging this until that's determined (sorry Anand :-) ).

tumido:
I agree. I don't like seeing more than one instance of Prometheus and if there's a way to avoid it, I'd try that.

4n4nd:
> I agree. I don't like seeing more than one instance of Prometheus and if there's a way to avoid it, I'd try that.

Actually, I like multiple Prometheus instances. It's like poor man's "sharding". But I guess we are not at a point where we need to worry about scaling Prometheus yet.

tumido:
You can scale and shard a single prometheus just fine. The operator itself supports it. With multiple instances of the same deployment, you would also need to keep them in sync and you need to know from which route to pull the data, compared to a single route. That's a mess.

4n4nd:
> The operator itself supports it

Sure, but it's an experimental feature. 
> With multiple instances of the same deployment

You still get the same issues there, each shard means a separate statefulset.

tumido:
> You still get the same issues there, each shard means a separate statefulset.

Precisely my point why to keep a single instance until we face issues we can't solve with deployment scaling.

PR #80, opened 2020-12-11
fix: Use Grafana/Prometheus as an ODH ocomponent

tumido:
Fixes: #79, #49, https://github.com/operate-first/odh-dashboard/issues/2  #44

Fixes multitude problems with monitoring deployment:

- The prometheus instance was deployed twice - once by us once by ODH https://github.com/operate-first/apps/issues/79
- The `grafana-operator` ODH component was not used, therefore it wasn't showing up on the dashboard #49 https://github.com/operate-first/odh-dashboard/issues/2 
- Since have no control on whey `kfdef` is processed by the operator, we ran into sync issues with Argo CD when deploying CRDs which definitions are deployed via ODH. https://github.com/operate-first/apps/issues/44

This simple reorg of the monitoring manifests layout moves all our custom manifests into ODH overrides

1. Add [`grafana-operator` component](https://github.com/opendatahub-io/odh-manifests/tree/master/grafana) and overrides it's `grafana.yaml` `Grafana` resource with our custom one, instead of deploying it aside of ODH.
2. Override [`prometheus-operator` component](https://github.com/opendatahub-io/odh-manifests/tree/master/prometheus/operator)'s `prometheus.yaml` - the `Prometheus` instance - which solves the 2 Prometheus instances issue.
3. Remove the unnecessary route, since we already had 2 routes pointing to the exact service

Since we would deploy the `grafana-operator` now, we'll also get [all the default dashboards](https://github.com/opendatahub-io/odh-manifests/tree/master/grafana/grafana/base) it comes with it, which should solve https://github.com/operate-first/SRE/issues/11 as well.

Issue #44, opened 2020-12-02
Update monitoring manifests to be deployed in waves


hemajv:
When deploying the monitoring setup in quicklab (via argocd), ran into the following ` Application Sync Error`:

![Screenshot from 2020-12-02 15-35-06](https://user-images.githubusercontent.com/7343099/100929152-2e925900-34b5-11eb-91b2-cb2cf5f2246d.png)

The monitoring manifests should be updated to be deployed in waves: https://argoproj.github.io/argo-cd/user-guide/sync-waves/
so that the kfdef is deployed first, and then followed by the grafana/prometheus custom resources

cc @HumairAK @4n4nd

HumairAK:
Current workaround is to sync the kfdef manually then click the overall sync button.

tumido:
@hemajv waves won't help, since you have no control over when ODH operator picks up the `kfdef` and actually does the deployment. It may be long after the `kfdef` resouce is up in the cluster. We can do this differently via overrides though. I'll solve this as part of https://github.com/operate-first/apps/issues/49

HumairAK:
ah that's a good point, thanks @tumido  !

anishasthana:
Hey folks, @hemajv  was running into some issues with overrides for grafana dashboard deployment that me think a little further.

We probably don't want all dashboards to go into ODH overrides either as there are some cases in which  we will have dashboards that are user specific (say some MOC user wants to visualize time series data). I think these types of dashboards should be deployed  via ArgoCD. So to summarize:

1. Dashboards that are relevant to upstream -> overrides -> Deployed by ODH operator
2. Dashboards that are not relevant to upstream -> Kustomize overlays -> Deployed by ArgoCD

tumido:
we don't have that `GrafanaDashboard` resource kind available until ODH operator deploys it for us. We can't deploy it side by side since it's dependent on the ODH deployment. The solution you're proposing for the 2nd case would reintroduce all the sync issues.

4n4nd:
I like @anishasthana's suggestion. In the 2nd case, if odh has not deployed `GrafanaDashboard` yet, argocd will fail and try again since we have autosync enabled for all our apps. I'd rather have the argocd sync fail couple of times than have the odh operator deploy more resources, this is because the odh operator can be a bit difficult to monitor and with argocd we have more visibility.

tumido:
[...]

PR #85, opened 2020-12-15
Move jh grafana dashboard from /overrides into /base/monitoring

hemajv:
Moving the grafana dashboards into `base/monitoring` so that they can be deployed by argo cd


odh/base/monitoring/kustomization.yaml
```diff
@@ -4,3 +4,4 @@ kind: Kustomization
 namespace: opf-monitoring
 resources:
   - kfdef.yaml
+  - dashboards/jupyterhub-dashboard.yaml
```
HumairAK, line 7: thinking more on it I think it would be better to change this to `dashboards` and add a new kustomization file like so `/dashboards/kustomization.yaml` and include the yaml in there instead. This way when adding a dashboard you only need to make changes to the `dashboard` folder.

hemajv: ya that makes sense, updated it :+1:

HumairAK (approved):

tumido:
@hemajv @HumairAK I don't think you want to move it out of the overrides.

The `GrafanaDashboard` CRD is deployed by ODH operator, you can't make Argo CD to deploy the resources itself before that.

The file moved in this PR introduces a regression - this issue is re-introduced:  https://github.com/operate-first/apps/issues/44

tumido:
@hemajv Also, if you need to add additional resources to the ODH-dependent deployments please do it rather via additional ODH overlays than via Argo CD deployed resources, because that would again lead to the sync issues.

https://github.com/operate-first/apps/pull/80#issuecomment-743239132

HumairAK:
Hold on this, let's come to a conclusion [here](https://github.com/operate-first/apps/issues/87) -- I'm not convinced we should be forced into always using odh to deploy things.

Issue #87, opened 2020-12-16
Override Vs Overlay and CRD rejection issue


HumairAK:
Related Issues: 
- #44 
- #85 

Recently we've been coming across situations where we'd prefer to overlay instead of override resources. The difference being, when we overlay, we are having argocd deploy them, whereas  with overriding in ODH, we have ODH deploy them. 

Context: 
So in some cases ODH deploys operators, which includes CRDs. Let's take for example KafkaTopics. Say I have ArgoCD deploy the KafkaTopics and a KFDef that deploys strimzi. Great, say ArgoCD tries to deploy `KafkaTopics` first, well it can't because it needs to deploy the KFDef first. So the application will fail. 

We can use ArgoCD `waves` to tell ArgoCD to deploy our kfdef (and thus Strimzi) first, then deploy `KafkaTopics` second. This too will fail because the ODH operator needs to read the KFDEF and deploy the CRDs itself. So we need ODH to deploy to the `KafkaTopics` by using overrides if we want to prevent errors from showing up. \

Problem:
I personally don't know if ODH is at a point where we can confidently override certain resources and expect to have them be sync'ed immediately. I have personally encountered situations where ODH will take a while to update the resources, sometimes having to restart ODH itself. I would be interested to hear what your own experiences have been.  

Another situation is, by being forced into override, we are forced into following the folder structure of the `odh-manifests` repo. For example if I want to override grafana dashboards, then I need to update this file [here](https://github.com/opendatahub-io/odh-manifests/blob/master/grafana/grafana/base/datasource.yaml). But what if I want to add another dashboard? The way I understand it  I would need to override the [kustomization here](https://github.com/opendatahub-io/odh-manifests/blob/master/grafana/grafana/base/kustomization.yaml). Okay, but then I have a kustomization, that looks like itt's pulling some files from this repo, and some files from another repo (and the other repo's existence isn't even clear right away by looking at this directory). It just seems messy to me. 

Yet another issue is that we cannot see kfdef child resources any more, see [here](https://github.com/kubeflow/kfctl/issues/460) for more details. So we have no sync/visual control over these resources via argocd.

HumairAK:
Also note that this is only a problem for first deploys, once the CRDs are available, then using overlays doesn't result in any errors/issues. 

[...]

PR #93, opened 2020-12-18
Move topics into overrides and update docs

### LINKED ISSUE #90: Update directory address in kafka topic instructions
The instructions to [create new Kafka topics](https://github.com/operate-first/apps/blob/master/docs/odh/kafka/add_kafka_topics.md) need to be updated after the recent directory restructure.

HumairAK:
Also fixes #90 
Related #87 

Also added some example topics for crc, and quicklab deployments so the topics folder isn't sitting empty.


anishasthana (approved):
No complaints here.

odh/overlays/moc/kafka/kustomization.yaml
```diff
@@ -4,17 +4,18 @@ kind: Kustomization
 
 resources:
   - ../../../base/kafka
-  - topics
 
 patchesJson6902:
   - patch: |
```
4n4nd, line 9: @HumairAK This patch only adds the topics overlay. Do we need something that also deploys the base? 
i.e. the [kafka cluster](https://github.com/operate-first/apps/blob/7f876c9a36e24c8c0502ecebbf6f2a7438a8d48b/odh/overlays/moc/kafka/overrides/kafka/base/kafka-cluster.yaml)?

4n4nd: I think same would go for other environments as well

4n4nd: nvm it deploys the base as well. Just tested :+1:

4n4nd (approved):
LGTM :+1:

PR #94, opened 2020-12-22
Add grafana documentation

### LINKED ISSUE #92: Documentation for instructions on adding new grafana dashboards
Based on our discussion in #87, we should add documentation for adding new grafana dashboards
(also related to https://github.com/operate-first/SRE/issues/5)

hemajv:
Including documentation on how to add grafana dashboards

This closes #92.

### DIFF
--- a/docs/odh/grafana/README.md
+++ b/docs/odh/grafana/README.md
@@ -1,0 +1,25 @@
+## Grafana Monitoring Dashboards
+
+[Grafana](https://grafana.com/oss/grafana) is a multi-platform open source analytics and interactive visualization tool. It allows you to query, create dashboards, explore, and alert on metrics when connected to supported data sources.
+
+For various Operate First managed services and applications, we have created example Grafana dashboards which can be used to visualize our [SLI/SLOs](https://github.com/operate-first/SRE/tree/master/sli-slo) defined for each of the applications.
+
+These monitoring dashboards are useful for visualizing the metrics (such as Prometheus metrics) collected from the applications to track and analyze the system's overall health.
+The dashboards can be altered, saved and reused as per your specific monitoring needs.
+
+Dashboards can be represented by the `GrafanaDashboard` custom resource or by their raw `JSON` files. You can find example monitoring dashboards defined in separate `JSON` files for each of the Operate First managed applications in our [SRE repo](https://github.com/operate-first/SRE/tree/master/dashboards). For more information on how the Grafana dashboards are defined, please refer to [this documentation](https://github.com/integr8ly/grafana-operator/blob/master/documentation/dashboards.md).
+
+## Adding the Dashboards
+
+Grafana can be accessed at: https://grafana-route-opf-monitoring.apps.cnv.massopen.cloud/.
+
+To add your dashboards to Grafana, you will need to create a `GrafanaDashboard` custom resource. To do this, you can follow the instructions available [here](./add_grafana_dashboard.md).
+
+## Using the Dashboards
+
+Grafana dashboards can easily be exported and imported, either from the UI or from the HTTP API.
+Dashboards are exported in Grafana JSON format, and contain everything you need (layout, variables, styles, data sources, queries, etc) to import the dashboard at a later time.
+
+You can import these dashboards in Grafana either by pasting the dashboard JSON text directly into the text area or directly uploading the JSON file. Make sure to add/connect your [data sources](https://grafana.com/docs/grafana/latest/datasources/) to Grafana and pick what data source you want the dashboard to use.
+
+For more details on importing/exporting dashboards, you can refer to the Grafana documentation [here](https://grafana.com/docs/grafana/latest/dashboards/export-import/).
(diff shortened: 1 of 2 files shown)


docs/odh/grafana/README.md
```diff
@@ -0,0 +1,25 @@
+## Grafana Monitoring Dashboards
+
+[Grafana](https://grafana.com/oss/grafana) is a multi-platform open source analytics and interactive visualization tool. It allows you to query, create dashboards, explore, and alert on metrics when connected to supported data sources.
+
+For each of the Operate First managed services and applications, we have created example Grafana dashboards which can be used to visualize our [SLI/SLOs](https://github.com/operate-first/SRE/tree/master/sli-slo) defined for each of the applications.
```
HumairAK, line 5: I would rephrase this to say: 
"For various Operate First managed services and applications..."

The distinction being, that in the original version, we're saying we already have dashboards for _all_ the operate-first managed services, which is not the case. Hope that makes sense.

docs/odh/grafana/add_grafana_dashboard.md
```diff
@@ -0,0 +1,32 @@
+# Add a grafana dashboard to OperateFirst Grafana
+
+To be able to include your dashboards in Grafana, you will need to add a `GrafanaDashboard` custom resource. This custom resource is defined as a yaml file that downloads the dashboard contents from a specified URL and falls back to embedded json if the URL can not be resolved.
```
HumairAK, line 3: Replace: 
> This custom resource is defined as a yaml file that downloads the dashboard contents 

with: 
> This custom resource is defined as a yaml file that can remotely reference the dashboard contents

docs/odh/grafana/add_grafana_dashboard.md
```diff
@@ -0,0 +1,32 @@
...
+  name: grafana-dashboard-from-url
+  labels:
+    app: grafana
+spec:
+  name: grafana-dashboard-from-url.json
+  url: https://raw.githubusercontent.com/integr8ly/grafana-operator/master/deploy/examples/remote/grafana-dashboard.json
+  json:
+```
+
+In this template, all you need to do is update the dashboard name, the dashboard JSON file name and the URL pointing to the location of your dashboard JSON file.
```
HumairAK, line 21: I would be a bit verbose here, and  also list the actual yaml paths, e.g. instead of just "dashboard name" also say "spec.name".

docs/odh/grafana/add_grafana_dashboard.md
```diff
@@ -0,0 +1,32 @@
...
+
+Pick a suitable name, ensure that its unique in the [`odh/base/monitoring/overrides/grafana-operator/overlays/dashboards`](https://github.com/operate-first/apps/tree/master/odh/base/monitoring/overrides/grafana-operator/overlays/dashboards) folder.
```
HumairAK, line 23: Suitable name, as in `metadata.name` ? let's specify that as well.

HumairAK:
Other than the above comments, this lgtm.

docs/odh/grafana/README.md
hemajv: yeah that makes sense, updated it :+1:

docs/odh/grafana/add_grafana_dashboard.md
hemajv: fixed :+1:

docs/odh/grafana/add_grafana_dashboard.md
hemajv: yes I was actually thinking the same too. modified it

docs/odh/grafana/add_grafana_dashboard.md
```diff
@@ -0,0 +1,37 @@
...
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata:
+  name: grafana-dashboard-from-url
+  labels:
+    app: grafana
+spec:
+  name: grafana-dashboard-from-url.json
+  url: https://raw.githubusercontent.com/integr8ly/grafana-operator/master/deploy/examples/remote/grafana-dashboard.json
+  json:
```
tumido, line 18: If both the `url` and `json` properties are specified, the `json` value is used only if the `url` is unresolvable (it's a fallback value). I think it's not necessary to keep it here, I would rather see the operator fail to load the dashboard (so we can track it and alert on it) than fallback to an "empty" one, wdyt?

reference:
- https://github.com/integr8ly/grafana-operator/blob/master/documentation/dashboards.md
- https://github.com/integr8ly/grafana-operator/issues/45#issuecomment-529467580

docs/odh/grafana/add_grafana_dashboard.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+In this template, all you need to do is update:
+1. `metadata.name` - the dashboard name
+2. `spec.name` - the dashboard JSON file name
+3. `spec.url` - the URL pointing to the location of your dashboard JSON file
+
+***Note:** Ensure that the `spec.json` field is empty if setting the URL, if both url and json are specified then the json field won't be updated when fetched.*
```
tumido, line 26: See above. I would avoid specifying both `url` and `json` fields at the same time. Let's either inline it (`json`) or have an `url` without a "magic" empty `json` field fallback.

tumido (commented):
I would change the `json`/`url` fields handling in the `GrafanaDashboard` a bit. WDYT?

docs/odh/grafana/add_grafana_dashboard.md
hemajv: @tumido good poin! I think it would be nice if we can alert on it as well, will retain only the `url` field then

tumido (approved):
LGTM! :+1: Nice piece of documentation! :rocket:

HumairAK (approved):

PR #98, opened 2021-01-07
Add operate-first project admin rbac to as components

### LINKED ISSUE #97: Create MOC rolebindings for component namespaces
We need to create MOC rolebindings for the operatefirst declaratively for each component / namespace, they should go here: 
https://github.com/operate-first/apps/tree/master/odh/overlays/moc

HumairAK:
Fixes #97 

I'm using components here so I don't have to copy paste the same rolebinding everywhere. However, I do copy/paste it once for observatorium and odh, as they seem like different entities, open to changing this however and having a broader shared components folder at the root.

### DIFF
--- a/observatorium/components/rbac/kustomization.yaml
+++ b/observatorium/components/rbac/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+resources:
+  - rolebinding.yaml
--- a/observatorium/components/rbac/rolebinding.yaml
+++ b/observatorium/components/rbac/rolebinding.yaml
@@ -1,0 +1,12 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: admin-rolebinding
+subjects:
+  - kind: Group
+    apiGroup: rbac.authorization.k8s.io
+    name: operate-first
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: admin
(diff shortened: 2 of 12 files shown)


HumairAK:
fyi, you'll need kustomize version 3.7.0 for components support, otherwise your kustomize build will fail, I've already confirmed our current argocd instance is running 3.8+

4n4nd (approved):
LGTM :+1:

4n4nd:
@HumairAK should we remove the [namespaces](https://github.com/operate-first/apps/blob/master/odh/base/jupyterhub/kustomization.yaml#L4) in a separate PR then?

HumairAK:
> @HumairAK should we remove the [namespaces](https://github.com/operate-first/apps/blob/master/odh/base/jupyterhub/kustomization.yaml#L4) in a separate PR then?

well I need them here so that the rolebinding adopt the namespace, so it needs to be a part of this pr

4n4nd:
> @HumairAK should we remove the [namespaces](https://github.com/operate-first/apps/blob/master/odh/base/jupyterhub/kustomization.yaml#L4) in a separate PR then?

oh my bad I should make this more clear. I meant rn we have the namespaces specified in 2 different places. maybe we should just remove them from the base and add them to all the overlays.

tumido (approved):
LGTM.

For the namespaces: I would say let's remove the `namespace` declaration from the base and keep them in overlays only, so we're not redefining them. It sounds like the better option out of 2 bad ones. :smile:

HumairAK:
/unhold 

I'll make a follow up issue for for namespace removal from base

PR #103, opened 2021-01-11
Add prometheus availability alerts

### LINKED ISSUE #101: [Monitoring][Alerts] Add basic `up` alerts for operate-first services
Let's add basic alerts that notify us when any of our monitored services go down.
This will also work as an example alert.

hemajv:
Adding basic availability alerts for the operate first services (currently, we have the `up` metric available only for the prometheus service)
This closes #101

### DIFF
--- a/odh/base/monitoring/overrides/prometheus-operator/base/prometheus-rules.yaml
+++ b/odh/base/monitoring/overrides/prometheus-operator/base/prometheus-rules.yaml
@@ -1,0 +1,24 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: odh-monitoring
+    role: alert-rules
+  name: prometheus-alerting-rules
+spec:
+  groups:
+    - name: Operate First Availability
+      rules:
+        - alert: Prometheus Service Down
+          expr: up{service="prometheus-operated"} == 0
+          for: 5m
+          annotations:
+            summary: "Prometheus service is no longer available"
+            severity: "HIGH"
+        - alert: Prometheus Alertmanager Down
+          expr: up{service="alert-manager"} == 0
+          for: 5m
+          annotations:
+            summary: "Prometheus Alertmanager is no longer available"
+            severity: "HIGH"
--- a/odh/base/monitoring/overrides/prometheus-operator/base/prometheus.yaml
+++ b/odh/base/monitoring/overrides/prometheus-operator/base/prometheus.yaml
@@ -17,7 +17,10 @@ spec:
   serviceAccountName: prometheus-k8s
   serviceMonitorSelector: {}
   securityContext: {}
-  ruleSelector: {}
+  ruleSelector:
+    matchLabels:
+      prometheus: odh-monitoring
+      role: alert-rules
   replicas: 2
   remoteRead:
     - url: "http://prometheus-operated.opf-jupyterhub.svc.cluster.local:9090\


odh/base/monitoring/overrides/prometheus-operator/base/prometheus-rules.yaml
```diff
@@ -0,0 +1,24 @@
...
+  groups:
+    - name: Operate First Availability
+      rules:
+        - alert: Prometheus Service Down
+          expr: up{service="prometheus-operated"} == 0
+          for: 5m
+          annotations:
+            summary: "Prometheus service is no longer available"
+            severity: "HIGH"
+        - alert: Prometheus Alertmanager Down
```
4n4nd, line 19: Can we add an alert for the Jupyterhub service as well?

4n4nd (changes requested):
let's add JH service as well

odh/base/monitoring/overrides/prometheus-operator/base/prometheus-rules.yaml
```diff
@@ -0,0 +1,24 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: k8s
+    role: alert-rules
```
tumido, line 7: I know these labels maps to the `ruleSelector` in `kind: Prometheus`.

Is there any significance for the labels used? I'm not sure I understand "why" these labels and not "others". Why there are 2 different labels? That's just a question out of curiosity, not a critique. Maybe we can think of something more fitting to our usecase than these...

odh/base/monitoring/overrides/prometheus-operator/base/prometheus-rules.yaml
```diff
@@ -0,0 +1,24 @@
...
+  name: odh-monitoring
+spec:
+  groups:
+    - name: Operate First Availability
+      rules:
+        - alert: Prometheus Service Down
+          expr: up{service="prometheus-operated"} == 0
+          for: 5m
```
tumido, line 15: Is `5m` just a safe value for now? I think we should be able to react/alert faster.

odh/base/monitoring/overrides/prometheus-operator/base/prometheus-rules.yaml
tumido: I think we should first verify the workflow just for Prometheus and then we can start onboarding apps. JupyterHub sounds like a logical next step to me.

odh/base/monitoring/overrides/prometheus-operator/base/prometheus-rules.yaml
```diff
@@ -0,0 +1,24 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: k8s
```
anishasthana, line 6: I feel like we shouldn't have this label on here. Maybe better/more descriptive to use something like `prometheus: operate-first` or even just rely on `name: odh-monitoring`

odh/base/monitoring/overrides/prometheus-operator/base/prometheus-rules.yaml
anishasthana: Generally network blips mean things less than 5 minutes are fairly noisy. I think it depends on the MOC cluster though :-)

anishasthana (changes requested):

odh/base/monitoring/overrides/prometheus-operator/base/prometheus-rules.yaml
tumido: Yeah, I was thinking about something like that. :slightly_smiling_face:

odh/base/monitoring/overrides/prometheus-operator/base/prometheus-rules.yaml
hemajv: yes, I wanted to first confirm our setup with prometheus services and then continue to add more

tumido: I think alert manager is part of the workflow itself so it make sense to me to keep it as you have it now. :+1:

odh/base/monitoring/overrides/prometheus-operator/base/prometheus-rules.yaml
hemajv: Yeah, I wasn't sure exactly the right time interval suitable for the MOC cluster so as of now set it to `5m`. We can always revisit and modify this once we actually start observing the alerts, wdyt?

anishasthana: I'm good with that

odh/base/monitoring/overrides/prometheus-operator/base/prometheus-rules.yaml
hemajv: So I just followed the recommended best practice based on the documentation [here](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/user-guides/alerting.md#fire-alerts). According to which it mentions:

```
Use the label selector field ruleSelector in the Prometheus object to define the rule files that you want to be mounted into Prometheus.

The best practice is to label the PrometheusRules containing rule files with role: alert-rules as well as the name of the Prometheus object, prometheus: example in this case.
```

anishasthana: What is the deployed Prometheus called?

hemajv: so [this](http://prometheus-portal-opf-monitoring.apps.cnv.massopen.cloud/) is our Prometheus, and we have defined the name as `odh-monitoring`: https://github.com/operate-first/apps/blob/master/odh/base/monitoring/overrides/prometheus-operator/base/prometheus.yaml#L7

anishasthana: I think lets go with `prometheus: odh-monitoring`

tumido: > retain it as prometheus: odh-monitoring

:+1:  to that. I like that

hemajv: updated :+1:

odh/base/monitoring/overrides/prometheus-operator/base/prometheus-rules.yaml
hemajv: updated :+1:

anishasthana (approved):

tumido (approved):

4n4nd (approved):
Looks good!

PR #104, opened 2021-01-12
feat: Introduce declarative namespace and ClusterRole RBAC management

tumido:
Resolves: https://github.com/open-infrastructure-labs/moc-cnv-sandbox/issues/8, https://github.com/open-infrastructure-labs/moc-cnv-sandbox/issues/7, https://github.com/open-infrastructure-labs/ops-issues/issues/1

Introduces a new Argo CD application that would take care of deploying namespaces and the `admin` `ClusterRole` bindings for them, as well as some generic user management via `Group`s.

Adds admin permissions to the `operate-first` group as well as creates 2 new namespaces:
- `mesh-for-data` (https://github.com/open-infrastructure-labs/moc-cnv-sandbox/issues/8) with a `mesh-for-data` user group
- `open-aiops` (https://github.com/open-infrastructure-labs/moc-cnv-sandbox/issues/7) with `open-aiops` user group

The user stories/request tickets are linked as a label in the `Namespace` resource as well.

### DIFF
--- a/namespaces/overlays/moc/kustomization.yaml
+++ b/namespaces/overlays/moc/kustomization.yaml
@@ -1,0 +1,17 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - mesh-for-data
+  - observatorium-operator
+  - odh-operator
+  - open-aiops
+  - opf-argo
+  - opf-dashboard
+  - opf-datacatalog
+  - opf-jupyterhub
+  - opf-kafka
+  - opf-monitoring
+  - opf-observatorium
+  - opf-superset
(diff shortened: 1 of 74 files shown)


namespaces/overlays/moc/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - mesh-for-data
```
HumairAK, line 6: since we only include it for moc, should we just move the base files here, and leave it out of base? wdyt?

tumido: Sure, will do

HumairAK: second thought, I think this way maybe better, it keeps similar files closer together

tumido: It make sense to me, I've moved it to the overlay. :+1:

HumairAK (approved):

HumairAK:
LGTM, thanks @tumido

4n4nd (approved):
:+1: let's discuss later in the adr

PR #105, opened 2021-01-12
Add argocd admins

HumairAK:
I've intentionally left out the group/namespace for now. I don't have permissions to see the namespace so I can't see what the current annotations (and other fields) are there just yet, once I have permissions to do that I will update with the namespace. 

I've left the group out because, I'd like argocd to create these permissions first so I can add secret management capabilities, as it stands argocd won't be able to decrypt this group.

[Related](https://github.com/open-infrastructure-labs/moc-cnv-sandbox/issues/1)

### DIFF
--- a/namespaces/overlays/moc/argocd/kustomization.yaml
+++ b/namespaces/overlays/moc/argocd/kustomization.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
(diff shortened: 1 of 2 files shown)


anishasthana (approved):

4n4nd (approved):

HumairAK:
FYI this gives the openshift groups `argocd-admins` access to the namespace where argocd is installed

tumido (approved):

PR #106, opened 2021-01-12
fix: move issue url from labels to annotations

tumido:
Labels are limited in length on values. Annotations are not. Would this work, @HumairAK ?


HumairAK (approved):
LGTM

4n4nd (approved):

PR #113, opened 2021-01-13
Remove oauth proxy from grafana

### LINKED ISSUE #112: [Monitoring] Grafana proxy breaks communication with the Grafana Operator
Because of the grafana proxy, the operator is not able to communicate with the Grafana deployment and as a result, it cannot update dashboards as well as datasources. 
I think we were planning to keep grafana open and remove the proxy altogether, which would fix this issue.

4n4nd:
fix JH grafana dashboard
Closes #112


hemajv (approved):
lgtm :ship:

anishasthana (approved):

Issue #115, opened 2021-01-15
Groups should not be in the namespace folder


HumairAK:
I think we should probably pull the groups out of the namespace folder, since it's not a 1:1 mapping between groups and namespaces, rather, one group can be owners of many namespaces, wdyt?

tumido:
Yeah, I was thinking about something like that as well. I think we should reconsider the dir structure to accommodate this a bit better. We should also move the rbac to components etc. Let's discuss this in the ADR, I'll have that PR out today. For now the current layout is sufficient, there's no risk in keeping it as is for few days. Let's finalize the layout in the ADR and base the onboarding docs on that.

Related: https://github.com/operate-first/odh-moc-support/issues/39

anishasthana:
I think discussing in the ADR makes sense.

HumairAK:
Fixed by #120

PR #119, opened 2021-01-18
request for namespace required for continuous integration components

harshad16:
namespace requirements for continuous integration components
Signed-off-by: Harshad Reddy Nalla <hnalla@redhat.com>

Related-to: #118

PR #118, opened 2021-01-18
include thoth components namespace requirements

harshad16:
include thoth components namespace requirements
Signed-off-by: Harshad Reddy Nalla <hnalla@redhat.com>

what should be done regarding the groups?
should it be initialized in all namespaces?

PR #120, opened 2021-01-18
feat: Reorganize cluster-scope app according to the ADR

tumido:
Signed-off-by: Tomas Coufal <tcoufal@redhat.com>

Implements: https://github.com/operate-first/blueprint/pull/19
Resolves: https://github.com/open-infrastructure-labs/ops-issues/issues/2
Adds declarative tracking to: https://github.com/operate-first/support/issues/29

I'm moving around too many files in here, please review fully the "new" layout and don't pay attention to what changed and from where it was moved - git doesn't get it always right.

All `Groups` that are live on the MOC cluster are now synced down from there into this PR.

**This PR is not to be merged now** - it should serve as a base for ADR discussion so everybody can understand how the  proposed solution actually looks and feels like.


tumido:
Precommit will fail, but the HEREDOC in the bash script require tabs instead of spaces... because bash.

HumairAK:
Also fixes #115

cluster-scope/base/groups/groups.yaml
```diff
@@ -0,0 +1,54 @@
+---
```
HumairAK, line 1: Given how the groups file is organized aren't we basically saying that if you import this groups folder, then you _have_ have to  implement all these groups regardless of cluster?

tumido: Yes we are saying that... Kinda. Would you like me to change that? It's a bit of inconsistency, right? Yeah, let's change it. :thinking:

tumido:
Needs rebase/apply of https://github.com/operate-first/apps/pull/118 and https://github.com/operate-first/apps/pull/119

cluster-scope/base/groups/groups.yaml
tumido: I think removing the possibility to include "all" of a certain kind to a specific environment would also make it less error prone. Good call. Changed.

tumido:
Rebased. Changes fro #118 and #119 included.

cluster-scope/base/groups/cluster-admins/group.yaml
```diff
@@ -0,0 +1,6 @@
+---
+apiVersion: user.openshift.io/v1
+kind: Group
+metadata:
+  name: cluster-admins
```
4n4nd, line 5: should we change this to `opf-cluster-admins` to avoid any possible future conflicts?

tumido: I think this is generic. This group comes from MOC implementation actually and clones the live resource. It also includes the cluster admins from BU for the MOC cluster. I agree with the possible conflict, but we aim to have the whole cluster ArgoCD managed - therefore we should not be afraid to touch these groups as well. :slightly_smiling_face:

4n4nd (approved):
looks awesome :ship:

cluster-scope/overlays/moc/groups/approved-users.enc.yaml
```diff
@@ -0,0 +1,58 @@
+apiVersion: user.openshift.io/v1
+kind: Group
+metadata:
+    name: approved-users
+    annotations:
+        kustomize.config.k8s.io/behavior: replace
+users:
```
HumairAK, line 7: I'm seeing that there is a duplicate for harshad's user.

HumairAK: looks like that's how it is on live -- hrmm, we should still fix it

anishasthana: I think making it a separate PR is better. This way we're not introducing multiple changes at once

cluster-scope/overlays/moc/groups/thoth.enc.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: user.openshift.io/v1
+kind: Group
+metadata:
+    name: thoth
+    annotations:
+        kustomize.config.k8s.io/behavior: replace
+users:
+- ENC[AES256_GCM,data:LYosGvpof2CiUW9WwLZlSA==,iv:Qr7bX+QA+viqI2BJaoukJJhCyXwhXowfJTJelhAt63M=,tag:DHARXG7CdBdsj5QCfxn5Rg==,type:str]
+- ENC[AES256_GCM,data:Ls+fuG5HFyXxAZ8fNECQX0k=,iv:q7kpyVNDS7dPVMNc9HBVTdidQmLo+RDOor4gvl/iR9w=,tag:7DcURkZuz5FHbGaIqP55Hg==,type:str]
```
HumairAK, line 9: Christoph's email here is different from what's on live, and what was introduced [here](https://github.com/operate-first/apps/blob/master/namespaces/overlays/moc/thoth-amun-api-prod/group.enc.yaml#L6).

tumido: My content was taken from the MOC cluster before the Harshad's PRs. But Harshad's email is the correct one. Fixing it now.

cluster-scope/overlays/moc/groups/approved-users.enc.yaml
tumido: In all that rebasing I'm doing today this is only a tiny change. I'll remove the duplicity right away

HumairAK (approved):

HumairAK:
LGTM, thanks @tumido  -- awesome work

anishasthana (approved):
:ship:

PR #121, opened 2021-01-18
feat: Add onboarding script

tumido:
Related: #120 
Required by: https://github.com/operate-first/support/pull/40

Pre commit will fail, because of TABs instead of spaces, but HEREDOC requires TABS. :exploding_head:

### DIFF
--- a/hack/onboarding.sh
+++ b/hack/onboarding.sh
@@ -1,0 +1,109 @@
+#!/bin/sh
+if [ $# -le 1 ] || [ "$#" == "--help" ] || [ "$#" == "-h" ]; then
+	echo "Populate the namespace resource definition."
+	echo -e "\nUsage: $0 namespace owner-team [description]\n"
+	exit 0
+fi
+
+APP_NAME="cluster-scope"
+NAMESPACE=$1
+OWNER=$2
+DESCRIPTION=$3
+
+create_namespace() {
+	mkdir -p $APP_NAME/base/namespaces/$NAMESPACE
+	cat <<-EOF > $APP_NAME/base/namespaces/$NAMESPACE/namespace.yaml
+	---
+	apiVersion: v1
+	kind: Namespace
+	metadata:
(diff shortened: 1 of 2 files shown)


HumairAK:
can you update the permissions on this file to enable execute `chmod +x onboarding.sh`

HumairAK:
> Pre commit will fail, because of TABs instead of spaces, but HEREDOC requires TABS. exploding_head 

this effects the entire file so let's just ignore it in the `.pre-commit-config.yaml` add the following line: 

```
exclude: hack/onboarding.sh
```
To this file [here](https://github.com/tumido/apps/blob/namespace-script/.pre-commit-config.yaml#L1)

Similar to how we ignore `/configs` [here](https://github.com/operate-first/continuous-deployment/blob/master/.pre-commit-config.yaml#L2)

hack/onboarding.sh
```diff
@@ -0,0 +1,109 @@
...
+	resources:
+	- namespace.yaml
+
+	components:
+	- ../../../components/project-admin-rolebindings/$OWNER
+	EOF
+	echo "Namespace base created at '$APP_NAME/base/namespaces/$NAMESPACE'."
+}
+
+create_project-admin-rolebinding() {
```
HumairAK, line 42: so afaik hyphens (i.e. `-`) in `names` are a bash thing, not posix -- so this could break on some systems, can we rename this to `create_project_admin_rolebinding`, and the function call below as well.

HumairAK:
Thanks @tumido  -- besides the above, I was able to test the script and it runs fine (also tested `kustomize build` on the resulting manifests, it works), good to merge once the comments above are addressed

tumido:
@HumairAK  updated. Your review on this seems to slipped my attention. :facepalm:

HumairAK (approved):

4n4nd (approved):

PR #130, opened 2021-01-20
Update github receiver

hemajv:
This will:
1. Update the github receiver to enable auto closing of issues for resolved alerts
2. Add the metrics port for the github receiver service so that prometheus can monitor it

### DIFF
--- a/odh/base/monitoring/github-receiver.yaml
+++ b/odh/base/monitoring/github-receiver.yaml
@@ -25,11 +25,13 @@ spec:
                 secretKeyRef:
                   name: github-secrets
                   key: auth-token
-          args: ["--authtoken=$(GITHUB_AUTH_TOKEN)",
-                 "--org=operate-first",
-                 "--repo=SRE",
-                 "--label=in progress",
-                 "--enable-inmemory=false"]
+          args:
+            - --auth-token=$GITHUB_AUTH_TOKEN
+            - --org=operate-first
+            - --repo=SRE
+            - --label="in progress"
+            - --enable-inmemory="false"
+            - --enable-auto-close="true"
           ports:
             - containerPort: 9393
               name: receiver
--- a/odh/base/monitoring/github-receiver-service.yaml
+++ b/odh/base/monitoring/github-receiver-service.yaml
@@ -7,6 +7,8 @@ spec:
   ports:
     - port: 9393
       targetPort: receiver
+    - port: 9990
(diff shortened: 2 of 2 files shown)


4n4nd (approved):
looks good :+1: 
just a heads up, this might close all issues with the same alert label (https://github.com/m-lab/alertmanager-github-receiver#auto-close)

HumairAK (approved):

goern:
What is the rationale behind 'auto close'? Is this something like "close 10 days after the incident is solved", so that a review can happen? do we expect some AI to label auto closed issues that might be interesting to be reviewed?

odh/base/monitoring/github-receiver.yaml
```diff
@@ -29,7 +29,8 @@ spec:
                  "--org=operate-first",
                  "--repo=SRE",
                  "--label=in progress",
-                 "--enable-inmemory=false"]
+                 "--enable-inmemory=false",
+                 "--enable-auto-close=true"]
```
tumido, line 33: Can we make this `args` property into a  block style YAML sequence? Like:
```yaml
- --authtoken=...
- --repo=SRE
- --label="in progress"
...
```
It'll make it a bit more maintainable + easier to add values in future (you're not modifying the values around, since you don't need the `,` and `[]`)

tumido (changes requested):
LGTM, except a tiny nitpick for maintaining coding standards

4n4nd:
> What is the rationale behind 'auto close'? Is this something like "close 10 days after the incident is solved", so that a review can happen? do we expect some AI to label auto closed issues that might be interesting to be reviewed?

I think this a simple "close the issue when the alert is resolved"

hemajv:
> > What is the rationale behind 'auto close'? Is this something like "close 10 days after the incident is solved", so that a review can happen? do we expect some AI to label auto closed issues that might be interesting to be reviewed?
> 
> I think this a simple "close the issue when the alert is resolved"

yep @goern this would just close each issue as its corresponding alert is resolved. It searches for matching issues by filtering open issues on the value of alert label and then matching issue titles: https://github.com/m-lab/alertmanager-github-receiver#auto-close

odh/base/monitoring/github-receiver.yaml
hemajv: @tumido updated :+1:

Issue #126, opened 2021-01-20
Move MOC operators from moc-cnv-sandbox repo to this repo under cluster-scope app


HumairAK:
This will require a sibling PR in [this](https://github.com/open-infrastructure-labs/moc-cnv-sandbox) repo that removes the operators.

tumido:
Operator status board for all operators found on MOC cluster:
| STATUS                                                                   | NAMESPACE           | NAME                            | PACKAGE                         | SOURCE              | CHANNEL        |
| ------------------------------------------------------------------------ | ------------------- | ------------------------------- | ------------------------------- | ------------------- | -------------- |
| No idea where it's defined                                               | local-storage       | local-storage-operator          | local-storage-operator          | redhat-operators    | 4.5            |
| No idea where it's defined                                               | openshift-cnv       | kubevirt-hyperconverged         | kubevirt-hyperconverged         | redhat-operators    | 2.4            |
| https://github.com/operate-first/apps/pull/131                           | openshift-metering  | metering-operator               | metering-ocp                    | redhat-operators    | 4.5            |
| https://github.com/operate-first/apps/pull/131                           | openshift-operators | opendatahub-operator            | opendatahub-operator            | community-operators | beta           |
| https://github.com/operate-first/apps/pull/131                           | openshift-operators | openshift-pipelines-operator-rh | openshift-pipelines-operator-rh | redhat-operators    | ocp-4.5        |
| In the ODH app                                                           | openshift-operators | strimzi                         | strimzi-kafka-operator          | community-operators | strimzi-0.19.x |
| No idea where it's defined                                               | openshift-storage   | ocs-operator                    | ocs-operator                    | redhat-operators    | stable-4.5     |
| In the ODH app                                                           | opf-datacatalog     | radanalytics-spark              | radanalytics-spark              | community-operators | alpha          |
| In the ODH app                                                           | opf-jupyterhub      | prometheus-operator             | prometheus                      | community-operators | beta           |
| In the ODH app                                                           | opf-monitoring      | grafana-operator                | grafana-operator                | community-operators | alpha          |
| In the ODH app                                                           | opf-monitoring      | prometheus-operator             | prometheus                      | community-operators | beta           |
| In the ODH app                                                           | opf-observatorium   | grafana-operator                | grafana-operator                | community-operators | alpha          |
| No idea where it's defined, probably in the thoth app (namespace scoped) | thoth-infra-prod    | prometheus-operator             | prometheus                      | community-operators | beta           |


[...]

PR #131, opened 2021-01-21
feat: Sync down MOC operators to cluster-scope app

tumido:
Resolves: #126

Syncs down all operators found in [`moc-cnv-sandbox` bootstrap repo](https://github.com/open-infrastructure-labs/moc-cnv-sandbox/tree/master/manifests/operators).

PLEASE READ:
1. We may want to split this PR into a PR per operator, because it's big and the probability of me breaking something is huge.
2. The actual change set compared to what is available on the cluster should be minimal, hopefully zero.
2. Requires us to introduce separate apps for some of the operator resources as follow up PRs. https://github.com/operate-first/apps/pull/203 https://github.com/operate-first/apps/pull/202 https://github.com/operate-first/apps/pull/201 https://github.com/operate-first/argocd-apps/pull/60
3. I'm undecided on what to do with the `openshift-*` namespaces, I don't think we should be creating them, but it seems they were not there from the beginning also (they were created later than the cluster was created).


larsks:
> And another uncertainty. This PR syncs down only the operators found in moc-cnv-sandbox/manifests/operators . But the cluster itself is deploying other operators as well. This is reflected back in the #126 issue.

I think it would be a good idea to transition the operators installed via Ansible to use Kustomize instead. That wasn't possible initially because some operators (in particular the 'local storage operator' and OCS) are required before we can deploy something like ArgoCD, but if we're managing the cluster from an outside location, this becomes much easier.

That means adding resources for:

- LSO
- OCS
- CNV

I think once we have those in place, we will want to re-install the
CNV cluster and configure it using ACM/argocd as much as possible
(ideally, we would be able to handle everything via that mechanism).

tumido:
> That means adding resources for:
>
> - LSO
> - OCS
> - CNV

Precisely what I described at https://github.com/operate-first/apps/issues/126

@larsks can you please point me in that issue to all the resources needed and where to find them in the ansible-backed repo?

HumairAK:
> ArgoCD, but if we're managing the cluster from an outside location, this becomes much easier.

@larsks 
I should note that we aren't doing this just yet, the moc-cnv cluster itself is hosting the argocd instance, this is until we migrate to the managed cluster. So as of now we still need to be able to deploy argocd to deploy these operators, maybe it would be best to just include the appropriate manifests for these particular operators, but not have argocd deploy them _just yet_ and have an attached issue to include them once we are done migration. 

> TBH, I'm not sure if we really have to track the Subscriptions in here, since they are namespace scoped resource. WDYT @HumairAK ? 

@tumido 
Looking over this pr and the original location we were storing the operators (i.e. [here](https://github.com/open-infrastructure-labs/moc-cnv-sandbox/tree/master/manifests/operators)) I think we may need to re-think _what exactly we need to track_ for operators. My knowledge of OLM is a bit limited, so feel free to correct me if I've gotten something wrong. 

AFAIK you're right, `subscriptions` are namespace scoped, and we actually don't care about these. What we want to track is, **who** has access to **which** set of extended APIs. And this is done by operatorgroups. Teams cannot deploy the operator in their namespace (or globally) without access to the operator groups, and it seems to me this is what we need to track here. 

OperatorGroups are namespace scoped, but project admins cannot edit them. 

We also need to decide if we will allow operators that target multiple namespaces (i.e. more than one namespace AND not all namespaces). Since multinamespace support maybe [deprecated](https://docs.openshift.com/container-platform/4.6/operators/understanding/olm/olm-understanding-operatorgroups.html#olm-operatorgroups-target-namespace_olm-understanding-operatorgroups) in the future. 

Depending on the requirement the workflow would be:
1. single/multi namespace -> just add an operatorgroup with the required set of apis, but with only that particular namespace(s) as the targetnamespace(s)
2. all namespaces -> If it's a global operator, then you need to append the global operatorgroup (which we would track in `cluster-scope`) with the additional set of APIs introduced by the operator (for example for ODH, it would be kfdef).

For (2) we would deploy the subscription, since it would go in the `openshift-operators` namespace, for (1) the teams would deploy /track the subscriptions themselves.

tumido:
@HumairAK My OLM understanding is limited as well. To my best knowledge I agree with you but it still needs a bit more research on my end - I have that on schedule for tomorrow. :wink: I'll keep you updated here.

cluster-scope/base/clusterroles/observatorium-operator/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+---
+# TODO: Ideally this can be replaced by a observatorium-operator subscription kind resource
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - https://raw.githubusercontent.com/observatorium/operator/03be03063a5cfebb0bf4fbbc8faca3f252bf5006/manifests/cluster_role.yaml
```
4n4nd, line 7: @tumido the observatorium operator kustomization is already available in the [apps repo](https://github.com/operate-first/apps/blob/master/observatorium/base/operator/kustomization.yaml). Can you either remove the Observatorium stuff from this PR or remove the existing observatorium kustomization from the apps repo?

tumido: @4n4nd this is a WIP pr. It's not even following the ADR properly... The ADR is result of me trying things for this PR.

This PR will need to be changed a lot before it's ready to be reviewed.

4n4nd: oops my bad, I didn't see the wip title sorry :see_no_evil:

tumido:
@HumairAK @4n4nd @anishasthana @larsks ready to be reviewed! Please read the PR description above.

larsks (approved):
This all looks fine to me.

It would be nice if we could treat something like "CNV"  as a single module (`resources: [../../modules/cnv]`) rather than as multiple collections of different resource types, but I understand why that poses complications when something may have both cluster-wide resources and namespaced resources.

In particular, with the current organization I can see how when removing something from the configuration one might leave orphan resources lying around.

tumido:
@larsks I get that, the problem we are trying to solve though is that modules obscure the visibility range - we are trying to ensure we don't have 2 different definitions of the same namespace or the same cluster role... I think kustomize itself wouldn't allow us to deploy the same resource multiple times though... I don't know... This is similar request to https://github.com/operate-first/apps/issues/164

4n4nd:
it looks like this is:
"potential orphan resources" vs "potentially overwriting same named resourcs" 
And rn I think "potential orphan resources" is the lesser of the evil. so @tumido 's approach makes sense here.
/lgtm

HumairAK:
/lgtm
Put on hold to stop auto-merge, since this pr is affecting various resources live, we should turn off auto-sync in ArgoCD and carefully match what's added here with live via the ArgoCD UI, then manually sync.

PR #133, opened 2021-01-21
Fix kfdef patch positioning for jh

HumairAK:
Build for JH currently fails


anishasthana (approved):

4n4nd (approved):

4n4nd:
@HumairAK I don't think we need this patch anymore, we can just remove it

PR #136, opened 2021-01-21
Create new app for argocd-monitoring

### LINKED ISSUE #135: Add ArgoCD monitoring in Prometheus
We can use servicemonitors available from [here](https://argoproj.github.io/argo-cd/operator-manual/metrics/)

4n4nd:
Add monitoring rbacs
Add servicemonitor

Closes #135

### DIFF
--- a/odh/overlays/moc/monitoring/servicemonitors/argocd-servicemonitor.yaml
+++ b/odh/overlays/moc/monitoring/servicemonitors/argocd-servicemonitor.yaml
@@ -1,0 +1,14 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: argocd-monitor
+  labels:
+    monitor-component: argocd
+spec:
+  endpoints:
+    - port: metrics
+  namespaceSelector:
+    matchNames:
+      - argocd
+  selector: {}
--- a/argocd-monitoring/base/rbac/kustomization.yaml
+++ b/argocd-monitoring/base/rbac/kustomization.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - monitoring-role.yaml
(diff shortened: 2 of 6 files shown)


4n4nd:
Creating a new app in this repo for argocd monitoring is to keep all our monitoring rbacs in one repo

anishasthana (approved):
I think this looks good

argocd-monitoring/components/rbac/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
```
tumido, line 2: It should not be a component, if this is the core reason for the application to be present. Each kustomize scheme should have a base which is overlayed with optional components pulled to the mix. Can you please move this to the base?

odh/overlays/moc/monitoring/servicemonitors/argocd-servicemonitor.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: monitoring.coreos.com/v1
```
tumido, line 2: I think this should be a separate PR.

tumido (changes requested):

odh/overlays/moc/monitoring/servicemonitors/argocd-servicemonitor.yaml
tumido: Scratch that.. I didn't get this bit fully. It should stay here. Sorry. :facepalm:

argocd-monitoring/components/rbac/kustomization.yaml
4n4nd: moved it to base :+1:

tumido:
@HumairAK @4n4nd 's gonna follow up with a PR moving this into https://github.com/operate-first/apps/tree/master/cluster-scope/components :+1:

tumido (approved):

PR #137, opened 2021-01-21
feat: Update ODH Argo to latest

tumido:
Updates Argo to 2.12.3

Requires: https://github.com/opendatahub-io/odh-manifests/pull/208 and https://github.com/operate-first/odh-dashboard/pull/4

### DIFF
--- a/odh/base/argo/kfdef.yaml
+++ b/odh/base/argo/kfdef.yaml
@@ -14,15 +14,17 @@ spec:
         name: odh-common
     - kustomizeConfig:
         repoRef:
-          name: manifests
+          name: master
           path: odhargo/cluster
       name: odhargo-cluster
     - kustomizeConfig:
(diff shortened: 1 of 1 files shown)


anishasthana:
Are you sure we want to base things on master?

tumido:
Would you rather see me lock it to a commit? I don't see it as much of an issue if we're pulling the apps we're kinda maintaining and contributing to anyways. We're not waiting for ODH release cycle. We do the same with the data catalog. We're lock it to the nearest release when possible.

anishasthana:
Fair enough

anishasthana (approved):

HumairAK (approved):
We can update it to a release in the moc env once a release captures these changes, that's where it really matters anyway since it's supposed to be a prod environment.

PR #142, opened 2021-01-22
feat: Remove obsolete project-admin rbac component from ODH and Observatorium

### LINKED ISSUE #140: Clean up ODH/observatorium RBAC components for monitoring
Declutter the monitoring RBAC spec in a way that we have only a single definition within the `cluster-scope/components` and enable it for our namespaces in `cluster-scope/namespaces`.

tumido:
Resolves #140 

Identical RBAC is deployed within the `cluster-scope` for the namespaces in:

```
cluster-scope/base/namespaces/opf-argo/kustomization.yaml
cluster-scope/base/namespaces/opf-dashboard/kustomization.yaml
cluster-scope/base/namespaces/opf-datacatalog/kustomization.yaml
cluster-scope/base/namespaces/opf-jupyterhub/kustomization.yaml
cluster-scope/base/namespaces/opf-kafka/kustomization.yaml
cluster-scope/base/namespaces/opf-monitoring/kustomization.yaml
cluster-scope/base/namespaces/opf-observatorium/kustomization.yaml
cluster-scope/base/namespaces/opf-superset/kustomization.yaml
```


HumairAK (approved):

4n4nd:
@tumido should we keep the kustomization and just remove the rolebinding since this is the only kustomization that applies to all of our opf namespaces and not any other namespace

tumido:
@4n4nd I don't understand. Which kustomization you mean right now? If you look up the namespace definitions, they add a component for the `operate-first` group admin them. Other namespaces may have different project admin components applied, and that's fine. Let's define the project admins on the namespace level only.

4n4nd:
@tumido I mean the ` odh/components/rbac/kustomization.yaml` kustomization, it applies only to the opf namespaces

HumairAK:
The way I understand what @4n4nd  is asking is that the components folder in odh is one way to easily update all odh components since it's already importing it (though in this case it was rbac specifically, but I guess it could be more generalized). Is this right @4n4nd ? 

In either case, I think we should remove it for now and if the need for something like this arises we'll accommodate then. Otherwise it could end up being clutter.

4n4nd:
> The way I understand what @4n4nd is asking is that the components folder in odh is one way to easily update all odh components since it's already importing it (though in this case it was rbac specifically, but I guess it could be more generalized). Is this right @4n4nd ?

yes exactly

4n4nd:
> In either case, I think we should remove it for now and if the need for something like this arises we'll accommodate then. Otherwise it could end up being clutter.

sounds good :+1:

4n4nd (approved):

tumido:
Ah, I see! @4n4nd  yes, I kinda agree with removing it for now and once we have really something this ODH specific we can add it any time. This rbac role was too general to my liking to be kept within the application - I'd prefer all the project admins of a specific namespace to be well defined and in single place only. What do you think?

I know such component might be useful, but keeping empty kustomize components around while still being imported to all ODH namespaces sounds quite error prone to me...

Issue #146, opened 2021-01-22
Prow CRD not available


goern:
```
oc get prowjobs.prow.k8s.io                                                                                                                                                                                                                                 Error from server (Forbidden): prowjobs.prow.k8s.io is forbidden: User "cgoern@redhat.com" cannot list resource "prowjobs" in API group "prow.k8s.io" in the namespace "opf-ci-prow"
```

this is also not working of all of prow's services accounts.

/sig devops
/kind bug

HumairAK:
guessing we need to add an aggregate role here

tumido:
I'll take care of this as promised. :wink:

tumido:
PR Pending https://github.com/operate-first/apps/pull/149

PR #149, opened 2021-01-22
feat: make Prow CRD static and add aggregate cluster roles

tumido:
Fixes: https://github.com/operate-first/apps/issues/146

Also make the CRD static and not pulling it from Thoth (they can put anything in that file actually...) This locks it down.


HumairAK:
Yes I was wondering about this as well, we probably don't want to allow remote references.

HumairAK (approved):

anishasthana (approved):
Definitely safer. +1

PR #152, opened 2021-01-22
Add basic up alerts for opf services

hemajv:
Adding basic "up" alerts for available opf services, based on grafana dashboard created [here](https://grafana-route-opf-monitoring.apps.cnv.massopen.cloud/d/r7WqgaBMk/operatefirst-availability?orgId=1&refresh=1m)

This closes https://github.com/operate-first/SRE/issues/33

### DIFF
--- a/odh/base/monitoring/overrides/prometheus-operator/overlays/alerts/prometheus-rules.yaml
+++ b/odh/base/monitoring/overrides/prometheus-operator/overlays/alerts/prometheus-rules.yaml
@@ -11,14 +11,74 @@ spec:
     - name: Operate First Availability
       rules:
         - alert: Prometheus Service Down
-          expr: up{service="prometheus-operated"} == 0
+          expr: count(up{service="prometheus-operated", pod=~"prometheus-odh-monitoring.*"} == 0) > 1
           for: 5m
           annotations:
             summary: "Prometheus service is no longer available"
             severity: "HIGH"
         - alert: Prometheus Alertmanager Down
-          expr: up{service="alert-manager"} == 0
+          expr: absent(up{service="alert-manager"} == 1)
           for: 5m
           annotations:
             summary: "Prometheus Alertmanager is no longer available"
+            severity: "HIGH"
+        - alert: Grafana Service Down
+          expr: absent(up{service="grafana-service"} == 1)
+          for: 5m
+          annotations:
+            summary: "Grafana service is no longer available"
+            severity: "HIGH"
+        - alert: Grafana Operator
+          expr: absent(up{service="grafana-operator-metrics"} == 1)
+          for: 5m
+          annotations:
+            summary: "Grafana operator metrics not available"
+            severity: "LOW"
+        - alert: JupyterHub Service Down
+          expr: absent(up{service="jupyterhub"} == 1)
+          for: 5m
+          annotations:
+            summary: "JupyterHub service is no longer available"
+            severity: "HIGH"
+        - alert: ArgoCD Repo Server Down
+          expr: absent(up{service="argocd-repo-server"} == 1)
+          for: 5m
+          annotations:
+            summary: "ArgoCD repo server is no longer available"
+            severity: "HIGH"
+        - alert: ArgoCD Server Down
+          expr: absent(up{service="argocd-server-metrics"} == 1)
+          for: 5m
+          annotations:
+            summary: "ArgoCD server is no longer available"
+            severity: "HIGH"
+        - alert: ArgoCD Application Controller Down
+          expr: absent(up{service="argocd-metrics"} == 1)
+          for: 5m
+          annotations:
+            summary: "ArgoCD application controller is no longer available"
+            severity: "HIGH"
+        - alert: Observatorium Loki Ingester Down
+          expr: absent(up{service="opf-observatorium-loki-ingester-http"} == 1)
+          for: 5m
+          annotations:
+            summary: "Loki ingester is no longer available"
+            severity: "HIGH"
+        - alert: Observatorium Loki Distributor Down
+          expr: absent(up{service="opf-observatorium-loki-distributor-http"} == 1)
+          for: 5m
+          annotations:
+            summary: "Loki distributor is no longer available"
+            severity: "HIGH"
+        - alert: Observatorium Loki Querier Down
+          expr: absent(up{service="opf-observatorium-loki-querier-http"} == 1)
+          for: 5m
+          annotations:
+            summary: "Loki querier is no longer available"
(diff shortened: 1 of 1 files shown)


4n4nd (changes requested):
@hemajv can we please add the argocd services in here?
http://prometheus-portal-opf-monitoring.apps.cnv.massopen.cloud/graph?g0.range_input=1h&g0.expr=up%7Bnamespace%3D%22argocd%22%7D&g0.tab=1

hemajv:
> @hemajv can we please add the argocd services in here?
> http://prometheus-portal-opf-monitoring.apps.cnv.massopen.cloud/graph?g0.range_input=1h&g0.expr=up%7Bnamespace%3D%22argocd%22%7D&g0.tab=1

yes we can, but do we want to alert on each of them?:
- `argocd-repo-server`
- `argocd-dex-server`
- `argocd-server-metrics`
-  `argocd-metrics`

same for observatorium: http://prometheus-portal-opf-monitoring.apps.cnv.massopen.cloud/graph?g0.range_input=1h&g0.expr=up%7Bnamespace%3D%22opf-observatorium%22%7D&g0.tab=1 , not sure if we want to alert on all of them?

4n4nd:
@HumairAK and @tumido might know better what each deployment does and if it's necessary to set alerts on each of them

HumairAK:
yes we should alert on all components

odh/base/monitoring/overrides/prometheus-operator/overlays/alerts/prometheus-rules.yaml
```diff
@@ -11,14 +11,32 @@ spec:
     - name: Operate First Availability
       rules:
         - alert: Prometheus Service Down
-          expr: up{service="prometheus-operated"} == 0
+          expr: count(up{service="prometheus-operated", pod=~"prometheus-odh-monitoring.*"} == 0) > 1
```
anishasthana, line 14: Should this say less than 1?

odh/base/monitoring/overrides/prometheus-operator/overlays/alerts/prometheus-rules.yaml
```diff
@@ -11,14 +11,32 @@ spec:
...
             summary: "Prometheus Alertmanager is no longer available"
             severity: "HIGH"
+        - alert: Grafana Service Down
+          expr: absent(up{service="grafana-service"} == 1)
+          for: 5m
+          annotations:
+            summary: "Grafana Service is no longer available"
+            severity: "HIGH"
+        - alert: Grafana Operator
+          expr: absent(up{service="grafana-operator-metrics"} == 1)
```
anishasthana, line 32: I feel like the grafana operator being unavailable is less bad than grafana/prometheus/argocd/alertmanager not being available.

anishasthana (changes requested):

tumido:
@hemajv For the `argocd`, I think "the most important" `Service`s are:
- `argocd-repo-server`
- `argocd-server` (maps to the same deployment as `argocd-server-metrics` service)
- `argocd-metrics` (maps to a stateful set `argocd-application-controller`)

There's also `argocd-redis` `Service`, which seems not to be exported to the Prometheus, though this is not a "HIGH" severity, but would degrade the ArgoCD if down.

The `argocd-dex-service` will be replaced by a standalone Dex deployment in near future, so you might want to skip that one for now.

@HumairAK am I saying that right?

odh/base/monitoring/overrides/prometheus-operator/overlays/alerts/prometheus-rules.yaml
hemajv: This basically checks how many pods are down i.e. `up{service="prometheus-operated", pod=....} == 0`. Since we have 2 pods and if **both** are down we would want to be alerted on it, so hence if `count(up{service="prometheus=operated", pod=....}) == 0) > 1` would indicate that both pods are down. Does that make sense?

odh/base/monitoring/overrides/prometheus-operator/overlays/alerts/prometheus-rules.yaml
hemajv: so just have this in the grafana dashboard, but not be alerted on it?

tumido: Maybe just lowering the severity to `LOW` would be fine?

HumairAK:
@tumido  yeah that's absolutely correct, `repo-server`, `application-controller`, and `server` are the top 3, `redis` would be nice, but if the metrics aren't available then, we can put it on the back log for now.

odh/base/monitoring/overrides/prometheus-operator/overlays/alerts/prometheus-rules.yaml
hemajv: updated

hemajv:
@4n4nd I added the `observatorium` components as well. Could you please review them and let me know if the `severity` labels for each of them need to be changed?

odh/base/monitoring/overrides/prometheus-operator/overlays/alerts/prometheus-rules.yaml
anishasthana: Wasn't our concern also around the metrics disappearing altogether? @4n4nd^^

hemajv: ah yeah, then in that case wouldn't it be the DeadMansSwitch alert, because that essentially is to notify if prometheus itself is down and `count` will return "No Data" since the `pod` labels would be missing

4n4nd:
> @4n4nd I added the `observatorium` components as well. Could you please take a look and also let me know if you think the `severity` labels for each of them need to be changed?

@hemajv this looks good :+1: , all the deployments should be alerted on

odh/base/monitoring/overrides/prometheus-operator/overlays/alerts/prometheus-rules.yaml
4n4nd: it was, I think this should work and if it does not this is still a good starting point

4n4nd (approved):
looks great! :ship:

tumido (approved):

tumido:
@anishasthana  can you please re-review this PR, I think your requested-changes is blocking this PR to be merged. :slightly_smiling_face:

anishasthana (approved):

PR #153, opened 2021-01-24
Deploy Cluster-Logging-Operator (CLO) 

eranra:
Including dependency on Elastic-Search-Operator

### DIFF
--- a/cluster-scope/base/clusterloggings/clusterLogging.yaml
+++ b/cluster-scope/base/clusterloggings/clusterLogging.yaml
@@ -1,0 +1,35 @@
+---
+apiVersion: "logging.openshift.io/v1"
+kind: "ClusterLogging"
+metadata:
+  name: "instance"
+spec:
+  managementState: "Managed"
+  logStore:
+    type: "elasticsearch"
+    retentionPolicy:
+      application:
+        maxAge: 1d
+      infra:
+        maxAge: 7d
+      audit:
+        maxAge: 7d
+    elasticsearch:
+      nodeCount: 3
+      storage:
+        # TODO: If need to set specific storageClassName
+        # storageClassName: "<storage-class-name>"
+        size: 200G
+      redundancyPolicy: "SingleRedundancy"
+  visualization:
+    type: "kibana"
+    kibana:
+      replicas: 1
+  curation:
+    type: "curator"
+    curator:
+      schedule: "30 3 * * *"
+  collection:
+    logs:
+      type: "fluentd"
+      fluentd: {}
--- a/cluster-scope/base/operatorgroups/cluster-logging-operator/operatorgroup.yaml
+++ b/cluster-scope/base/operatorgroups/cluster-logging-operator/operatorgroup.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: cluster-logging
+spec:
+  targetNamespaces:
+    - openshift-logging
--- a/cluster-scope/overlays/moc/kustomization.yaml
+++ b/cluster-scope/overlays/moc/kustomization.yaml
@@ -4,6 +4,7 @@ kind: Kustomization
 
 resources:
   - ../../base/clusterroles/prow.k8s.io
+  - ../../base/clusterloggings
   - ../../base/crds/prow.k8s.io
   - ../../base/groups/approved-users
   - ../../base/groups/argocd-admins
@@ -19,6 +20,8 @@ resources:
   - ../../base/namespaces/observatorium-operator
   - ../../base/namespaces/odh-operator
   - ../../base/namespaces/open-aiops
+  - ../../base/namespaces/openshift-logging
+  - ../../base/namespaces/openshift-operators-redhat
   - ../../base/namespaces/opf-argo
   - ../../base/namespaces/opf-ci-pipelines
   - ../../base/namespaces/opf-ci-prow
@@ -38,6 +41,10 @@ resources:
   - ../../base/namespaces/thoth-graph-prod
   - ../../base/namespaces/thoth-infra-prod
   - ../../base/namespaces/thoth-middletier-prod
+  - ../../base/operatorgroups/elastic-operator
+  - ../../base/operatorgroups/cluster-logging-operator
+  - ../../base/subscriptions/elastic-operator
+  - ../../base/subscriptions/cluster-logging-operator
 
 generators:
   - secret-generator.yaml
--- a/cluster-scope/base/clusterloggings/kustomization.yaml
+++ b/cluster-scope/base/clusterloggings/kustomization.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: openshift-logging
+
+resources:
+  - clusterLogging.yaml
--- a/cluster-scope/base/namespaces/openshift-logging/kustomization.yaml
+++ b/cluster-scope/base/namespaces/openshift-logging/kustomization.yaml
@@ -1,0 +1,11 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: openshift-logging
+
+resources:
+  - namespace.yaml
+
+components:
+  - ../../../components/project-admin-rolebindings/operate-first
--- a/cluster-scope/base/namespaces/openshift-logging/namespace.yaml
+++ b/cluster-scope/base/namespaces/openshift-logging/namespace.yaml
@@ -1,0 +1,12 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    openshift.io/display-name: "Cluster Logging Operator"
+    openshift.io/requester: operate-first
+    openshift.io/node-selector: ""
+  labels:
+    openshift.io/cluster-monitoring: "true"
+  name: openshift-logging
+spec: {}
--- a/cluster-scope/base/namespaces/openshift-operators-redhat/kustomization.yaml
+++ b/cluster-scope/base/namespaces/openshift-operators-redhat/kustomization.yaml
@@ -1,0 +1,11 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: openshift-operators-redhat
+
+resources:
+  - namespace.yaml
+
+components:
+  - ../../../components/project-admin-rolebindings/operate-first
--- a/cluster-scope/base/namespaces/openshift-operators-redhat/namespace.yaml
+++ b/cluster-scope/base/namespaces/openshift-operators-redhat/namespace.yaml
@@ -1,0 +1,12 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    openshift.io/display-name: "Openshift Operators Redhat"
+    openshift.io/requester: operate-first
+    openshift.io/node-selector: ""
+  labels:
+    openshift.io/cluster-monitoring: "true"
+  name: openshift-operators-redhat
+spec: {}
--- a/cluster-scope/base/operatorgroups/cluster-logging-operator/kustomization.yaml
+++ b/cluster-scope/base/operatorgroups/cluster-logging-operator/kustomization.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: openshift-logging
+
+resources:
+  - operatorgroup.yaml
--- a/cluster-scope/base/operatorgroups/elastic-operator/kustomization.yaml
+++ b/cluster-scope/base/operatorgroups/elastic-operator/kustomization.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: openshift-operators-redhat
+
+resources:
+  - operatorgroup.yaml
--- a/cluster-scope/base/operatorgroups/elastic-operator/operatorgroup.yaml
+++ b/cluster-scope/base/operatorgroups/elastic-operator/operatorgroup.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: openshift-operators-redhat
+spec: {}
--- a/cluster-scope/base/subscriptions/cluster-logging-operator/kustomization.yaml
+++ b/cluster-scope/base/subscriptions/cluster-logging-operator/kustomization.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: openshift-logging
+
+resources:
+  - subscription.yaml
--- a/cluster-scope/base/subscriptions/cluster-logging-operator/subscription.yaml
+++ b/cluster-scope/base/subscriptions/cluster-logging-operator/subscription.yaml
@@ -1,0 +1,11 @@
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: cluster-logging
+spec:
+  channel: "4.6"
+  installPlanApproval: Automatic
+  name: cluster-logging
+  source: redhat-operators
+  sourceNamespace: openshift-marketplace
--- a/cluster-scope/base/subscriptions/elastic-operator/kustomization.yaml
+++ b/cluster-scope/base/subscriptions/elastic-operator/kustomization.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: openshift-operators-redhat
+
+resources:
+  - subscription.yaml
--- a/cluster-scope/base/subscriptions/elastic-operator/subscription.yaml
+++ b/cluster-scope/base/subscriptions/elastic-operator/subscription.yaml
@@ -1,0 +1,11 @@
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: elasticsearch-operator
+spec:
+  channel: "4.6"
+  installPlanApproval: Automatic
+  name: elasticsearch-operator
+  source: redhat-operators
+  sourceNamespace: openshift-marketplace


anishasthana:
@eranra a lot of the files are missing new lines. Could you add them in?

eranra:
@anishasthana done

cluster-scope/base/crds/clusterloggings.logging.openshift.io/crd.yaml
```diff
@@ -0,0 +1,25 @@
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: clusterloggings.logging.openshift.io
```
tumido, line 5: This CRD should be created automatically by the operator subscription, isn't it? I don't think we need it at all.

cluster-scope/components/logging-rbac/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+---
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
```
tumido, line 3: Instead of defining another component, I'd suggest to use the recently added [`components/monitoring-rbac`](https://github.com/operate-first/apps/tree/master/cluster-scope/components/monitoring-rbac) component.

cluster-scope/base/routes/elastic-operator/route.yaml
```diff
@@ -0,0 +1,19 @@
+---
+apiVersion: route.openshift.io/v1
+kind: Route
```
tumido, line 3: I don't like routes to be created within the cluster-scope application, they are really namespace specific are are just cluttering things in here. WDYT @HumairAK , I'd suggest to piggy back on the `Subscription` resource in this case and just deploy it together with the operator.

tumido (changes requested):

cluster-scope/base/crds/clusterloggings.logging.openshift.io/crd.yaml
eranra: @tumido, this is not part of subscription ... we need seperate crd, look here: https://docs.openshift.com/container-platform/4.6/logging/cluster-logging-deploying.html#cluster-logging-deploy-cli_cluster-logging-deploying

cluster-scope/components/logging-rbac/kustomization.yaml
eranra: Monitoring and logging are diffrent Obseravbility components in OCP. BTW; To the best of my knowledge monitoring is "out-of-the-box" so I do not really understand why you need to deploy. In any event, having specific role and rolebinding will allow specific rbac control on logging, we might need that for exposure etc.

cluster-scope/base/routes/elastic-operator/route.yaml
eranra: @tumido This route exposes the entire elastic-search instance - there was another option which is to connect Logging to Kafka Bus ... if you can explain / point to Kafka ... we can drop this route and forward the logs to Kafka

eranra:
Uploaded new version - with more accurate deployment details from:: 

https://docs.openshift.com/container-platform/4.6/logging/cluster-logging-deploying.html#cluster-logging-deploy-cli_cluster-logging-deploying

cluster-scope/base/routes/elastic-operator/route.yaml
eranra: @tumido for now I disabled (comment) the deployment of this route as part of the main kustomization.yaml file ... so let's see that this deployment works and continue from there

cluster-scope/base/crds/clusterloggings.logging.openshift.io/crd.yaml
tumido: I don't see it anywhere in there. Neither the install steps mention anything about deploying the CRD manually, it only talks about `Subscription`, `OperatorGroup` and the instance itself. What am I missing? Please point me to the exact location, I feel dumb now. :smile: 

My point is: I don't want the operator and ArgoCD to fight over the CRD if there are any changes to the spec/scheme of it. Because that would mean there are 2 apps trying to deploy the same resource which would lead to out of sync issues and errors.

eranra: @tumido Now I get your point ... **you are correct** ... I am not adding crd, I am adding **"the instance itself"** ... but I put that under crd's (the kind of the object is "ClusterLogging") ... should I put this in a diffrent folder than ???

tumido: That is a good point! I think within the `crds` folder only resources of `CustomResourceDefinition` kind should live.

As I can see the `ClusterLogging` resource is a namespaced resource. Does a user need cluster-admin privileges to be able to deploy it or can a project admin do it?

If a project admin can deploy this resource, I don't think the `ClusterLogging` resource should live within the `cluster-scope` application then. We reserved this application only for the things you need cluster-admin access to deploy + some basic bootstrapping, like creating project admins etc (see [ADR 0009](https://www.operate-first.cloud/blueprints/blueprint/docs/adr/0009-cluster-resources.md)).

If something is a regular namespaced resource, it should be deployed as part of a namespace-bound application (see [ADR 0004](https://www.operate-first.cloud/blueprints/blueprint/docs/adr/0004-argocd-apps-of-apps-structure.md) for reference). If you want this to be fully managed by Operate-first org and admins, you can put it within the `operate-first/apps` repo root folder. Or, you can put it within your own repository that we would deploy via ArgoCD.

If it's the other case and this resource really requires cluster-admin permissions to be even deployed, we need to include it within the `cluster-scope` application. In that case I'd put it within a separate folder like: `cluster-scope/base/clusterloggings` (using the [plural name from CRD](https://github.com/openshift/cluster-logging-operator/blob/7b557d8beb7de82617b9df1612d26312db619a42/manifests/4.7/logging.openshift.io_clusterloggings_crd.yaml#L12) as the folder name).

Sorry to bother you with this, I know it's not that well documented and if you have any suggestions on how we can improve it, we'd be glad to do it.

cluster-scope/components/logging-rbac/kustomization.yaml
tumido: I see what's the change. Your rolebinding uses a different `namespace` in the `subjects`. I get it now. I suggest you enhance the monitoring-rbac component rolebinding to include one additional subject for your namespace and it would grant you monitoring access to all the namespaces where we deploy the monitoring RBAC. WDYT?

cluster-scope/base/routes/elastic-operator/route.yaml
tumido: I don't want to clutter the `cluster-scope` application with a namespaced resources that much. Can you please either remove it or put it within a separate application in the repository root?

cluster-scope/overlays/moc/kustomization.yaml
```diff
@@ -37,6 +37,16 @@ resources:
   - ../../base/namespaces/thoth-graph-prod
   - ../../base/namespaces/thoth-infra-prod
   - ../../base/namespaces/thoth-middletier-prod
+  # Elastic Search Operator & Cluster Logging Operator
+  # Details: ( https://docs.openshift.com/container-platform/4.6/logging/cluster-logging-deploying.html#cluster-logging-deploy-cli_cluster-logging-deploying )
+  - ../../base/namespaces/openshift-operators-redhat
+  - ../../base/namespaces/openshift-logging
+  - ../../base/operator-groups/elastic-operator
```
tumido, line 44: Few suggestions:
1. Can you please rename the `operator-groups` to be `operatorgroups`? This way it would match the "plural name" for the resource kind.
2. Can you please use the alphabetical ordering of the resources? We don't keep them together by their usage, but rather make them easier to locate in the repo.

odh/overlays/moc/monitoring/dashboards/operate-first/argocd.yaml
```diff
@@ -3852,4 +3852,4 @@ spec:
       "uid": "97AYZMTGk",
       "version": 1
     }
-  name: argocd.json
\ No newline at end of file
+  name: argocd.json
```
tumido, line 3855: Please undo this change.

tumido (changes requested):
I think you can greatly benefit from decoupling this application - please keep in the `cluster-scope` only the necessary things which require cluster admin permissions (all the rest should live in namespaced applications) and reuse as much of the available components as possible.

odh/overlays/moc/monitoring/dashboards/operate-first/argocd.yaml
tumido: Oh, I see.. this one is breaking the pre-commit check. Don't mind my comment. It seems multiple people are running into this...

cluster-scope/overlays/moc/kustomization.yaml
eranra: (1) Done
(2) Done

odh/overlays/moc/monitoring/dashboards/operate-first/argocd.yaml
eranra: I removed this from this #PR ... as you requested ... you can create another small PR to fix only this ... this is not related to this PR so make sense not to include ... :-)

cluster-scope/base/operator-groups/cluster-logging-operator/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: openshift-logging
+
+resources:
+  - operatorgroup.yaml
+
+components:
+  - ../../../components/project-admin-rolebindings/operate-first
```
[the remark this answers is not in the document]
eranra: Fixed, thanks .. DONE

cluster-scope/base/operator-groups/elastic-operator/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: openshift-operators-redhat
+
+resources:
+  - operatorgroup.yaml
+
+components:
+  - ../../../components/project-admin-rolebindings/operate-first
```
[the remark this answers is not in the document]
eranra: Correct, thanks ... removed, DONE :-)

cluster-scope/base/crds/clusterloggings.logging.openshift.io/crd.yaml
```diff
@@ -0,0 +1,36 @@
+---
+apiVersion: "logging.openshift.io/v1"
+kind: "ClusterLogging"
+metadata:
+  name: "instance"
+  namespace: "openshift-logging"
```
4n4nd, line 6: @eranra this looks like a namespaced CR and not a cluster-scope CRD, I don't think this should be in base/crds/..

tumido: @4n4nd yes, I think you're pointing to the same thing: https://github.com/operate-first/apps/pull/153#discussion_r564420099

cluster-scope/base/crds/clusterloggings.logging.openshift.io/crd.yaml
4n4nd: I think this should be another application just like [`odh`](https://github.com/operate-first/apps/tree/master/odh/base/operator) and [`observatorium`](https://github.com/operate-first/apps/tree/master/observatorium/base/operator), we can keep the subscriptions and operatorgroups in the base.

cluster-scope/base/crds/clusterloggings.logging.openshift.io/crd.yaml
4n4nd: oh oops nvm

cluster-scope/base/crds/clusterloggings.logging.openshift.io/crd.yaml
eranra: @tumido and @4n4nd cluster logging is not application level deployment, this is cluster level deployment and must be (1) deployed by cluster-admin (it deployed privilaged containers that listen to host level logs etc ...)  (2) be deployed in openshift-logging namespace which is cluster level namespace and not belong to users. This is very diffrent from application like `odh` and such.

eranra: @tumido I moved the .yaml into `clusterloggings` folder

cluster-scope/components/logging-rbac/kustomization.yaml
eranra: I removed the entire folder ... from what I see in 4.6 deployments this is not needed

cluster-scope/base/routes/elastic-operator/route.yaml
eranra: removed the route all together for now ... let's start to see logs and that I will work with Bruno on how to expose them ... I agree that this potentially can be app level.

cluster-scope/base/operatorgroups/cluster-logging-operator/operatorgroup.yaml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: cluster-logging
+  namespace: openshift-logging
+spec:
+  targetNamespaces:
+    - openshift-logging
```
tumido, line 9: I'm still thinking about this from the #131 perspective. I'm inclined to unblock you here and merge this and then do some global pruning on the namespaced operators in #131. I'm think we need an ADR for operator handling. :smile: 

https://github.com/operate-first/blueprint/issues/23

@HumairAK mind you take a look at this PR as well?

cluster-scope/base/operatorgroups/cluster-logging-operator/operatorgroup.yaml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: cluster-logging
+  namespace: openshift-logging
```
tumido, line 6: You don't need to specify a namespace in here, since the `namespace` from `kustomization.yaml` would overwrite this anyways. You can delete those from multiple resources in here.

cluster-scope/base/clusterloggings/clusterLogging.yaml
```diff
@@ -0,0 +1,36 @@
+---
+apiVersion: "logging.openshift.io/v1"
+kind: "ClusterLogging"
+metadata:
+  name: "instance"
+  namespace: "openshift-logging"
```
tumido, line 6: Please move the namespace here into the `kustomization.yaml` so we are consistent across all the manifests.

cluster-scope/base/clusterloggings/clusterLogging.yaml
```diff
@@ -0,0 +1,36 @@
+---
+apiVersion: "logging.openshift.io/v1"
+kind: "ClusterLogging"
```
tumido, line 3: I like this much more :+1: We might want to move it to a separate app (outside of the `cluster-scope`) if it's deployable without a cluster-admin permissions. I need to test this out.

tumido (changes requested):
Except to small nitpicks about namespaces (in resources vs. in kustomization file - the 2nd is preferred), this is much more polished. I think the operators would require much more precise directions for the future, so I'm fine with the current layout for now - I don't want to block you.

cluster-scope/base/clusterloggings/clusterLogging.yaml
eranra: fixed, done.

cluster-scope/base/operatorgroups/cluster-logging-operator/operatorgroup.yaml
eranra: fixed here, and multiple other places :-)

tumido (approved):
LGTM :+1: Except those discussion points about operators - but we'd need a broader discussion on that. Tracking at: https://github.com/operate-first/apps/issues/166

HumairAK:
This was not meant to be merged, a recent change to branch protections caused our ci bot to instantly merge this. Because our deployments are automated, some of these changes have been applied live. I'm looking into the deployment and it seems and there seems to be a few issues,  we'll likely need to have a follow up pr to fix it. 

image
image

The namespaces, subscriptions, rbac, and operatorgroups were created, but the cluster-logging instance cannot be created since the operator is not deploying (see above).

eranra:
@HumairAK maybe this is due to something with the subscription ... maybe https://github.com/operate-first/apps/pull/153/files#diff-9b83488b876a5447a689d7b36fbec3808ac5edaae2e9ae7ae9adeb7dea3d0d2aR11 requiers some additional prereqs .... not sure.

Can I access this area in the cluster ? If not I count on you and the team to fix ... :-)

4n4nd:
@eranra we were able to fix that issue by changing the subscription channel to `4.5` since the cluster version we have is `4.5.17`. Can you please read through the instructions for [v4.5](https://docs.openshift.com/container-platform/4.5/logging/cluster-logging-deploying.html) and check if there are any other changes that we should make?

HumairAK:
see [here](https://github.com/operate-first/apps/pull/169) for update

HumairAK:
this one CR looks like it's parent to quite a large number of resources, it's really bloating the ArgoCD ui for cluster resources, I'm going to follow up with another pr and move the CR to a separate ArgoCD application. (This is mostly organizational)

4n4nd:
> this one CR looks like it's parent to quite a large number of resources, it's really bloating the ArgoCD ui for cluster resources, I'm going to follow up with another pr and move the CR to a separate ArgoCD application. (This is mostly organizational)

@eranra @tumido I still think it should be it's own app just like odh and observatorium

eranra:
@HumairAK **thanks** for the fix, I did not know that this was 4.5 cluster, I thought we are on 4.6 👍 

I see that https://kibana-openshift-logging.apps.cnv.massopen.cloud/ is working so elastic and kibana operators are probably working as expected ... but I do not see any data coming in from CLO into elastic so no data in kibana ... and I have no way to see those namespaces (because I do not have access) ... do you see the ClusterLoggingOperator working and the instance CRD deployed correctly ... if so ... I will need logs/events from the containers or access to understand why things are not working as expected ... 

We might need to check if there are diffrences between CLO 4.5 and CLO 4.6 ideployment instruction diffrences 

FYI @tumido ^^^

durandom:
@eranra what namespace do you need read access to? I think we should be able to get everybody read access.

HumairAK:
@eranra what's your moc login?

HumairAK:
@eranra  you should have view access to `openshift-logging` let me know if there's anything else we can do

tumido:
> @eranra @tumido I still think it should be it's own app just like odh and observatorium

@4n4nd and you are absolutely correct about that. I think the same - I was actually willing to include it in the `cluster-scope` in order to unblock @eranra and then follow up on it.

PR #155, opened 2021-01-25
rework OWNERS [WIP]

goern:
:sparkles: rework OWNERS as mentioned in https://github.com/open-infrastructure-labs/ops-issues/issues/16

Signed-off-by: Christoph Görn <goern@redhat.com>

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,18 +1,15 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
 approvers:
-  - anishasthana
-  - hemajv
   - HumairAK
   - tumido
-  - martinpovolny
-  - 4n4nd
-  - durandom
 
 reviewers:
+  - durandom
   - anishasthana
   - hemajv
-  - HumairAK
-  - tumido
   - martinpovolny
   - 4n4nd
+
+labels:
+  - sig/documentation


OWNERS
```diff
@@ -1,18 +1,15 @@
...
+  - durandom
   - anishasthana
   - hemajv
-  - HumairAK
-  - tumido
   - martinpovolny
   - 4n4nd
+
+labels:
+  - sig/documentation
```
tumido, line 15: This PR is against the `operate-first/apps` root folder. I don't think `sig/documentation` is the right label. :) I think you can drop this bit here.

tumido (changes requested):

OWNERS
HumairAK: Looking at [this](https://github.com/kubernetes/kubernetes/blob/master/OWNERS#L33) it looks like you can apply labels to a file type regex. Maybe we can have one for *.md and `sig/documentation` 

I may not be fully understanding `filters` however.

HumairAK: Other cool ones would be if the `cluster-scope` is changed, a "cluster-scope` label or something similar would be nice, or if it's the odh path have it labelled `odh`

I could be getting carried away here.

tumido: I think we can also do this with separate `OWNERS` files per each sub folder or via [filters](https://www.kubernetes.dev/docs/guide/owners/#filters)

goern: all the things said are right and good, feel free to open a PR against this PR ;)

goern:
BTW... I just figured this is the wrong repo ;) we said we gonna do it on `support/` :facepalm:

anishasthana:
So from some discussion, we settled on the following list for the file:

OWNERS:
Anand, Humair, Tom
Reviewers:
OWNERS + Anish, Hema, Harshad, Martin

PRs need 2 Reviews followed by 1 approval

anishasthana:
We can close this PR, https://github.com/operate-first/apps/pull/162 meets the same requirement.

PR #161, opened 2021-01-26
Add monitoring rbc for openshift-monitoring namespace

### LINKED ISSUE #160: Add the monitoring rbac to `openshift-monitoring` namespace
To be able to collect metrics exported from kube-state-metrics and the node-exporter, the `opf-monitoring` prometheus needs access to the services in the namespace.

related #60

4n4nd:
Closes #160 

once this is pushed, we can add servicemonitors to monitor kube-state-metrics and node-exporter metrics

ping @tumido @larsks

### DIFF
--- a/cluster-scope/base/namespaces/openshift-monitoring/kustomization.yaml
+++ b/cluster-scope/base/namespaces/openshift-monitoring/kustomization.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: openshift-monitoring
+
+components:
+  - ../../../components/monitoring-rbac
--- a/cluster-scope/overlays/moc/kustomization.yaml
+++ b/cluster-scope/overlays/moc/kustomization.yaml
@@ -32,6 +32,7 @@ resources:
   - ../../base/namespaces/opf-jupyterhub
   - ../../base/namespaces/opf-kafka
   - ../../base/namespaces/opf-monitoring
+  - ../../base/namespaces/openshift-monitoring
   - ../../base/namespaces/opf-observatorium
   - ../../base/namespaces/opf-superset
   - ../../base/namespaces/thoth-amun-api-prod
--- a/cluster-scope/overlays/moc/kustomization.yaml
+++ b/cluster-scope/overlays/moc/kustomization.yaml
@@ -5,6 +5,8 @@ kind: Kustomization
 resources:
   - ../../base/clusterroles/prow.k8s.io
   - ../../base/clusterloggings
+  - ../../base/clusterroles/metrics
+  - ../../base/clusterrolebindings/metrics
   - ../../base/crds/prow.k8s.io
   - ../../base/groups/approved-users
   - ../../base/groups/argocd-admins
--- a/cluster-scope/base/clusterrolebindings/metrics/kustomization.yaml
+++ b/cluster-scope/base/clusterrolebindings/metrics/kustomization.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - metrics-clusterrolebinding.yaml
--- a/cluster-scope/base/clusterrolebindings/metrics/metrics-clusterrolebinding.yaml
+++ b/cluster-scope/base/clusterrolebindings/metrics/metrics-clusterrolebinding.yaml
@@ -1,0 +1,13 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: metrics
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: metrics
+subjects:
+- kind: ServiceAccount
+  name: prometheus-k8s
+  namespace: opf-monitoring
--- a/cluster-scope/base/clusterroles/metrics/kustomization.yaml
+++ b/cluster-scope/base/clusterroles/metrics/kustomization.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - metrics-clusterrole.yaml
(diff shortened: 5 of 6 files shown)


HumairAK:
Did you want to add this specifically to moc as well? if so you need to add it to the moc overlay kustomization

4n4nd:
> Did you want to add this specifically to moc as well? if so you need to add it to the moc overlay kustomization

yess updated :+1:

HumairAK (approved):

anishasthana (approved):

hemajv (approved):

HumairAK:
Just FYI putting this on hold since it adds new resources to what's typically a namespace only accessible by cluster-admin (openshift-monitoring), so I'd like everyone to get a chance to look at it.

4n4nd:
> Just FYI putting this on hold since it adds new resources to what's typically a namespace only accessible by cluster-admin (openshift-monitoring), so I'd like everyone to get a chance to look at it.

Also these might not be enough privileges for SA to read node-metrics, I am looking into what more we will need

4n4nd:
I am not sure if namespaces is the right place for the metrics clusterrole I kept it in the `openshift-monitoring` namespace dir because it isn't used by any other namespace, if you folks know of a better place I could keep it please lmk

tumido (changes requested):
> I am not sure if namespaces is the right place for the metrics clusterrole I kept it in the openshift-monitoring namespace dir because it isn't used by any other namespace, if you folks know of a better place I could keep it please lmk

@4n4nd I would definitely suggest to put this resource (and the binding) out of the namespaces. `ClusterRole` is unique on cluster scope as well as `ClusterRoleBinding`. There must be no other resource with the same name of the same kind for the cluster scoped resources. Therefore it can't be hidden within a namespace despite that being the only place where it's used. Because there might be somebody else defining a "unique" cluster role for their single namespace and accidentally choose the same name - we can't let that happen.

I'd suggest you to move it to `cluster-scope/base/<RESOURCE_KIND_PLURAL_NAME>/<RESOURCE_NAME>`, e.g.:
- `cluster-scope/base/clusterroles/metrics`
- `cluster-scrope/base/clusterrolebindings/metrics`

cluster-scope/base/namespaces/openshift-monitoring/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: openshift-monitoring
+
+resources:
+  - metrics-clusterrole.yaml
+  - metrics-clusterrolebinding.yaml
```
tumido, line 9: I've also noticed, you're defining a `cluster-scope/base/namespaces/<NAMESPACE_NAME>` without any `Namespace` kind resource. Why is that? IIf you are defining a namespace, a namespace resource should be present

tumido (changes requested):

cluster-scope/base/namespaces/openshift-monitoring/kustomization.yaml
HumairAK: Ignoring the fact that these resources should not be under a namespace (in agreement with your other comment), 

AFAIK this namespace gets created upon openshift install, so it's expected to be there. We probably should be careful over-writing these namespaces, a quick check shows that this particular namespace has specific labels and annotations that look to have been auto-generated.

4n4nd: so this is because this namespace is automatically created by openshift and I thought we should not mess with an `openshift-` namespace

4n4nd:
> > I am not sure if namespaces is the right place for the metrics clusterrole I kept it in the openshift-monitoring namespace dir because it isn't used by any other namespace, if you folks know of a better place I could keep it please lmk
> 
> @4n4nd I would definitely suggest to put this resource (and the binding) out of the namespaces. `ClusterRole` is unique on cluster scope as well as `ClusterRoleBinding`. There must be no other resource with the same name of the same kind for the cluster scoped resources. Therefore it can't be hidden within a namespace despite that being the only place where it's used. Because there might be somebody else defining a "unique" cluster role for their single namespace and accidentally choose the same name - we can't let that happen.
> 
> I'd suggest you to move it to `cluster-scope/base/<RESOURCE_KIND_PLURAL_NAME>/<RESOURCE_NAME>`, e.g.:
> 
>     * `cluster-scope/base/clusterroles/metrics`
> 
>     * `cluster-scrope/base/clusterrolebindings/metrics`

@tumido updated

cluster-scope/overlays/moc/kustomization.yaml
```diff
@@ -27,6 +27,7 @@ resources:
   - ../../base/namespaces/opf-jupyterhub
   - ../../base/namespaces/opf-kafka
   - ../../base/namespaces/opf-monitoring
+  - ../../base/namespaces/openshift-monitoring
```
tumido, line 30: Please include it as a:

```
../../base/clusterroles/...
../../base/clusterrolebindings/...
```

Unless we are deploying a namespace we shouldn't be pulling things into a namespace within the `base/namespaces/` folder and hiding things in there.

I'd rather be this explicit and referred to as "yes, we are really deploying a cluster role".

tumido (changes requested):

cluster-scope/overlays/moc/kustomization.yaml
4n4nd: I don't think I can do that, it still needs to populate the `namespace` param, which is done from the namespace directory kustomzation

HumairAK: clusterroles and clusterrolebindings don't need the namespace paramater, you have these: 

```
  - ../../../base/clusterrolebindings/metrics
  - ../../../base/clusterroles/metrics
```
In `base/openshift-monitoring/kustomization.yaml`. Remove it from there and include it in the `cluster-scope/overlays/moc/kustomization.yaml`

4n4nd: no for the namespace scoped monitoring rbac: https://github.com/operate-first/apps/tree/master/cluster-scope/components/monitoring-rbac

4n4nd: I have added the clusterrole/binding in the moc overlay kustomization

HumairAK (approved):

HumairAK:
@tumido feel free to  /unhold if your requested changes have been addressed

tumido (approved):
LGTM
@4n4nd @HumairAK yeah, I forgot that this is an Openshift-deployed namespace which we want to "just" bootstrap with RBAC. I like the final solution @4n4nd :+1:

PR #163, opened 2021-01-27
feat: Create m4d-system and m4d-blueprints namespaces

tumido:
Resolves: https://github.com/operate-first/support/issues/59

PR #165, opened 2021-01-27
fix: Onboarding script group and rbac conditions

tumido:
When implementing #163 I've noticed few issues with our onboarding script that are caused by the changes based in a https://github.com/operate-first/apps/pull/120#discussion_r559822617

### DIFF
--- a/hack/onboarding.sh
+++ b/hack/onboarding.sh
@@ -70,7 +70,7 @@ create_project_admin_rolebinding() {
 
 create_group() {
 	mkdir -p $APP_NAME/base/groups/$OWNER
-	cat <<-EOF >> $APP_NAME/base/groups/$OWNER/group.yaml
+	cat <<-EOF > $APP_NAME/base/groups/$OWNER/group.yaml
 	---
 	apiVersion: user.openshift.io/v1
 	kind: Group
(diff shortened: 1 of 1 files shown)


hack/onboarding.sh
```diff
@@ -98,12 +98,12 @@ fi
...
 create_namespace
 
-if [ ! -d $APP_NAME/components/rbac/$OWNER ]; then
+if [ ! -d $APP_NAME/components/project-admin-rolebindings/$OWNER ]; then
 	echo "RBAC for '$OWNER' group does not exist yet. Creating..."
 	create_project_admin_rolebinding
 fi
 
-if ! grep -q "name: $OWNER" "$APP_NAME/base/groups/groups.yaml"; then
+if ! grep -r -q "name: $OWNER" "$APP_NAME/base/groups/"; then
```
HumairAK, line 106: why does it need to be recursive? is this for an edge case I'm not seeing?

HumairAK (approved):
Tested, lgtm

hack/onboarding.sh
tumido: Each group is within a folder within the `base/groups` We can also use `find $APP_NAME/base/groups/ -name $OWNER -type d` or something... this was just a laziness on my end, it was a lesser change set. What would you prefer?

HumairAK: Why you didn't use the same logic that you use to create the group to target that particular file. I.e.
```
mkdir -p $APP_NAME/base/groups/$OWNER
cat <<-EOF >> $APP_NAME/base/groups/$OWNER/group.yaml
```
So we know the file is always $APP_NAME/base/groups/$OWNER/group.yaml
Can't we just point it there?

PR #172, opened 2021-01-27
fix storageclassname and redundancy policy

### LINKED ISSUE #170: Cluster Logging configuration issue with PVC and redundancy
1. For the PVC, if no storage class is set in the clusterlogging CR then by default it is set to an `emptyDir`, we need to set it manually [here](https://github.com/operate-first/apps/blob/master/cluster-scope/base/clusterloggings/clusterLogging.yaml#L21).
2. We also need to change the redundancy policy to `ZeroRedundancy` [here](https://github.com/operate-first/apps/blob/master/cluster-scope/base/clusterloggings/clusterLogging.yaml#L23), with `SingleRedundancy` the kibana server isn't able to communicate with the Elasticsearch instance.

cc @eranra

4n4nd:
Closes #170 
Related: https://github.com/openshift/cluster-logging-operator/issues/869

### DIFF
--- a/cluster-scope/base/clusterloggings/clusterLogging.yaml
+++ b/cluster-scope/base/clusterloggings/clusterLogging.yaml
@@ -15,12 +15,14 @@ spec:
       audit:
         maxAge: 7d
     elasticsearch:
-      nodeCount: 3
+      nodeCount: 1
       storage:
-        # TODO: If need to set specific storageClassName
-        # storageClassName: "<storage-class-name>"
+        # TODO: manually set the storageClassName
+        # Upstream Issue:
+        # https://github.com/openshift/cluster-logging-operator/issues/869
+        storageClassName: "ocs-storagecluster-cephfs"
         size: 200G
-      redundancyPolicy: "SingleRedundancy"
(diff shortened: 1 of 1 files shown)


cluster-scope/base/clusterloggings/clusterLogging.yaml
```diff
@@ -17,10 +17,12 @@ spec:
     elasticsearch:
       nodeCount: 3
```
eranra, line 18: change to  nodeCount: **1**

Tere is no need for 3 ES nodes at this point. This is aligned with "ZeroRedundancy"

eranra: @4n4nd thanks for the support from my PoV if this can merge post this small fix 👍

tumido (approved):
Except the node count, /lgtm

cluster-scope/base/clusterloggings/clusterLogging.yaml
4n4nd: updated

4n4nd:
@tumido should be good to merge now

Issue #171, opened 2021-01-27
Cluster Logging RBAC to be able to view cluster logs on kibana


4n4nd:
We need to figure out what kind of RBAC is required for a user to be able to view logs in [kibana](https://kibana-openshift-logging.apps.cnv.massopen.cloud/)

I am able to create indexes and see logs as an admin of the `openshift-logging` namespace, but @eranra was not able to see the logs with `view` privileges to the namespace

eranra:
Once we get that working, need to assign those credentials to the entire team under `open-aiops`.

@drbwa @durandom @tumido

Issue #168, opened 2021-01-27
Will Cluster Logger logging pick up our individual pod logs?


goern:
If my pod is writing to stdout, something like `{"component":"hook","event-GUID":"ce0242c0-60d4-11eb-97f4-45029f1fcbd1","event-type":"issue_comment","external-plugin":"needs-rebase","file":"prow/hook/server.go:227","func":"k8s.io/test-infra/prow/hook.(*Server).demuxExternal.func1","level":"info","msg":"Dispatched event to external plugin","severity":"info","time":"2021-01-27T19:21:07Z"}` is it picked up by cluster logging and where and I find it?

/kind question
/sig documentation

tumido:
yes it should be picked up by cluster logging. We're working on that bit here https://github.com/operate-first/apps/issues/171

tumido:
I see - this issue calls for documentation, doesn't it? :slightly_smiling_face: In that case.. yes, we still have to document this.

tumido:
FTR from the chat:

```
Marcel Hild
@Tom Coufal @Humair Khan do we have ocp monitoring installed in the CNV cluster? I.e. node-exporter and the default grafana dashboards

Tom Coufal
Yes it is. However it is not yet managed by operate-first. The namespace and the resources are in place, the access seems to be limited to cluster admins. I think this is comes from the initial install/setup by @Lars Kellogg-Stedman  we didn't yet fully got to that bit.

Tom Coufal
alertmanager-main-openshift-monitoring.apps.cnv.massopen.cloud
grafana-openshift-monitoring.apps.cnv.massopen.cloud
prometheus-k8s-openshift-monitoring.apps.cnv.massopen.cloud
thanos-querier-openshift-monitoring.apps.cnv.massopen.cloud
I'm getting forbidden myself on those routes, but they exist
```
we need to resolve the permissions by aggregating our rbac to that namespace similar to this:
https://github.com/operate-first/apps/blob/master/cluster-scope/base/namespaces/openshift-monitoring/kustomization.yaml

I'll follow up in a PR on that.

tumido:
@goern you should be able to track down your logs at  https://kibana-openshift-logging.apps.cnv.massopen.cloud/ for example. Does that work for you or you'd prefer Loki?

We need some consolidation on all of this metrics/monitoring. I'll start another issue for that.

tumido:
[...]

PR #175, opened 2021-01-28
wip: feat: allow access to cluster-logging to open-aiops and thoth groups

tumido:
Possible solution: https://github.com/operate-first/apps/issues/168 https://github.com/operate-first/apps/issues/171

This is a backup option in case this is not an elastic search health problem.

### DIFF
--- a/cluster-scope/base/namespaces/openshift-logging/kustomization.yaml
+++ b/cluster-scope/base/namespaces/openshift-logging/kustomization.yaml
@@ -9,3 +9,5 @@ resources:
 
 components:
   - ../../../components/project-admin-rolebindings/operate-first
+  - ../../../components/project-view-rolebindings/open-aiops
+  - ../../../components/project-view-rolebindings/thoth
--- a/cluster-scope/base/namespaces/openshift-monitoring/kustomization.yaml
+++ b/cluster-scope/base/namespaces/openshift-monitoring/kustomization.yaml
@@ -6,3 +6,6 @@ namespace: openshift-monitoring
 
 components:
   - ../../../components/monitoring-rbac
+  - ../../../components/project-admin-rolebindings/operate-first
+  - ../../../components/project-view-rolebindings/open-aiops
+  - ../../../components/project-view-rolebindings/thoth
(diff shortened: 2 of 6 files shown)


tumido:
this is just a possible solution. I'm not certain of it. @4n4nd WDYT>

4n4nd:
> this is just a possible solution. I'm not certain of it. @4n4nd WDYT?

I don't think this might work, as I stated in the issue @eranra already has this role but he still can't access the logs unfortunately :(

tumido:
> I don't think this might work, as I stated in the issue @eranra already has this role but he still can't access the logs unfortunately :(

And as I messaged you and @eranra I think that might be because he didn't try to logging out of Kibana and OpenShift first before trying to access it second time. Because that's the only time the permissions get propagated to the app - when the user is logging in. So he needs to log out first and when he logs in second time it finally picks up the view permissions.

eranra:
@tumido I tried that, so I might have done something wrong in the process of logout/in or this is something else. Maybe we can find time tomorrow to continue are get this to work

tumido:
@eranra I don't think you did anything wrong. We probably need to do something with the RBAC. What namespaces you can see when you log in to the Openshift console/UI?

Based on the docs I think each account is limited to only the indices which maps to projects they do have access to.

https://docs.openshift.com/container-platform/4.3/logging/cluster-logging.html#cluster-logging-about-elasticsearch_cluster-logging

image


But even the things I see in Kibana with my account (I have access to 17 namespaces right now) doesn't seem to capture it all/in the right format:
image

@4n4nd @HumairAK any thoughts?

eranra:
image

eranra:
@tumido ^^^

Also, you should be able to see indexs that start with `infra*` those are from the infrastructure layer

tumido:
@eranra I see only indices that start with `app-*` nothing else. :confused: Somethings wrong there...

eranra:
@tumido I this that this is all connected to premissions ... in other OCP cluster where I am `cluster admin` I see both

tumido:
let me try playing with this. I suspect we need the `cluster-reader` cluster role assigned.

https://github.com/fabric8io/openshift-elasticsearch-plugin/issues/168

tumido:
Closing in favor of https://github.com/operate-first/apps/pull/194

PR #177, opened 2021-01-28
Add clusterrolebinding for opf prometheus to get clustermetrics

4n4nd:
According to the [openshift monitoring docs](https://docs.openshift.com/container-platform/4.5/monitoring/cluster_monitoring/prometheus-alertmanager-and-grafana.html#monitoring-accessing-prometheus-alertmanager-grafana-directly_accessing-prometheus), we need to add a clusterrolebinding for opf prometheus to be able to access (federate) metrics from openshift-monitoring prometheus

PR #178, opened 2021-01-29
chore: Fix metrics ClusterRoleBindings structure

tumido:
Fixes: https://github.com/operate-first/apps/pull/177


HumairAK:
hrmm so if we're doing one resource per directory, I have a couple of questions: 
1. Why have directories then and not just directly include the resource in the env kustomization? 
2. How is this different from having multiple roles in one directory like for [prow.k8s.io](https://github.com/operate-first/apps/tree/master/cluster-scope/base/clusterroles/prow.k8s.io) ?

tumido:
@HumairAK let's discuss it here https://github.com/operate-first/apps/pull/179

ad 1:
A kustomization file is required, since we are pulling it in from a `overlay`. We could have use `.yaml` in the overlay resources directly, but that wouldn't be consistent, since sometimes we need to apply `components` to the resources in the base. So you would end up with a mix of `resource_kind/resource_name.yaml` and `resource_kind/resource_name_folder` mix in the overlay...

ad 2:
I think we should change the prow. My motivation for this is that the cluster-scoped resource has to be unique in name. And if we enforce a naming convention that requires the name to be also a folder name, we can't have 2 resources of the same kind and name by default.

HumairAK:
as this is a relatively minor change I think one review is enough
/approve

Issue #164, opened 2021-01-27
Order of resources under overlays/moc/kustomization.yaml


eranra:
Hi @tcoufal with regards to the file https://github.com/operate-first/apps/blob/master/cluster-scope/overlays/moc/kustomization.yaml ... I really think that lexicographic order of the `resource` files under might not be the correct way to go ... consider somehow to group them in logical order (maybe per subject, area) or maybe have another hierarchical file per subject ... not critical but will probably help.

Also, I was looking to add some references to where the information on deployment was taken from ... maybe as part of that file, and maybe outside somewhere

tumido:
> consider somehow to group them in logical order (maybe per subject, area)

Yeah, maybe we would need to group them by some key. The main point of the `cluster-scope` application though is not to "deploy set of projects" or "deploy all resources in a cluster", it should rather be a bootstrap repository we can use to set up the cluster resources. So we don't want to have any project specific resources like `Deployment`s, `Service`s or `Route`s in there at all... Those resources should be kept in the application/project specific folder/repository and pulled in to the cluster by ArgoCD as a separate application.

The `cluster-scope` application (the term "application" may be misleading here but I don't have a better word for it) is ops-centric, not dev-centric. It should be friendly to the ops users who take care of what gets deployed to a particular cluster - on the cluster administrator level. Considering this aspect is the main focus of the application - I think we should be grouping the resources by a resource kind rather than the subject.

In the end, I agree, that some grouping may be necessary (like to introduce some nesting) because we don't want that resource list in `overlay/X/kustomization.yaml` to grow indefinitely.

Grouping by subject is achieved at the application/project level, since each application is meant to be about a single project. On the other hand, the `cluster-scope` app purpose is to (mostly) define resources which have to be unique for the whole cluster 0 therefore grouping them or ordering them per subject doesn't make much sense to me - you need to be certain that the resource is pulled into the overlay only once, etc. 

> Also, I was looking to add some references to where the information on deployment was taken from ... maybe as part of that file, and maybe outside somewhere

Can you elaborate on that? Do you mean like additional comments and descriptions to a resource and such? If so, I think I'd rather see that part of the individual resources - like annotations or labels, since this appears to be to be orthogonal to the resource kinds and even namespaces and projects. Would something like that be acceptable solution?

@HumairAK do you have any thoughts on that?

(btw. `@tcoufal` is not my GH handle. :smile: - I had to double check this is not my account by some accident. It seems it isn't...)

[...]

PR #179, opened 2021-01-29
feat: Cluster-scope maintenance guide

tumido:
Resolves: https://github.com/operate-first/apps/issues/164

Related: https://github.com/operate-first/apps/pull/180 https://github.com/operate-first/apps/pull/178

### DIFF
--- a/cluster-scope/README.md
+++ b/cluster-scope/README.md
@@ -1,0 +1,33 @@
+# Cluster management application organization
+
+This is just a short addendum to [ADR 0009](https://www.operate-first.cloud/blueprints/blueprint/docs/adr/0009-cluster-resources.md) declaring the direct consequences and organization structure of this `cluster-scope` application. These are the best practises we expect every maintainer who does changes to this application to follow.
+
+## Application folder structure
+
+- `base`: Contains all the actual resource manifests which can be deployed to any clusters. Manifests are required to be stored in a format that can be applied to all the clusters (no cluster specific `.spec`s).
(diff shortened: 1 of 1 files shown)


tumido:
Btw, there might be another exception to this that might need mentioning and that's crds, since there may exist CRDs with the same name in different API groups so this structure may be wrong for CRDs. WDYT?

tumido:
I mean, the motivation is to enforce consistency... I'd really like us to crave for perfection in this case, because we need to gain trust that this way can be used to effectively manage clusters.

tumido:
Also: linking relevant discussion from https://github.com/operate-first/apps/pull/178#issuecomment-769793299

cluster-scope/README.md
```diff
@@ -0,0 +1,33 @@
+# Cluster management application organization
+
+This is just a short addendum to [ADR 0009](https://www.operate-first.cloud/blueprints/blueprint/docs/adr/0009-cluster-resources.md) declaring the direct consequences and organization structure of this `cluster-scope` application. This are the best practises we expect every maintainer who does changes to this application to follow.
```
HumairAK, line 3: > This are the best 

to

>These are the best

HumairAK:
> Btw, there might be another exception to this that might need mentioning and that's crds, since there may exist CRDs with the same name in different API groups so this structure may be wrong for CRDs. WDYT? 

This seems like a pretty niche case, I'm good with just doing something like: 

`cluster-scope/base/crds/prow.k8s.io/apigroupA/kustomization.yaml`
and:
`cluster-scope/base/crds/prow.k8s.io/apigroupB/kustomization.yaml` 
So we can stay somewhat consistent.

HumairAK:
Looks great, I would imagine in the future we can place CI guards against potential changes to downstream environments. 

/lgtm

PR #180, opened 2021-01-29
chore: Fix ClusterRole structure

tumido:
Related to: https://github.com/operate-first/apps/pull/178 https://github.com/operate-first/apps/pull/179


4n4nd:
Should we keep a single kustomization for the whole `clusterroles` directory?

HumairAK:
@4n4nd  no because we do not want clusters to import all roles, rather we want them to pick and choose which ones they want to include

tumido:
@4n4nd https://github.com/operate-first/apps/blob/master/cluster-scope/README.md

HumairAK:
@4n4nd if the moc kustomization gets over-crowded, we can break into subdirectories in the moc env folder

PR #182, opened 2021-01-29
Enable anonymous access to grafana

### LINKED ISSUE #59: Need a GrafanaDatasource resource for Prometheus
We have set up Grafana, but we need to add a datasource ([example resource](https://github.com/integr8ly/grafana-operator/blob/master/deploy/examples/datasources/Prometheus.yaml)) so that Grafana is connected to the monitoring prometheus instance.

4n4nd:
Add odh grafana datasource
This will enable anonymous users to see our dashboards and edit them, but they won't be able to save their changes.
So no need for people to login to view existing dashboards.

Closes #59

### DIFF
--- a/odh/base/monitoring/overrides/grafana-operator/base/grafana.yaml
+++ b/odh/base/monitoring/overrides/grafana-operator/base/grafana.yaml
@@ -1,3 +1,4 @@
+---
 apiVersion: integreatly.org/v1alpha1
 kind: Grafana
 metadata:
@@ -15,9 +16,11 @@ spec:
         users:
             allow_sign_up: false
             auto_assign_org: true
-            auto_assign_org_role: Admin
+            auto_assign_org_role: Viewer
+            viewers_can_edit: true
         auth.anonymous:
-            enabled: false
+            enabled: true
+            org_role: Viewer
         log:
             level: warn
             mode: console
--- a/odh/base/monitoring/kfdef.yaml
+++ b/odh/base/monitoring/kfdef.yaml
@@ -43,6 +43,7 @@ spec:
     - kustomizeConfig:
         overlays:
           - dashboards
+          - datasources
         repoRef:
           name: opf
           path: odh/base/monitoring/overrides/grafana-operator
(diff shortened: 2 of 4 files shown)


anishasthana (changes requested):
Please split this into two different  commits -- one for the viewer change and one for the new data source.

anishasthana (approved):
/lgtm

odh/base/monitoring/overrides/grafana-operator/base/grafana.yaml
```diff
@@ -15,9 +16,12 @@ spec:
         users:
             allow_sign_up: false
             auto_assign_org: true
-            auto_assign_org_role: Admin
+            auto_assign_org_role: Viewer
+            viewers_can_edit: true
         auth.anonymous:
-            enabled: false
+            enabled: true
+            org_name: Main Org.
```
tumido, line 23: Can we name this more fitting? So it really tells this is an anonymous user? It's just a nit pick.. :slightly_smiling_face: I leave this up to you

tumido (commented):
I think this PR resolves https://github.com/operate-first/apps/issues/59 issue, is that correct? Do we have an issue for the anonymous user?

HumairAK:
@4n4nd  can you add `fix #59`  comment : )

HumairAK:
@4n4nd can you also explain what the viewer permissions are doing in your PR description for the record

odh/base/monitoring/overrides/grafana-operator/base/grafana.yaml
4n4nd: I think this is the default group that grafana creates, but let me test

4n4nd: Yeah here we are not creating a new org, we are assigning the anon user to the existing default `Main org.`

4n4nd: I just removed it :+1

4n4nd:
there already exists a grafana datasource [upstream](https://github.com/opendatahub-io/odh-manifests/blob/master/grafana/grafana/base/datasource.yaml), we need to name it the same thing

PR #183, opened 2021-01-29
feat: Scale down and always fetch ODH Dashboard 

tumido:
This scale down the ODH Dashboard from unnecessary 2 replicas to a single one for now. I don't think there's any need for 2, they are idle all the time anyways.

Also specifies a `ImagePullPolicy` to `Always` so in case we update the `quay.io/operate-first/odh-dashboard:latest` openshift can update the image on a new pod rollout.


HumairAK:
@harshad16 you are the 2nd reviewer, can you review this : )

harshad16 (commented):
/lgtm

PR #184, opened 2021-01-29
Move odh grafana datasource from overlay to base

4n4nd:
#182 introduced a new grafana datasource for odh-prometheus, but there exists one in upstream. 
This should move the newly added grafana datasource to overrides/base so it overrides the [upstream resource](https://github.com/opendatahub-io/odh-manifests/blob/master/grafana/grafana/base/datasource.yaml).


HumairAK:
Hrm I'm not sure if we have a rule on using strategicpatchmerge in these sorts of changes or not, it seems like good practice to me to do that (e.g. [here](https://github.com/operate-first/apps/pull/183) ) but this is a fix for breaking a live change so I'm willing to delegate that discussion for later. 

/lgtm

4n4nd:
> Hrm I'm not sure if we have a rule on using strategicpatchmerge in these sorts of changes or not, it seems like good practice to me to do that (e.g. [here](https://github.com/operate-first/apps/pull/183) ) but this is a fix for breaking a live change so I'm willing to delegate that discussion for later.
> 
> /lgtm

@HumairAK there isn't a kustomization for a patch to be applied

PR #186, opened 2021-01-29
Add storage to opf prometheus to retain metrics for 2 weeks

4n4nd:
This should enable metrics to persist for 2 weeks

Just to be clear because the number of replicas are set to 2, this will create 2 10gig PVCs

I am planning to let this run over the weekend and see how much storage is used by these metrics for 2 days. Using that information, we can adjust the size of the PVC


HumairAK:
/unhold 
As per the comment, let's see how much storage we use over the weekend, then we can adjust the size. This is of course a temporary solution to using [thanos](https://github.com/operate-first/apps/issues/74).

PR #187, opened 2021-01-29
Add servicemonitor to federate metrics from cluser-monitor prometheus

4n4nd:
This will scrape the cluster monitoring prometheus every 60 seconds

### DIFF
--- a/odh/overlays/moc/monitoring/servicemonitors/cluster-metrics-servicemonitor.yaml
+++ b/odh/overlays/moc/monitoring/servicemonitors/cluster-metrics-servicemonitor.yaml
@@ -1,0 +1,29 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: cluster-monitor-federation
+  labels:
+    monitor-component: cluster-resources
+spec:
+  endpoints:
+    - bearerTokenSecret:
+        key: token
+        name: prometheus-k8s-token-g9n8s
+      honorLabels: true
+      params:
+        'match[]':
+          - '{__name__=~".+"}'
+      path: /federate
+      port: web
+      scheme: https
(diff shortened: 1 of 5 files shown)


odh/overlays/moc/monitoring/servicemonitors/cluster-metrics-servicemonitor.yaml
```diff
@@ -0,0 +1,29 @@
...
+kind: ServiceMonitor
+metadata:
+  name: cluster-monitor-federation
+  labels:
+    monitor-component: cluster-resources
+spec:
+  endpoints:
+    - bearerTokenSecret:
+        key: token
+        name: prometheus-k8s-token-g9n8s
```
tumido, line 12: This token secret is unique for each SA "instance" (hence the `-g8n8s` suffix). If we would redeploy the Service Account, this token secret name would change. You'd probably need to create a new SA secret with a static name and refer to that one.

https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#to-create-additional-api-tokens

```
apiVersion: v1
kind: Secret
metadata:
  name: a-new-static-name
  annotations:
    kubernetes.io/service-account.name: prometheus-k8s
type: kubernetes.io/service-account-token
```

tumido (changes requested):

odh/overlays/moc/monitoring/servicemonitors/cluster-metrics-servicemonitor.yaml
4n4nd: @tumido I have added a secret in the secrets dir can you give it a look?

tumido (commented):
@4n4nd I think that should do the trick :+1: 

/lgtm one more pair of eyes please

PR #188, opened 2021-01-31
Increase Prometheus Storage and reduce Metrics Retention

4n4nd:
This should resolve issues:
https://github.com/operate-first/SRE/issues/38 to https://github.com/operate-first/SRE/issues/49

The reason these issues were created is because the current Prometheus instance is out of storage and cannot collect new metrics. Because the metrics for the services are absent, the alerts are triggered.


HumairAK:
/approve
Time critical

PR #191, opened 2021-02-01
Add Thoth Service Metrics dashboard

pacospace:
Related-To: https://github.com/thoth-station/thoth-application/pull/826

### DIFF
--- a/odh/overlays/moc/monitoring/dashboards/thoth-station/thoth-service-metrics.yaml
+++ b/odh/overlays/moc/monitoring/dashboards/thoth-station/thoth-service-metrics.yaml
@@ -1,0 +1,10 @@
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata:
+  name: thoth-service-metrics
+  labels:
+    app: grafana
+    component: thoth-station
+spec:
+  name: thoth-service-metrics
+  url: https://raw.githubusercontent.com/thoth-station/thoth-application/master/grafana-dashboard/overlays/cnv-prod/thoth-service-metrics.json
--- a/odh/overlays/moc/monitoring/dashboards/thoth-station/kustomization.yaml
+++ b/odh/overlays/moc/monitoring/dashboards/thoth-station/kustomization.yaml
@@ -5,6 +5,7 @@ resources:
   - thoth-kafka-argo-metrics.yaml
   - thoth-knowledge-graph-content-metrics.yaml
   - thoth-postgresql-metrics.yaml
+  - thoth-service-metrics.yaml
   - thoth-sli-slo.yaml
   - thoth-user-api-metrics.yaml
   - thoth-workflow-controllers-metrics.yaml


odh/overlays/moc/monitoring/dashboards/thoth-station/thoth-service-metrics.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata:
+  name: thoth-user-api-metrics
+  labels:
+    app: grafana
+    component: thoth-station
+spec:
+  name: thoth-user-api-metrics
```
tumido, line 9: I think (?) Grafana requires this name to contain the file type suffix (`.json`), because it translates to a filename which is then mounted to Grafana:

https://github.com/operate-first/apps/blob/master/docs/odh/grafana/add_grafana_dashboard.md#add-a-grafana-dashboard-to-operatefirst-grafana
https://github.com/integr8ly/grafana-operator/blob/master/documentation/dashboards.md

tumido (changes requested):

odh/overlays/moc/monitoring/dashboards/thoth-station/thoth-service-metrics.yaml
pacospace: Thanks @tumido, then I have to update all other dashboards as well. see for example: https://github.com/operate-first/apps/blob/7fac10d12109b0b53d9750bdb9c6269ce3bcc0d9/odh/overlays/moc/monitoring/dashboards/thoth-station/thoth-knowledge-graph-content-metrics.yaml#L9 and all others Thoth related: https://github.com/operate-first/apps/tree/master/odh/overlays/moc/monitoring/dashboards/thoth-station cc @harshad16

tumido: Oh, does work just fine for the other dashboards without the suffix? Maybe I interpret it wrong and you can leave it.. :smile:

pacospace: Even though it appears weird to have requirement to put `.json` in the name, dashboards can be shared in other formats in Grafana: https://github.com/integr8ly/grafana-operator/blob/faceedabf2864402cc45767fe44ccd56e1b7955a/pkg/apis/integreatly/v1alpha1/grafanadashboard_types.go#L20

pacospace: Maybe we need to improve doc upstream to grafana operator and operate first doc consequently :D

pacospace: https://github.com/integr8ly/grafana-operator/issues/339

HumairAK: Seems like this request is no longer needed, @tumido  can you resolve it?

HumairAK:
The kustomize build on this directory fails: 
```
Error: accumulating resources: 2 errors occurred:
        * accumulateFile error: "merging resources from 'thoth-user-api-metrics.yaml': may not add resource with an already registered id: integreatly.org_v1alpha1_GrafanaDashboard|~X|thoth-user-api-metrics"
        * loader.New error: "error loading thoth-user-api-metrics.yaml with git: url lacks orgRepo: thoth-user-api-metrics.yaml, dir: got 
```
This is because `thoth-user-api-metrics.yaml` has the same resource kind/name.

tumido (approved):
:+1: I've been told to stop complaining about everything.. So be it. :smile:

odh/overlays/moc/monitoring/dashboards/thoth-station/thoth-service-metrics.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata:
+  name: thoth-user-api-metrics
```
tumido, line 4: Name doesn't match the file. I think you forgot to change this

tumido (changes requested):

odh/overlays/moc/monitoring/dashboards/thoth-station/thoth-service-metrics.yaml
tumido: Ah.. yes... @HumairAK is faster again.. https://github.com/operate-first/apps/pull/191#issuecomment-770904582

pacospace:
> The kustomize build on this directory fails:
> 
> ```
> Error: accumulating resources: 2 errors occurred:
>         * accumulateFile error: "merging resources from 'thoth-user-api-metrics.yaml': may not add resource with an already registered id: integreatly.org_v1alpha1_GrafanaDashboard|~X|thoth-user-api-metrics"
>         * loader.New error: "error loading thoth-user-api-metrics.yaml with git: url lacks orgRepo: thoth-user-api-metrics.yaml, dir: got 
> ```
> 
> This is because `thoth-user-api-metrics.yaml` has the same resource kind/name.

Yes thank you :) I forgot to send new commit, as I was waiting from grafana operator people to know about: https://github.com/integr8ly/grafana-operator/issues/339

anishasthana:
Could you squash the commits @pacospace  :-D

pacospace:
> Could you squash the commits @pacospace :-D

Why @anishasthana ? :D

anishasthana:
Only the first commit out of the three has a message that tells me anything useful. Since all three commits are for the same feature, it'd make sense for them to be one commit. 
image

pacospace:
> Only the first commit out of the three tells me anything useful. Since all three commits are for the same feature, it'd make sense for them to be one commit.
> image

The feature is in the PR title for me, which can be made by different commits (you don't have to add only a dashboard, but link it in kustomize and also you have to update all data not just link in this case) :) Moreover, different changes might come in a PR based on comments, I prefer to see the commit changes in that case. Moreover for others to learn and to easily see what changes have been updated, but that is personal I think :)

anishasthana:
Re: different changes based in a PR based on comments, I'm a fan of [fixup](https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html) for PR reviews. I'm probably being overly pedantic here though, I like keeping the commit history clean in general.  Since I'm not an approver here though I'm not going to block on it :-)

HumairAK:
I agree with @anishasthana here. In my opinion the pros of squashing commits outweighs the cons. This is a single change, so it should be reflected in a single commit "Add thoth service metrics dashboard". This makes debugging easier/faster if we're trying to find the source of a change that introduced a bug. It keeps the history clean without introducing bloated commits like "update previous commit by fixing a typo". It also makes it easier to handle merge conflicts (since you're dealing with a lot fewer conflicting commits).

harshad16 (commented):
/lgtm
thanks for this addition

harshad16:
I believe squashing the commits in this PR would be good, as this pr is focusing on a particular dashboard setup.
one commit with all the information is good as this is part of one feature addition. 
multiple commits make sense when each additional commit is corresponding to a different feature.

pacospace:
Thanks @harshad16 @anishasthana @HumairAK I won't argue more in this case as you are the maintainers here. So I follow your change requested in this case. 

Can you please provide a contributor doc specifying this for this project? Because as I mention it is a personal decision and we could discuss for some time about pros/cons but in this case, I'm a contributor here so I ll squash them, but maybe a PR template specyfing this might help others from the beginning :) Thanks!

tumido (approved):

Issue #193, opened 2021-02-01
Consolidation for logging and monitoring facilities


tumido:
Right now we have too many logging/monitoring facilities and it's confusing for users, since we provide insufficient guidance on how to access/view their traces and which ever systems are even available for the users. I've started collecting some links and resources, feel free to chip in your findings.

| Namespace       | Applications     | Scope       | Required permissions  |
| --- | --- | --- | --- |
| `openshift-monitoring` | [Prometheus](https://prometheus-k8s-openshift-monitoring.apps.cnv.massopen.cloud), [Grafana](https://grafana-openshift-monitoring.apps.cnv.massopen.cloud), [AlertManager](https://alertmanager-main-openshift-monitoring.apps.cnv.massopen.cloud)  | Cluster | Minimal is `cluster-monitoring-view` cluster role |
| `openshift-logging`    | ElasticSearch, FluentD, [Kibana](https://kibana-openshift-logging.apps.cnv.massopen.cloud/) | Cluster | Minimal is read on `pods/log` for applications, `cluster-reader` for OpenShift resources |
| `opf-observatorium`    | [Grafana](https://grafana-route-opf-observatorium.apps.cnv.massopen.cloud/), Loki | Selected namespaces | Public |
| `opf-monitoring`       | [Prometheus](http://prometheus-portal-opf-monitoring.apps.cnv.massopen.cloud/), [AlertManager](http://alert-manager-opf-monitoring.apps.cnv.massopen.cloud/), [Grafana](https://grafana-route-opf-monitoring.apps.cnv.massopen.cloud/) | Selected namespaces | Public |

I'm getting a bit lost in what app deployment has what purpose, so we need to properly document which app should be accessed/used for which purpose and when to interact with each..

/sig documentation

Related: https://github.com/operate-first/apps/issues/168 https://github.com/operate-first/apps/issues/171

tumido:
cc @hemajv @4n4nd @HumairAK

4n4nd:
@tumido as far as I know, to access `opf-monitoring`/`opf-observatoium` components you don't need any special roles and everything is open.

For `openshift-monitoring`, I think you need [`cluster-monitoring-view`](https://github.com/operate-first/apps/blob/master/cluster-scope/base/clusterrolebindings/metrics/cluster-metrics-federation-clusterrolebinding.yaml#L9) to be able to access the metrics.

anishasthana:
[...]

PR #194, opened 2021-02-01
feat: Grant cluster-logs-reader ClusterRole to user groups

tumido:
Replaces: https://github.com/operate-first/apps/pull/175
Related: https://github.com/operate-first/apps/issues/193
Resolves: https://github.com/operate-first/apps/issues/171

Grants read access to all `pods` and `pods/log` in all namespaces on a cluster. This PR unblocks @eranra and team to access `infra-*` and other apps in cluster logging ElasticSearch and [Kibana](https://kibana-openshift-logging.apps.cnv.massopen.cloud/)

~~Question is if we care that people have `cluster-reader` access... I vote "I don't care, let people see things".~~

I've end up creating a new clusterrole - in the end it turned out that it only requires `pods` and `pods/log` access on `default`, `kube-*` and `openshift-*` namespaces. Since there's no easy way how to take into account all these namespaces via `Role`s, I'm for granting this access on the cluster level.

### DIFF
--- a/cluster-scope/base/clusterrolebindings/cluster-logs-readers/clusterrolebinding.yaml
+++ b/cluster-scope/base/clusterrolebindings/cluster-logs-readers/clusterrolebinding.yaml
@@ -1,0 +1,14 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: cluster-logs-readers
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-logs-reader
+subjects:
+  - kind: Group
+    name: operate-first
+  - kind: Group
+    name: open-aiops
--- a/cluster-scope/base/clusterrolebindings/cluster-logs-readers/kustomization.yaml
+++ b/cluster-scope/base/clusterrolebindings/cluster-logs-readers/kustomization.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - clusterrolebinding.yaml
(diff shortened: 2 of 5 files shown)


tumido:
cc @durandom @HumairAK @4n4nd

4n4nd:
@tumido can you add in the description which teams are getting access to this role?

4n4nd:
> I vote "I don't care, let people see things".

++
/lgtm

anishasthana:
Does cluster reader give access to all secrets?
(Are there any secrets on the MOC cluster that can't be shared?)

tumido:
@anishasthana secrets are not available to this role, but I'll create a different role and trim the permissions down

HumairAK:
I think we should be careful with how we treat the `cluster-reader` role on openshift. As @tumido said, secrets are not included, but routes/configmaps are. Routes can contain tls cert info on Openshift (unfortunately) and some configmaps go against best practice and can include sensitive info and we should be cognisant of that. There may be other resources I'm not aware of. But the point is, we should as @tumido  says, try to see if we can re-construct the cluster-reader role with just the permissions required for enabling access to cluster logs and nothing else.

4n4nd:
@tumido maybe we should remove thoth from this rolebinding since they don't need access to the whole cluster logs, but only access to their own namespace logs

tumido:
putting this on hold for now and prioritizing other tasks. I'll follow up on this tomorrow.

tumido:
@HumairAK @anishasthana @4n4nd 

I've end up creating a new clusterrole - in the end it turned out that it only requires `pods` and `pods/log` access on `default`, `kube-*` and `openshift-*` namespaces. Since there's no easy way how to take into account all these namespaces via `Role`s, I'm for granting this access on the cluster level.

I'm also granting this access to `operate-first` and `open-aiops` groups only. WDYT?

tumido:
My account with this role:
image

anishasthana (commented):
/lgtm I think this is a great starting point!

HumairAK:
This is awesome stuff, thanks @tumido . We should consider sending this to upstream docs (provided it works w/o issues). 
/lgtm

HumairAK:
Changes were applied

PR #195, opened 2021-02-01
Add kubeflow component namespaces

4n4nd:
This will create 4 new namespaces:
1. kubeflow
2. knative-serving
3. tekton-pipelines
4. istio-systems

And add 2 existing namespaces:
1. cert-manager
2. kube-system

These namespaces are intended for installation of kubeflow and it's components: kubeflow-pipeline, istio, Tekton pipelines, knative

Related: https://github.com/operate-first/support/issues/62 https://github.com/operate-first/support/issues/63

### DIFF
--- a/cluster-scope/base/namespaces/cert-manager/kustomization.yaml
+++ b/cluster-scope/base/namespaces/cert-manager/kustomization.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: cert-manager
+
+components:
+  - ../../../components/project-admin-rolebindings/operate-first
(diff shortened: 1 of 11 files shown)


larsks (commented):
For something like the `kube-system` namespace, which is intrinsic to k8s, why does that need to be declared here?

4n4nd:
> For something like the `kube-system` namespace, which is intrinsic to k8s, why does that need to be declared here?

I don't think it needs to be declared here, it is here just to keep track of what namespaces we are using. I can remove the declaration just like what we do for [`openshift-*` namespaces](https://github.com/operate-first/apps/tree/e98e7e261416de230bd5d2befeaead27ef3ddfb1/cluster-scope/base/namespaces/openshift-monitoring)

4n4nd:
@larsks I have removed the namespace declaration for the existing `cert-manager` and `kube-system` namespaces

tumido (approved):
/approve

PR #201, opened 2021-02-02
feat: Add a Rook/Noobaa standalone application

tumido:
Related to: https://github.com/operate-first/apps/pull/131

Syncs down the Ceph provider from MOC CNV in `openshift-storage` namespace

### DIFF
--- a/ceph/base/storagecluster.yaml
+++ b/ceph/base/storagecluster.yaml
@@ -1,0 +1,24 @@
+---
+apiVersion: ocs.openshift.io/v1
+kind: StorageCluster
+metadata:
+  name: ocs-storagecluster
+spec:
+  manageNodes: false
+  monDataDirHostPath: /var/lib/rook
+  storageDeviceSets:
+    - count: 3
+      dataPVCTemplate:
+        spec:
+          accessModes:
+            - ReadWriteOnce
+          resources:
+            requests:
+              storage: 500Gi
+          storageClassName: localblock
+          volumeMode: Block
+      name: ocs-deviceset
+      placement: {}
+      portable: false
+      replica: 3
+      resources: {}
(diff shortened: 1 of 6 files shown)


HumairAK:
I'm seeing the following diff between what's on live (left) and this pr (right): 
image

Is `spec.version` and `spec.externalStorage` intentionally left out?

tumido:
@HumairAK I think this is again populated from the operator. @larsks is not defining either of those in the deployed manifest:

https://github.com/open-infrastructure-labs/moc-cnv-sandbox/blob/f161a3c315d8eafc0b06d9aa7d3e8900f9dfc5bf/roles/ocp/ocs/tasks/main.yml#L119-L142

HumairAK:
Hrmm in that case
/lgtm

but can we get a +1 from @larsks  as well

ceph/base/storagecluster.yaml
```diff
@@ -0,0 +1,25 @@
+---
+apiVersion: ocs.openshift.io/v1
+kind: StorageCluster
+metadata:
+  name: ocs-storagecluster
+  namespace: openshift-storage
```
4n4nd, line 6: @tumido Can we move the `namespace` to the kustomization please?

4n4nd: or just remove this, it is already specified in the kustomization

4n4nd (changes requested):
@tumido should we track the `openshift-storage` in our list of namespaces?

tumido:
> @tumido should we track the `openshift-storage` in our list of namespaces?

@4n4nd 
Please read the PR description and see the parent PR: https://github.com/operate-first/apps/pull/131

larsks:
@HumairAK here is your +1.

I'm curious to see how all this will work when bringing up a new cluster: e.g., trying to create a `StorageCluster` resource will fail before the OCS operator is installed; do we just wait for subsequent ArgoCD runs to converge?

ceph/base/storagecluster.yaml
tumido: good catch :+1:

PR #203, opened 2021-02-02
feat: Add a kubevirt standalone application

tumido:
Related to: https://github.com/operate-first/apps/pull/131

Syncs down the Kubevirt deployment and resources in `openshift-cnv` namespace

### DIFF
--- a/kubevirt-hyperconverged/README.md
+++ b/kubevirt-hyperconverged/README.md
@@ -1,0 +1,3 @@
+# Openshift CNV deployment
+
+https://github.com/kubevirt/hyperconverged-cluster-operator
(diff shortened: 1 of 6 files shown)


HumairAK:
This too seems like it's missing the `spec.version` from what's on `live` (left) and what's in the pr (right): 

image

tumido:
@HumairAK I think that spec is updated by the operator later on. Maybe we would have to ignore that spec in ArgoCD?

@larsks is not deploying any `.spec` content for this resource:

https://github.com/open-infrastructure-labs/moc-cnv-sandbox/blob/f161a3c315d8eafc0b06d9aa7d3e8900f9dfc5bf/roles/ocp/cnv/tasks/main.yml#L34-L43

HumairAK:
Hrmm in that case 
/lgtm

but can we get a +1 from @larsks  as well

larsks (approved):

PR #208, opened 2021-02-04
Add kubeflow-components

4n4nd:
This includes Istio/Knative/Kfpipeline/tekton
The main purpose of this kfdef is to deploy kf-pipeline and it's dependencies

The source of this kfdef is from [upstream kubeflow](https://github.com/kubeflow/manifests/blob/master/kfdef/kfctl_openshift.master.kfptekton.yaml)

Closes https://github.com/operate-first/support/issues/6 https://github.com/operate-first/support/issues/63


tumido (commented):
/lgtm

PR #217, opened 2021-02-06
Test kustomize build

goern:
This script could be use to test all subdirectories containing a `kustomization.yaml` and just run a `kustomize build` on them...

### DIFF
--- a/test-kustomize-build
+++ b/test-kustomize-build
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+set -o errexit
+trap 'echo "Aborting due to errexit on line $LINENO. Exit code: $?" >&2' ERR
+set -o errtrace
+set -o pipefail
+
+k=$(find . -name kustomization.yaml)
+
+for d in $k; do
+    echo checking $(dirname $d)
+    kustomize build --enable_alpha_plugins $(dirname $d) 2>&1 >/dev/null
+done
+
+#end.


goern:
In conjunction with https://github.com/thoth-station/thoth-application/blob/master/prow/overlays/cnv-prod/config.yaml#L278-L291 it could `kustomize build` on each PR.

seeAlso https://github.com/AICoE/aicoe-ci/issues/81

4n4nd:
@goern does this work for sops encrypted manifests? If so, how is the key for decryption provided?

4n4nd (changes requested):
Could you please clean up the commits?

test-kustomize-build
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
```
tumido, line 2: I'd add:

```suggestion
set -e
```

Otherwise I have no idea which one of the builds in the for loop failed, because stderr goes to `/dev/null` ... Or you can have some conditional `echo` or something...

4n4nd:
> @goern does this work for sops encrypted manifests? If so, how is the key for decryption provided?

has this question been answered already?

HumairAK:
> @goern does this work for sops encrypted manifests? If so, how is the key for decryption provided? 

@goern  -- can you elaborate on this question above ^ ?

goern:
> > @goern does this work for sops encrypted manifests? If so, how is the key for decryption provided?
> 
> @goern -- can you elaborate on this question above ^ ?

erm. idk. I think if will fail, as no key is passed in, can we push it into the `kustomize` call via ENV? if so, we can provide a secret containing the key

HumairAK:
> erm. idk. I think if will fail, as no key is passed in, can we push it into the kustomize call via ENV? if so, we can provide a secret containing the key 

So we would need to ad the gpg key as a secret in the namespace the job (that executes this script) runs. We can update this script to import that gpg first before running kustomize builds.

The namespace where these jobs will is `opf-ci-prow` correct? So we would need to add the secret there, so we would need to consider who has/will have access to this namespace going forward. 

@tumido  @4n4nd

anishasthana:
Is this prow instance also what is running stuff for https://github.com/AICoE/aicoe-cd?

4n4nd:
> so we would need to consider who has/will have access to this namespace going forward.

Yeah we will need to strictly limit access to that namespace

tumido:
What if we just ignore the secret resources in CI instead? I know the CI check wouldn't be "complete", but we can live with manual verification of the secret resources, I'd say.

Instead of installing `ksops` in the test pipeline's `kustomize` plugins, you can define a stub exec plugin yourself like this:

```sh
PLUGIN_ROOT=$HOME/.config/kustomize/plugin
mkdir -p $PLUGIN_ROOT/viaduct.ai/v1/ksops/
echo '#!/bin/sh' > $PLUGIN_ROOT/viaduct.ai/v1/ksops/ksops
chmod +x $PLUGIN_ROOT/viaduct.ai/v1/ksops/ksops

kustomize build --enable_alpha_plugins ....  # Just works! The secrets are ignored
```

Please backup your `ksops.so` before you try it out locally.

WDYT?

goern:
I think this is not how we can do it using the opf-toolbox container... see https://prow.operate-first.cloud/view/s3/ci-prow-artifacts--5de4780e-313e-46c0-8758-209adf0ea179/prow-logs/pr-logs/pull/thoth-station_thoth-application/883/kustomize-build/1359467484817657856

tumido:
Ok, then you can set a `KUSTOMIZE_PLUGIN_HOME` environment variable and define the plugin elsewhere:

```sh
export KUSTOMIZE_PLUGIN_HOME=$HOME/new_plugin_dir

mkdir -p $KUSTOMIZE_PLUGIN_HOME/viaduct.ai/v1/ksops/
echo '#!/bin/sh' > $KUSTOMIZE_PLUGIN_HOME/viaduct.ai/v1/ksops/ksops
chmod +x $KUSTOMIZE_PLUGIN_HOME/viaduct.ai/v1/ksops/ksops

kustomize build --enable_alpha_plugins ....  # Just works! The secrets are ignored
```

goern:
hmm, that seems to work... thx!

tumido:
@goern if I understand it properly, this script is rather for manual testing (when I have the keys) and the https://github.com/thoth-station/thoth-application/pull/883 + https://github.com/thoth-station/thoth-application/blob/master/test-kustomize-build is used as a PR check. Correct?

tumido (commented):
/lgtm

PR #219, opened 2021-02-08
feat: Disable registration flow in Kubeflow

tumido:
Temporary workaround until https://github.com/kubeflow/manifests/pull/1736 is merged

Resolves: https://github.com/operate-first/apps/issues/213

PR #220, opened 2021-02-08
Add a demo kubeflow profile

4n4nd:
since we disabled kubeflow registration in #219, I have added a demo profile with limited resources which can be accessed publicly

### DIFF
--- a/kubeflow-components/overlays/moc/profiles/demo-profile.yaml
+++ b/kubeflow-components/overlays/moc/profiles/demo-profile.yaml
@@ -1,0 +1,15 @@
+---
+apiVersion: kubeflow.org/v1beta1
+kind: Profile
+metadata:
+    name: kf-demo
+spec:
+    owner:
+        kind: Group
+        name: operate-first
+    resourceQuotaSpec:
+        hard:
+            cpu: "4"
+            memory: 8Gi
+            persistentvolumeclaims: "1"
+            requests.storage: "5Gi"
--- a/docs/kubeflow/kubeflow-dashboard/README.md
+++ b/docs/kubeflow/kubeflow-dashboard/README.md
@@ -1,0 +1,10 @@
+## Kubeflow Access
+[Kubeflow](https://www.kubeflow.org/docs/about/kubeflow/) is a machine learning toolkit for Kubernetes.
+
+An instance of Kubeflow is available for use in OperateFirst.
+
+You can access the Kubeflow dashboard with the following url.
+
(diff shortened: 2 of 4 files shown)


kubeflow-components/overlays/moc/profiles/demo-profile.yaml
```diff
@@ -0,0 +1,15 @@
+---
+apiVersion: kubeflow.org/v1beta1
+kind: Profile
+metadata:
+    name: kf-demo
+spec:
+    owner:
+        kind: User
+        name: asanmukh@redhat.com
```
tumido, line 9: we probably want to encrypt this field if it points to an user. Can this be set to a `Group` instead?

kubeflow-components/overlays/moc/profiles/demo-profile.yaml
```diff
@@ -0,0 +1,15 @@
+---
+apiVersion: kubeflow.org/v1beta1
+kind: Profile
+metadata:
+    name: kf-demo
```
tumido, line 5: This corresponds to the namespace created. Does it create the RoleBinding for the user?

tumido (changes requested):
I think we need to have some basic documentation in the repo, before we start adding additional namespaces (not this one, but for other users)... Can you please include a quick README into this PR as well?


https://www.kubeflow.org/docs/components/multi-tenancy/getting-started/#managing-contributors-manually

kubeflow-components/overlays/moc/profiles/demo-profile.yaml
4n4nd: Good idea :+1:

kubeflow-components/overlays/moc/profiles/demo-profile.yaml
4n4nd: It does create a rolebinding for the user. It  gives a role called `kubeflow-admin` ([link](https://github.com/kubeflow/manifests/blob/f95406b14a2025db1e8b88d72caa1316ec5d1f49/kubeflow-roles/base/cluster-roles.yaml#L22))

4n4nd:
> before we start adding additional namespaces

Yeah we will need to figure out authentication before that, I have just added this namespace/profile so we can limit resource usage

tumido (commented):
/lgtm Awesome stuff! :+1: :1st_place_medal:

tumido:
it's still on hold, even though @4n4nd /unhold it. Let me try once more.

tumido (approved):

PR #224, opened 2021-02-09
Add PVC manifest for user mcli

4n4nd:
Adds an encrypted PVC manifest to be used in the `opf-jupyterhub` namespace by Jupyterhub.

Closes https://github.com/operate-first/support/issues/79

### DIFF
--- a/odh/overlays/moc/jupyterhub/.sops.yaml
+++ b/odh/overlays/moc/jupyterhub/.sops.yaml
@@ -1,0 +1,3 @@
+creation_rules:
+  - encrypted_regex: "^annotations|name$"
+    pgp: "0508677DD04952D06A943D5B4DC4116D360E3276"
--- a/odh/overlays/moc/jupyterhub/kustomization.yaml
+++ b/odh/overlays/moc/jupyterhub/kustomization.yaml
@@ -5,6 +5,10 @@ namespace: opf-jupyterhub
 
 resources:
   - ../../../base/jupyterhub
+
+generators:
+  - pvcs/secret-generator.yaml
+
 patchesJson6902:
   - patch: |
       - op: add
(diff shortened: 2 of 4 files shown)


4n4nd:
@HumairAK @tumido @anishasthana does using the first 4 characters of the username in the manifest filename make sense? or do you folks have any better ideas on what the manifest filenames should be?

tumido:
You can consider stashing all the PVC resources into a single file. Like a multidocument yaml, containing resources with encrypted names... But I don't mind seeing few letters of a name in the filename though.

tumido (commented):
/lgtm

4n4nd:
> You can consider stashing all the PVC resources into a single file. Like a multidocument yaml, containing resources with encrypted names... But I don't mind seeing few letters of a name in the filename though.

I thought about a multidocument yaml, but I thought that would be messy and this way the users themselves might be able to change their PVC sizes.

tumido:
@4n4nd they can't change it themselves anyways, because even if you change an unencrypted part of a sops-encrypted file, ksops/sops will complain on decryption, because the checksum differs. So once a resource is encrypted, you can only edit it in the decrypted state.

You're right about a single yaml (multi-doc) being messy though. :slightly_smiling_face:

4n4nd:
> @4n4nd they can't change it themselves anyways, because even if you change an unencrypted part of a sops-encrypted file, ksops/sops will complain on decryption, because the checksum differs. So once a resource is encrypted, you can only edit it in the decrypted state.
> 
> 
> 
> You're right about a single yaml (multi-doc) being messy though. :slightly_smiling_face: 

ah that would make sense 👍

anishasthana (approved):
/lgtm

4n4nd:
This was confirmed working

PR #230, opened 2021-02-09
Add curator config for CLO

4n4nd:
Set default log retention to 7 days. 
Docs: https://docs.openshift.com/container-platform/4.5/logging/config/cluster-logging-curator.html#cluster-logging-curator-yaml_cluster-logging-curator

Currently the 200Gig PVC for cluster logging is at >90% usage. This is resulting in Elasticsearch not pushing in any new logs.

Related #https://github.com/operate-first/support/issues/83

### DIFF
--- a/cluster-logging/overlays/moc/configmaps/curator.yaml
+++ b/cluster-logging/overlays/moc/configmaps/curator.yaml
@@ -1,0 +1,79 @@
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: curator
+data:
+  actions.yaml: |
+    # ---
+    # Remember, leave a key empty if there is no value.  None will be a string,
+    # not a Python "NoneType"
+    #
+    # Also remember that all examples have 'disable_action' set to True.  If you
+    # want to use this action as a template, be sure to set this to False after
+    # copying it.
+    # actions:
+    #   1:
+    #     action: delete_indices
+    #     description: >-
+    #       Delete .operations indices older than 30 days.
+    #       Ignore the error if the filter does not
+    #       result in an actionable list of indices (ignore_empty_list).
+    #       See https://www.elastic.co/guide/en/elasticsearch/client/curator/5.2/ex_delete_indices.html
+    #     options:
+    #       # Swallow curator.exception.NoIndices exception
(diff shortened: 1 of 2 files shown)


cluster-logging/overlays/moc/configmaps/curator.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+    #       unit: days
+
+    #       unit_count: 30
+
+    #       exclude: False
+  config.yaml: |
+    # Logging example curator config file
+
+    uncomment and use this to override the defaults from env vars
```
HumairAK, line 87: Can you remove this line, I believe it was meant to be a comment :)

4n4nd: good catch :+1: 
updated

cluster-logging/overlays/moc/configmaps/curator.yaml
```diff
@@ -0,0 +1,117 @@
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: curator
+data:
+  actions.yaml: >
+    # ---
+
```
tumido, line 9: can you please remove all those empty lines, please? It just bloats the resource...

tumido (changes requested):

cluster-logging/overlays/moc/configmaps/curator.yaml
larsks: The blank lines are there because the content is using the folding block operator (`>`) instead of the literal block operator (`|`). 

This:

```
value: >
  one

  two

  three
```

Is equivalent to:

```
value: |
  one
  two
  three
```

Probably best just to use the literal block operator instead:

```
data:
  actions.yaml: |
    # ---
    # Remember, leave a key empty if there is no value. None will be a string...
```

tumido (commented):
/lgtm

cluster-logging/overlays/moc/configmaps/curator.yaml
4n4nd: updated :+1:

HumairAK:
Configmap updated

Issue #231, opened 2021-02-10
ODH 1.0 upgrade path


tumido:
Upgrade ODH components to 1.0.0 release

- [x] Operator upgrade - automated approval, already updated (cleanup was needed https://github.com/operate-first/apps/pull/234)
- [x] Argo (https://github.com/operate-first/apps/pull/233)
- [x] Dashboard (https://github.com/operate-first/odh-dashboard/issues/7, https://github.com/operate-first/odh-dashboard/pull/8)
- [x] Data catalog (https://github.com/operate-first/apps/pull/236)
- [x] Jupyterhub (https://github.com/operate-first/apps/pull/238)
- [x] Kafka (https://github.com/operate-first/apps/pull/235)
- [x] Monitoring (https://github.com/operate-first/apps/pull/239)
- [x] Superset (https://github.com/operate-first/apps/pull/237)

HumairAK:
For reference these are the patch notes: 

 - Update JupyterHub image to 0.1.5
 - Update Argo to v2.12.5
 - Add Thriftserver to data catalog
 - Add Hue to data catalog
 - Kafka dashboard: migrate singlestat to new stat panel
From [here](https://github.com/opendatahub-io/odh-manifests/releases)

tumido:
>  - Update Argo to v2.12.5
> - Add Thriftserver to data catalog
> - Add Hue to data catalog

We already used those from ODH `master`, so the only changes that applies to us are basically:

- Update JupyterHub image to 0.1.5
- Kafka dashboard: migrate singlestat to new stat panel

HumairAK:
Kafka kfdef updated

HumairAK:
Data catalog kfdef updated

HumairAK:
Argo kfdef Updated

HumairAK:
Superset kfdef updated: 

```
time="2021-02-10T14:10:56Z" level=info msg="Successfully applied application superset"
```

tumido:
Superset `DeploymentConfig`'s rollout is unable to complete due to PVC `readWriteOnce` (while the previous replica is trying to access it still)

![image](https://user-images.githubusercontent.com/7453394/107521254-9c908600-6bb2-11eb-9200-838598415d97.png)

HumairAK:
Killing old pods

anishasthana:
[...]

PR #233, opened 2021-02-10
feat: ODH v1.0.0 upgrade: Argo

tumido:
Part of: https://github.com/operate-first/apps/issues/231

Tested locally, there's no change in manifests, since we were already using the master for argo


4n4nd (commented):
/lgtm

PR #234, opened 2021-02-10
feat: Relocate ODH operator group

tumido:
Seems we had a bit of a mess where the ODH operator is defined. This should clean it up. Subscription was already present in the `cluster-scope` app. There's no ODH operator specific ArgoCD app to delete as far as I know of. 

Related to: https://github.com/operate-first/apps/issues/231


HumairAK:
So I'm looking at moc, and odh is installed in openshift operators, I'm wondering if that old deployment was ever in use.

tumido:
Ah,. that's a good point. Yeah, in that case we don't need the namespace and the operator group. That's right :+1:  I'll update this PR.

tumido:
@HumairAK updated :+1: That was a good call, thank you!

4n4nd (commented):
/lgtm

HumairAK:
changes applied

PR #242, opened 2021-02-10
fix kubeflow-pipelines docs directory name

4n4nd:
Directory was incorrectly in named in https://github.com/operate-first/apps/pull/220


durandom (approved):
APPRV / LGTN

tumido (approved):

durandom:
I'm merging, because I can 😈 

...and I need the fixed link

Issue #65, opened 2020-12-08
Deploy authenticator namespace with DEX


tumido:
Since many of the components require OpenID and Openshift itself can't provide it, we need a translation layer - DEX. Let's deploy one instance instead of a separate for each component.

Related to #37 #39 #47 #48

PR #248, opened 2021-02-11
feat: Create namespace for Dex

tumido:
Part of: https://github.com/operate-first/apps/issues/65

Preparation step, let's have the namespace first.

PR #249, opened 2021-02-11
feat: Add auth via DEX OpenID

tumido:
Resolves: https://github.com/operate-first/apps/issues/65
Requies: https://github.com/operate-first/apps/pull/248

Implements base secrets for Argo, Grafana, Grafana2 and Hue for now (for a quicker onboarding, the apps need to turn on SSO anyways)


auth/base/dex-cm.yaml
```diff
@@ -0,0 +1,61 @@
...
+          - http://URL/oauth2/callback
+        secretEnv: ARGO_SECRET
+
+      - id: grafana
+        name: Grafana
+        redirectURIs:
+          - https://URL/login/generic_oauth
+        secretEnv: GRAFANA_SECRET
+
+      - id: grafana_observatorium
```
4n4nd, line 38: @tumido can we remove observatorium grafana from this? It already has login disabled and we don't need it to be enabled.

tumido: I'm cool with that, np, I'll remove it

HumairAK (approved):
/lgtm
/approve

Let's try it!

PR #252, opened 2021-02-11
Update CLO nodecount to 3 (odd number)

4n4nd:
As @HumairAK suggested we should have a odd number of nodes to prevent complexities during master/worker elections.
I have updated the CLO elasticsearch nodecount to 3


anishasthana (commented):
/lgtm

Issue #229, opened 2021-02-09
Allow the ability to externally access elasticsearch


HumairAK:
For diagnosing/debugging this can be useful, the process is laid out here to do it securely: 

https://docs.openshift.com/container-platform/4.5/logging/config/cluster-logging-log-store.html#cluster-logging-elasticsearch-exposing_cluster-logging-store

anishasthana:
While logging in as the Elasticsearch Service Account is one way to access the contents (you could argue this is a good thing, since it forces you to deliberately log in as the SA), a better way might be to configure searchguard to allow the opf-admins group to have these rights. I also don't see the harm in opening up get access for anyone who has access to the namespace. I haven't used the logging operator before so I don't know what flags you'd be tweaking, but it's definitely possible to set permissions for interacting with the ES api.

HumairAK:
> a better way might be to configure searchguard to allow the opf-admins group to have these rights 

Yes I imagine we'll have to do something along these lines.  We probably will need to be able allow other users to use their ocp tokens as bearer tokens to query elasticsearch api endpoint.

drbwa:
> We probably will need to be able allow other users to use their ocp tokens as bearer tokens to query elasticsearch api endpoint.

Exactly. There are probably use cases where a user might want to retrieve logs programmatically (i.e. from ES). To feed logs into their own alerting and anomaly detection logic, to try and determine whether an issue is likely to be due to a fault in their own system or rather due to a dependency. I don't know to what extent all logs can be made available to all users. For our specific use case, the more logs of different components we have access to, the better the coverage or ability to detect issues of our tool is going to be.

HumairAK:
Okay so I went down the rabbithole that is opendistro for elasticsearch (which is what it seems like CLO uses). Will share my findings here: 

- Elasticsearch that is provisioned by ES operator (provisioned by CLO) has an [oauth proxy](https://github.com/openshift/elasticsearch-proxy) infront of it
- The proxy is configured in such a way that it will assign the `x-forwarded-roles` header one of 2 roles, they are: 
    - admin_reader : this is assigned if the user accessing the endpoint has the following access: `{"namespace": "default", "verb":
          "get", "resource": "pods/log"}` 
    - project_user : applied by default if `admin_reader` is not applicable to this user
[...]

PR #259, opened 2021-02-16
Add route for elasticsearch external access.

HumairAK:
This is related to the issue [here](https://github.com/operate-first/apps/issues/229). Though more work needs to be done to allow individual users that are not cluster-admin to be able to use it.


tumido (commented):
/lgtm

HumairAK:
Changes applied -- route was successfully added

PR #264, opened 2021-02-17
Remove odh-operator namespace from crc/quicklab

HumairAK:
Atm these builds are broken due to the odh-operator being recently removed.


anishasthana (approved):
/lgtm

Issue #211, opened 2021-02-05
Migrate Post-install CNV steps to cluster-scope application


tumido:
Finalize migration of post-install CNV setup to ArgoCD.
https://github.com/open-infrastructure-labs/moc-cnv-sandbox/blob/master/playbook-postinstall.yml

Migrate content from:
- [ ] https://github.com/open-infrastructure-labs/moc-cnv-sandbox/blob/master/roles/ocp/authz/tasks/main.yml
- [ ] https://github.com/open-infrastructure-labs/moc-cnv-sandbox/blob/master/roles/ocp/tls/tasks/main.yml
- [ ] https://github.com/open-infrastructure-labs/moc-cnv-sandbox/blob/master/roles/ocp/moc-sso/tasks/main.yml
- [ ] https://github.com/open-infrastructure-labs/moc-cnv-sandbox/blob/master/roles/ocp/registry/tasks/main.yml

Investigate if the following needs migration as well:
- [ ] https://github.com/open-infrastructure-labs/moc-cnv-sandbox/blob/master/roles/ocp/bmh/tasks/main.yml
- [ ] https://github.com/open-infrastructure-labs/moc-cnv-sandbox/blob/master/roles/ocp/firewall/tasks/main.yml
- [ ] https://github.com/open-infrastructure-labs/moc-cnv-sandbox/blob/master/roles/ocp/hostpath/tasks/main.yml
- [ ] https://github.com/open-infrastructure-labs/moc-cnv-sandbox/blob/master/roles/ocp/public-net/tasks/main.yml

/sig devops

HumairAK:
Since the new cluster is no longer using the moc-cnv-sandbox repo, I think this is stale, closing. Please re-open if there's more to be done.

PR #269, opened 2021-02-18
feat: Sync the authz tasks

tumido:
First part of: https://github.com/operate-first/apps/issues/211

Migrate content from:
- [x] https://github.com/open-infrastructure-labs/moc-cnv-sandbox/blob/master/roles/ocp/authz/tasks/main.yml, I've skipped:
  - `Group` definitions
  - `nad-view-user` `ClusterRole` and `self-provisioners`, `allow-view-nads` `ClusterRoleBindings` since it's related to the `approved-users` group, which was deleted
- [x] https://github.com/open-infrastructure-labs/moc-cnv-sandbox/blob/master/roles/ocp/tls/tasks/main.yml
- [x] https://github.com/open-infrastructure-labs/moc-cnv-sandbox/blob/master/roles/ocp/moc-sso/tasks/main.yml
- [x] https://github.com/open-infrastructure-labs/moc-cnv-sandbox/blob/master/roles/ocp/registry/tasks/main.yml


cluster-scope/base/imageregistryconfigs/cluster/config.yaml
```diff
@@ -0,0 +1,12 @@
+---
+apiVersion: imageregistry.operator.openshift.io/v1
+kind: Config
+metadata:
+  name: cluster
```
HumairAK, line 5: This looks very different from the one in live: 

```yaml
apiVersion: operator.openshift.io/v1
kind: Config
metadata:
  annotations:
    release.openshift.io/create-only: "true"
  name: cluster
spec:
  managementState: Managed
status: {}
```

tumido: that's a different `Config` kind, see the API group:
`operator.openshift.io/v1`'s`Config`  vs `imageregistry.operator.openshift.io`'s `Config`.

you'd need to query it as:
```
oc get configs.imageregistry.operator.openshift.io cluster
```

It confused me as well.

HumairAK: gotcha thanks!

HumairAK:
@tumido 
image

We could put some of these fields in, but I'm guessing `httpsecret` will need to be ignored in the cluster-resource app -- wdyt?

PR #270, opened 2021-02-22
Remove kubeflow managed namespaces.

HumairAK:
Currently kubeflow will over-write these apps, thus fighting with argocd to sync them. We cannot keep an exact copy because argocd will always add a new label which it uses to match what argocd app the resource belongs to. So just removing them all together for now.

image


tumido (approved):
/approve

PR #271, opened 2021-02-22
feat: MOC SSO for moc-infra

tumido:
clientId and clientSecret to be filled in by @larsks


larsks:
@tumido I added the client secret and updated the PR.

PR #273, opened 2021-02-22
Give ArgoCD SA argocd-manager cluster admin

HumairAK:
The `argocd-manager` SA is what argocd will be using to deploy apps to both `moc-cnv` and `moc-infra` clusters, as such this SA exists in both clusters and will require cluster admin in both of these clusters.

[Related](https://github.com/operate-first/SRE/issues/83).


tumido (approved):
/approve

PR #276, opened 2021-02-23
Add alerting rule for unusually high scrape durations in cluster-monitoring

4n4nd:
Addresses https://github.com/operate-first/SRE/issues/80

### DIFF
--- a/odh/overlays/moc/monitoring/rules/alerting/openshift-monitoring-scrape-duration.yaml
+++ b/odh/overlays/moc/monitoring/rules/alerting/openshift-monitoring-scrape-duration.yaml
@@ -1,0 +1,15 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: openshift-monitoring-scrape-duration-alerts
+spec:
+  groups:
+    - name: Operate First Cluster Monitoring
+      rules:
+        - alert: Operate First Cluster Monitoring High Scrape Duration
+          expr: scrape_duration_seconds{namespace="openshift-monitoring"} > 80
+          for: 5m
+          annotations:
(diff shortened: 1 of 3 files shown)


odh/overlays/moc/monitoring/rules/alerting/openshift-monitoring-scrape-duration.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: PrometheusRule
+metadata:
+  name: openshift-monitoring-scrape-duration-alerts
+spec:
+  groups:
+    - name: Operate First Cluster Monitoring
+      rules:
+        - alert: Operate First Cluster Monitoring High Scrape Duration
+          expr: scrape_duration_seconds{namespace="openshift-monitoring"} > 80
+          for: 5m
```
anishasthana, line 12: Do we want to change this to 10m instead of 5? Just to reduce the likelihood of networking blips resulting in noise.

4n4nd: @hemajv what do you suggest? I say let's keep all alerts at 5m for now, later we can update them if need be.

anishasthana: I believe she's on PTO for the next two weeks

HumairAK: I think we would eventually want to raise to 10m, but I think 5m is fine for now, we just  changed the timeouts recently, let's monitor them to see how they increase and react accordingly based on data.

4n4nd:
deployed :+1:

PR #277, opened 2021-02-23
Replace user with different account for argocd admins.

HumairAK:
@larsks  this will add your bu account as argocd admin instead of your redhat acount.


4n4nd (commented):
/lgtm

tumido (approved):
/approve

Issue #278, opened 2021-02-23
moc-sso-client-secret being copied to openshift-authentication namespace forcing argocd to prune it


HumairAK:
The `moc-sso-client-secret` found [here](https://github.com/operate-first/apps/blob/master/cluster-scope/overlays/moc/oauths/moc-sso-client-secret.enc.yaml) gets a copy created in the `openshift-authentication` namespace, with the copy being called `v4-0-config-user-idp-0-client-secret` (creatd by the auth operator?). 

This means that the label that argocd adds to `moc-sso-client-secret.enc.yaml` to match the secret to it's argocd application also gets copied to `v4-0-config-user-idp-0-client-secret`. But since `v4-0-config-user-idp-0-client-secret` is not part of the kustomize build, ArgoCD then thinks this secret should be pruned from the application because it's not actually a part of it. 

Workaround:
Add the following annotations to `v4-0-config-user-idp-0-client-secret`

```yaml
    argocd.argoproj.io/compare-options: IgnoreExtraneous
    argocd.argoproj.io/sync-options: Prune=false
```

This basically keeps the sync status "green" despite this one resource being outofsync (IgnoreExtraneous), and it will skip pruning this resource when a sync does happen (Prune=false). 

My issue with this is that we are not tracking this change anywhere, and on a fresh cluster install we'd have to perform this manual step. Thoughts?

HumairAK:
correction, the workaround may not work because I think the auth-operator (or w/e creates the copy) simply ends up overwriting it anyway and removing the labels. ¯\_(ツ)_/¯ 

I've kept auto-sync turned off for now.

tumido:
My original thought was to apply this annotations to `moc-sso-client-secret`. Then it will be propagated to the copy as well. The copy code in the oauth operator is trying to ensure parity between the original secret and the copy.

Having this annotations on the original would mark it unprunable but I don't think this is a big deal.

HumairAK:
Okay so #280 seems to have fixed it, thanks @tumido  -- this way we have it tracked as well, we still have to stare at that yellow resource :angry: but at least sync status is green :+1:

PR #280, opened 2021-02-24
Update moc-sso secret to be ignored by argocd

HumairAK:
Apply workaround suggested in #278


tumido (commented):
/lgtm

let's try it :slightly_smiling_face:

PR #288, opened 2021-03-01
feat: Prepare subscription for MOC CNV 4.7 upgrade

tumido:
We need to prepare CNV manifests to be deployable on OpenShift 4.7

**Do not merge while: 1) autosync on cluster-resources is on, 2) CNV cluster is not yet on OCP 4.7**

Part of: https://github.com/open-infrastructure-labs/ops-issues/issues/21

/hold

Update: Resolves https://github.com/operate-first/apps/issues/323

### DIFF
--- a/cluster-scope/base/subscriptions/cluster-logging-operator/subscription.yaml
+++ b/cluster-scope/base/subscriptions/cluster-logging-operator/subscription.yaml
@@ -4,7 +4,7 @@ kind: Subscription
 metadata:
   name: cluster-logging
 spec:
-  channel: "4.5"
+  channel: DEFINED_IN_OVERLAY
   installPlanApproval: Automatic
   name: cluster-logging
   source: redhat-operators
--- a/cluster-scope/base/subscriptions/elastic-operator/subscription.yaml
+++ b/cluster-scope/base/subscriptions/elastic-operator/subscription.yaml
@@ -4,7 +4,7 @@ kind: Subscription
 metadata:
   name: elasticsearch-operator
 spec:
-  channel: "4.5"
+  channel: DEFINED_IN_OVERLAY
   installPlanApproval: Automatic
   name: elasticsearch-operator
   source: redhat-operators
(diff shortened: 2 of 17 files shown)


larsks:
Should we be opening doc bugs on all the "guessing the channel because there is no explicit documentation" comments?

tumido:
@larsks I think the benefit of it may be disputable - first of all, we don't have 4.7 cluster available to us yet. The "go-to" install strategy for all OLM operators is to subscribe via UI. And in the OperatorHub UI, all the channels would be listed properly, so there's no "missing pieces" or something for this type of install.

In our case, since we always kinda expect special treatment, we're hitting the limits. :smile: 

There's also some confusion about channels and operator versions mapping to OpenShift releases, since some of the operators are maintaining a channel per OCP release, some operators were initially doing that (and now they seem to be lacking behind) or they already released 5.0, while previously they've followed the OCP release cycle.. it's really disorganized.

For these operators the docs are not clear:
- OpenShift Virtualization (lacking docs for 4.7) - https://github.com/openshift/openshift-docs/issues/29926
- OpenShift Local Storage (lacking channel description in the docs) - https://github.com/openshift/openshift-docs/issues/29927
- OpenShift Container Storage (Their latest release seems to be 4.6 which is suspicious) - I have no idea where to report this. Maybe Here? https://issues.redhat.com/secure/RapidBoard.jspa?rapidView=8377&projectKey=OCSDOCS

anishasthana (approved):
/lgtm

tumido:
@larsks @HumairAK review please :slightly_smiling_face:

larsks (approved):

4n4nd (changes requested):
add the local storage operator patch in main kustomize (https://github.com/operate-first/apps/blob/bc7a99bea520544d6b555a58ff308d32a79c7a5d/cluster-scope/overlays/moc/zero/kustomization.yaml#L89)

PR #290, opened 2021-03-01
feat: Add ODH Z-stream updater script

tumido:
Part of: https://github.com/operate-first/support/issues/89
Resolves: https://github.com/operate-first/toolbox/issues/10

1. Checks the latest release of ODH on GitHub
2. Loops through `kfdef.yaml`s and bumps version in all `opendatahub` resources:
    - Updates `.spec.version`
    - Updates the tarball `uri` pointer in `.spec.repos` where the repo `name` matches `manifests`.

Requires [`yq`](https://github.com/mikefarah/yq) to be installed:  https://github.com/operate-first/toolbox/pull/14

### DIFF
--- a/hack/set-odh-release.sh
+++ b/hack/set-odh-release.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+set -o errexit
+trap 'echo "Aborting due to errexit on line $LINENO. Exit code: $?" >&2' ERR
+set -o errtrace
+set -o pipefail
+
+k=$(find . -name kfdef.yaml)
+
+latest_version=$(basename $(curl -Ls -o /dev/null -w %{url_effective} https://github.com/opendatahub-io/odh-manifests/releases/latest))
+
+for kfdef in $k; do
+    if ! yq e -e ".metadata.name == \"opendatahub\"" $kfdef 2>&1 >/dev/null; then
+        echo skipping $(dirname $kfdef)
+        continue
+    fi
+    echo updating $(dirname $kfdef)
+
+    yq e ".spec.version = \"$latest_version\"" -i $kfdef
+    yq e "(.spec.repos.[] | select(.name == \"manifests\").uri = \"https://github.com/opendatahub-io/odh-manifests/tarball/$latest_version\")| \"---\", ." -i $kfdef
+done
+
+#end.


hack/set-odh-release.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+
+set -o errexit
```
anishasthana, line 3: Just curious, what do you think about having this in the toolbox repo?

tumido: I don't mind having it there, but to me it seems this script is very limited in it's context. I feel like it doesn't make much sense to have it as a generic script if it's usable in a single repo context. I feel the same about the `onboarding.sh` script.

https://github.com/operate-first/toolbox/issues/10#issuecomment-787993230

anishasthana: I don't think I disagree. Only counter-point is that this is applicable to any deployments of the ODH, so it's arguably somewhat general. 

Maybe for the time being we leave it here, and if there's a situation where we would add this to any other operate-first repo we can move it over.

anishasthana (approved):
/lgtm

hack/set-odh-release.sh
```diff
@@ -0,0 +1,23 @@
...
+trap 'echo "Aborting due to errexit on line $LINENO. Exit code: $?" >&2' ERR
+set -o errtrace
+set -o pipefail
+
+k=$(find . -name kfdef.yaml)
+
+latest_version=$(basename $(curl -Ls -o /dev/null -w %{url_effective} https://github.com/opendatahub-io/odh-manifests/releases/latest))
+
+for kfdef in $k; do
+    if ! yq e -e ".metadata.name == \"opendatahub\"" $kfdef 2>&1 >/dev/null; then
```
HumairAK, line 13: A small nitpick here, since this script requires `yq` I'd suggest moving this conditional statement outside of the if statement so it can be caught by errexit.

tumido: Sorry @HumairAK I'm a bit confused here and I don't probably fully understand what you want me to do. This if is checking on the exit status of the `yq` command - if `yq` exits with `0` we **don't** skip the file (checking for the opposite - `!=` instead of `==` results in ugly error prints)

I'm not that strong in bash (and I actually don't like it at all), so if you know about any better way to check on exit status with the `errexit` flag on, please help me out.

HumairAK: hrmm I guess what I'm trying to say is if the user doesn't have `yq` installed, this loop will still run to completion instead of breaking early, and just complain every iteration that `yq` command not found. But I suppose if you want to use the exit status as part of the conditional then let's just go with it. A work around would be to maybe do maybe add something like `yq --version > /dev/null` in the beginning, so it can fail early if yq is not installed.

This isn't that big of a deal, let's just go with it.

PR #294, opened 2021-03-02
reconfgured CI/CD to use prow and aicoe-ci

goern:
This uses Prow to run the unit tests.

Signed-off-by: Christoph Görn <goern@redhat.com>


tumido (approved):
/approve

PR #299, opened 2021-03-02
[CLO] Increase the ES node count to 5

4n4nd:
This is in order to prevent a networking bottleneck and hopefully solve network timeout issues with Elasticsearch.

Related https://github.com/operate-first/SRE/issues/98


tumido (commented):
/lgtm

PR #300, opened 2021-03-02
CLO - reduce cpu count per es node

4n4nd:


HumairAK:
This is to bring the cpu in line with what is actually be used as we have monitored in live deployment.

HumairAK (approved):
/lgtm
/approve

PR #302, opened 2021-03-04
integrate ray into MOC ODH

erikerlandson:
This PR will address https://github.com/operate-first/support/issues/102, by porting the following integration onto MOC ODH:
https://github.com/erikerlandson/ray-odh-demo

The following components will need to be added to the overlays:
- [x] ray-ml-notebook image
- [ ] ray-operator image
- [ ] ray-operator CRD
- [ ] ray singleuser profile
- [ ] ray cluster template
- [ ] ray-ml worker image

### DIFF
--- a/odh/base/jupyterhub/notebook-images/kustomization.yaml
+++ b/odh/base/jupyterhub/notebook-images/kustomization.yaml
@@ -11,4 +11,5 @@ resources:
   - ml-prague-workshop.yaml
   - ocp-ci-analysis.yaml
   - openshift-anomaly-detection.yaml
+  - ray-ml-notebook.yaml
   - time-series.yaml
--- a/odh/base/jupyterhub/notebook-images/ray-ml-notebook.yaml
+++ b/odh/base/jupyterhub/notebook-images/ray-ml-notebook.yaml
@@ -1,0 +1,16 @@
+kind: ImageStream
+apiVersion: image.openshift.io/v1
+metadata:
+  name: ray-ml-notebook
+  labels:
+    opendatahub.io/notebook-image: 'true'
+spec:
+  lookupPolicy:
+    local: true
+  tags:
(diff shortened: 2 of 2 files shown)


erikerlandson:
cc @durandom

tumido (approved):
seems like the test-kustomize-build test script is not referenced properly, don't mind that.

/lgtm
/approve

erikerlandson:
This should not have been merged, as you can see only one of the six tasks was completed.

HumairAK:
@erikerlandson may I suggest creating an issue instead with this same checklist, and linking more granular commits/PRs that link back to that issue?

tumido:
You can also use `/hold` command to prevent it from being merged. Sorry for being that happy on the trigger.

erikerlandson:
`[WIP]` is a common convention, I suppose I should have used it.  I've never seen `/hold` before.  Speaking of that - is there a "guide to contributors" page where this kind of stuff can live?  You can even reference the link to a guide in the PR template. "If this is your first PR, please go read guide-to-contibutors.url first..."

tumido:
We have this, though it would definitely need some polishing.. 
https://github.com/operate-first/apps/blob/master/contributing.md

We also use Prow for chat ops, so this can be used as a reference:
https://prow.k8s.io/command-help

But I admit we don't link that or explain that anywhere. Filing an issue for that rn.

erikerlandson:
If we go with multi-PR, I'm not opposed, but the components won't operate until all of them are active.  Possibly that is a benefit, we can look at them on the system before they are active.

tumido:
Let's do it step by step. It may also happen that different bits would require changes to different repos anyways.. so let's do small steps... Also, please understand that the CNV cluster can't be trusted now and is pretty unstable - we have scheduled a reinstall for Monday.

PR #311, opened 2021-03-05
Onboard user to data science and create ml-ws ns.

HumairAK:
As per https://github.com/operate-first/support/issues/107#issuecomment-790981338

Onboarding user to DS group + creating a new namespace.


4n4nd (commented):
/lgtm

PR #318, opened 2021-03-05
upgrade cnv subscription to 2.5.4

larsks:
this should be the latest version of this operator, preventing argocd
and the automatic upgrades from fighting eachother.


HumairAK (approved):
/lgtm

PR #321, opened 2021-03-05
wire up ray cluster CRD into kustomize overlays

erikerlandson:
The portion of https://github.com/operate-first/support/issues/102 that addresses the ray cluster CRD

### DIFF
--- a/cluster-scope/base/crds/cluster.ray.io/kustomization.yaml
+++ b/cluster-scope/base/crds/cluster.ray.io/kustomization.yaml
@@ -1,0 +1,5 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ray-cluster-crd.yaml
--- a/cluster-scope/base/crds/cluster.ray.io/ray-cluster-crd.yaml
+++ b/cluster-scope/base/crds/cluster.ray.io/ray-cluster-crd.yaml
@@ -1,0 +1,4293 @@
+kind: CustomResourceDefinition
+apiVersion: apiextensions.k8s.io/v1
+metadata:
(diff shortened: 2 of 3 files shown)


erikerlandson:
cc @tumido

tumido:
I'm sorry @erikerlandson this was probably a misscomunication on my end here https://github.com/operate-first/support/issues/102#issuecomment-791338447 The first point still applies.

The CRD is the only resource that has to be placed elsewhere than the JupyterHub namespace. We want to track the resources available to all users (which any custom resource definition is) to be available on a plain sight. If it doesn't come from an operator subscription, it has to be placed here:

https://github.com/operate-first/apps/tree/master/cluster-scope/base/crds

Can you please move it in there?

And then reference the resource in `kustomization.yaml` for the given cluster (`zero` now):
https://github.com/operate-first/apps/blob/master/cluster-scope/overlays/moc/zero/kustomization.yaml#L19

For further info see:
https://github.com/operate-first/blueprint/blob/main/docs/adr/0009-cluster-resources.md
https://github.com/operate-first/apps/blob/master/cluster-scope/README.md

tumido (changes requested):
See above

erikerlandson:
@tumido this didn't merge for some reason

tumido:
@erikerlandson yeah it did not, because our CI is running on the cluster that is currently down :man_facepalming: 

I can merge it manually once we have a cluster to actually deploy this on. :slightly_smiling_face:

HumairAK:
CRD was added to zero cluster: 
```
$ kubectl api-resources --namespaced=true  | grep RayCluster
rayclusters                                                                  cluster.ray.io                               true         RayCluster

```

PR #322, opened 2021-03-06
Add network interfaces on crc provisioning network

larsks:
Creates eno1.307 on all nodes and configures a different static
address for each node.

ACM requires access to the provisioning network of the target cluster
in order to communicate with the Ironic API. In 4.8 (and in future 4.7
releases) this communication will happen on the API network and these
cross-cluster links will no longer be necessary.


tumido (commented):
/lgtm

HumairAK:
Changes applied.

PR #324, opened 2021-03-08
update sso configuration for zero cluster

larsks:
configure new client id and secret for zero.massopen.cloud


tumido (approved):
/approve

PR #325, opened 2021-03-08
feature/zero moc tls

larsks:
issue new letsencrypt cert for zero.massopen.cloud


larsks:
I would prefer to automate this with cert-manager, but we can start with this.

tumido (approved):
/approve

PR #338, opened 2021-03-09
Update path for kafka/superset for override in zero cluster.

HumairAK:
Due to the restructuring, the location of these overrides change, we need to update the kfdefs to reflect that.


harshad16 (commented):
/lgtm

PR #351, opened 2021-03-11
encrypt groups/argocd-admins.enc.yaml to ipolonsk

larsks:


tumido:
@larsks we can't be merging PRs without any reviews. And author shouldn't force merge their own PRs please.

The PR is adding a GPG key of unknown origin and re-encrypts a resource with this key. (I presume it is owned by @ipolonsk but there's no description to it, I couldn't find it on GNUPG and Ubuntu keyservers). Without properly distributing the public key to this new GPG key, we're unable to update the resource now (because sops needs all public keys are required).

PR #352, opened 2021-03-11
Added ipolonsk user to the admin

ipolonsk:


larsks (approved):

tumido (commented):
Not sure this group will give you what you want. Can you explain what kind of permissions you're looking for?

The `argocd-admins` is used to access all the ArgoCD applications only:

https://github.com/operate-first/continuous-deployment/blob/5a9c679a1b88c940a8b6921786b4603107f5807e/manifests/overlays/moc-infra/configs/argo_rbac_cm/policy.csv#L13

Note: I have nothing against you being included in the group, I'm just unsure if this is the right solution to the issue you're trying to solve.

ipolonsk:
Hey @tumido,
I would love to be able to see the application and login into the console, 
I think we maybe need another group to be between admin and user for use as mine.

tumido:
ok, what about we start with including you into the `operate-first` group first? And if you need more permissions, we just change it? :slightly_smiling_face:

ipolonsk:
@tumido That will be great! thank you.
Do I need to delete this PR? or close it?

tumido:
I think we can wait for the gpg keys PR to merge and then just rebase it. Or I can do it for you. Whatever you prefer. :slightly_smiling_face:

HumairAK:
@ipolonsk can you rebase over #353  and add the user to the [operate-first group](https://github.com/operate-first/apps/blob/master/cluster-scope/overlays/moc/common/groups/operate-first.enc.yaml#L7) as per discussion above?

tumido:
@HumairAK we're working on a replacement PR instead

HumairAK:
@tumido does that mean this pr can be closed?

PR #372, opened 2021-03-12
fix: Datacatalog bucket

tumido:
Resolves: https://github.com/operate-first/support/issues/117

/hold requires manual sync after merge (because we're re-defining the storage class - original bucket has to be deleted first)


HumairAK (approved):
/lgtm

PR #378, opened 2021-03-14
Update monitoring

### LINKED ISSUE #215: [Metrics] Opf-Kafka metrics aren't available
We have Pod Monitors from [upstream ODH](https://github.com/opendatahub-io/odh-manifests/blob/master/prometheus/operator/base/kafka-podmonitors.yaml), but we still need to override them to add a namespace selector ([docs](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#podmonitorspec)) since the opf-kafka deployment is in a different namespace than the opf prometheus.

anishasthana:
Fixes #215 

Two Questions:
1) I haven't actually created overrides before, please let me know if I should have done things differently here
2) The prometheus-k8s service account used by the operator for monitoring doesn't have access to the `openshift-operators` namespace, as far as I can tell (I checked by logging in as the SA and trying to get services). I'm a _little_ confused as to how I'd give the SA access. (I was looking at https://github.com/operate-first/apps/tree/master/cluster-scope but I'm not clear).

### DIFF
--- a/cluster-scope/base/namespaces/openshift-operators-redhat/kustomization.yaml
+++ b/cluster-scope/base/namespaces/openshift-operators-redhat/kustomization.yaml
@@ -9,3 +9,4 @@ resources:
 
 components:
   - ../../../components/project-admin-rolebindings/operate-first
+  - ../../../components/monitoring-rbac
--- a/cluster-scope/base/namespaces/openshift-operators/kustomization.yaml
+++ b/cluster-scope/base/namespaces/openshift-operators/kustomization.yaml
@@ -2,7 +2,7 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
-namespace: openshift-operators-redhat
+namespace: openshift-operators
 
 components:
   - ../../../components/monitoring-rbac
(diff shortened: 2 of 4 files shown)


tumido (commented):
Ad 1. This looks good - the override is fine
Ad 2. For the monitoring RBAC this line does the trick: 
https://github.com/operate-first/apps/blob/8e879cec15ae06afa1262ad15586ad7b3168e2ec/cluster-scope/base/namespaces/opf-kafka/kustomization.yaml#L12

You would need to define a `kustomization.yaml` for the target namespace adding the monitoring component. That should be all you need to set.

/lgtm

tumido (commented):
good catch :+1: 

/lgtm

4n4nd (commented):
/lgtm looks good :+1: 
Thanks @anishasthana!

tumido (approved):
/approve

4n4nd:
This PR impements:
1. Updates the odhservice monitor to monitor the odh-operator deployed in `openshift-operators` namespace.
2. Adds pod monitor for kafka deployed in `opf-kafka` namespace
3. Adds monitoring rbac in `openshift-operators` and  `openshift-operators-redhat` namespaces

PR #390, opened 2021-03-16
attempt to upgrade cnv to 2.6.0

larsks:
There's an update available for 2.6.0 waiting for manual approval. We
talked about seeing what the behavior is if we update the subscription
with a startingCSV attribute matching the new version. This commit is
that test.


tumido (commented):
/lgtm

HumairAK (approved):
/lgtm

PR #398, opened 2021-03-17
Introduce manual ODH deployment and remove subscription.

### LINKED ISSUE #206: Remove ODH operator via OLM and use deployment instead
^ That is a provocative title

Resolve https://github.com/opendatahub-io/opendatahub-operator/issues/85 https://github.com/operate-first/apps/issues/167

and deploy the operator manually similar to Observatorium

HumairAK:
Merge in combination with: https://github.com/operate-first/argocd-apps/pull/105

Resolves #206 

We should manually uninstall the operator from olm, we will also likely need to remove the kfdef aggregate roles ourselves. In my testing these were not removed after uninstalling the operator. 

Sourced from https://github.com/opendatahub-io/opendatahub-operator/tree/master/kustomize/overlays/opendatahub

I think a follow up action item would be update the z-release script to make sure these resources are in sync. WDYT?


tumido (changes requested):
I've noticed `cluster-scope/base/crds/kfdef.apps.kubeflow.org/kustomization.yaml` is empty. It should include `cluster-scope/base/crds/kfdef.apps.kubeflow.org/crd.yaml` resource

HumairAK:
@tumido good catch, resolved.

tumido (commented):
/lgtm

anishasthana (approved):
/lgtm

4n4nd (approved):

tumido (approved):
/approve

PR #399, opened 2021-03-17
Add @maulikjs to the operate-first group

4n4nd:
Add .sops.yaml for easier sops encryption

/cc @maulikjs


cluster-scope/overlays/moc/common/.sops.yaml
```diff
@@ -0,0 +1,3 @@
+creation_rules:
+  - encrypted_regex: '^users$'
+    pgp: '0508677DD04952D06A943D5B4DC4116D360E3276,B27ED6FAF92F28FA805451F33F5CA0E2A1824018,5B2E9490ED4F7BB379EC93C17BF065CD97112F06,3625572E2E3C694CB19911FFB727FBE237CEADAC'
```
tumido, line 3: I don't think we want this. It means that every new group added to any MOC cluster has to be encrypted by @maulikjs's key. Why?

tumido (changes requested):

cluster-scope/overlays/moc/common/.sops.yaml
4n4nd: This is not maulikjs's key, this dir was missing a .sops.yaml so I just added it. These are the original keys.

tumido: Oh, gotcha. Did not expect it in this PR. Nevermind :slightly_smiling_face:

tumido (commented):
/lgmt

tumido (approved):
/approve

PR #401, opened 2021-03-18
Add untracked kfdefs to git.

HumairAK:
Related: https://github.com/operate-first/support/issues/133#issuecomment-801956856 , https://github.com/operate-first/apps/pull/398#issuecomment-801975054

### DIFF
--- a/docs/odh/adding_kfdefs.md
+++ b/docs/odh/adding_kfdefs.md
@@ -1,0 +1,17 @@
+## Instructions on how to add your KFDEF
+
+Due to security reasons, we do not allow users to deploy their own `Kfdefs` (see [here][1] for details). However, should users want to request a `Kfdef` to their namespace, they can follow the instructions below for submitting a PR.
+
+Steps:
+Let `ENV`, `CLUSTER`, and `NAMESPACE` be the environment, cluster, and namespace you would like to deploy this `Kfdef` to respectively.
+1. Fork & Clone https://github.com/operate-first/apps
+2. Add your `Kfdef` to `kfdefs/overlays/$ENV/$CLUSTER/$NAMESPACE/kfdef.yaml`, accompanied by a `Kustomization.yaml` that includes the `kfdef.yaml`.
+3. Add this `Kfdef` to the `kfdefs/overlays/$ENV/$CLUSTER/kustomization.yaml`
+4. Submit a PR
+
+Note:
+- Since `Kfdefs` can be used to deploy manifests from external repositories, we only allow repositories that come from trusted sources. In general these are repositories part of the Open Data Hub, Kubeflow, or Operate-First org. Exceptions may apply.
+- Please avoid adding cluster-wide operators to your `Kfdef` if they have already been deployed by another `Kfdef`.
+
+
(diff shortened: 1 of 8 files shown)


docs/odh/adding_kfdefs.md
```diff
@@ -0,0 +1,16 @@
+## Instructions on how to add your KFDEF
+
+Due to security reasons, we do not allow users to deploy their own `Kfdefs` (see [here][1] for details). However, should users want to request a `Kfdef` to their namespace, they can follow the instructions below for submitting a PR.
+
+Steps:
+Let `ENV`, `CLUSTER`, and `NAMESPACE` be the environment, cluster, and namespace you would like to deploy this `Kfdef` to respectively.
+1. Add your `Kfdef` to `operate-first/apps/kfdefs/overlays/$ENV/$CLUSTER/$NAMESPACE/kfdef.yaml`, accompanied by a `Kustomization.yaml` that includes the `kfdef.yaml`.
```
tumido, line 7: I'd drop the:
```suggestion
1. Add your `Kfdef` to `kfdefs/overlays/$ENV/$CLUSTER/$NAMESPACE/kfdef.yaml`, accompanied by a `Kustomization.yaml` that includes the `kfdef.yaml`.
```

since we're already in this repo.

docs/odh/adding_kfdefs.md
```diff
@@ -0,0 +1,16 @@
...
+2. Add this `Kfdef` to the `operate-first/apps/kfdefs/overlays/$ENV/$CLUSTER/kustomization.yaml`
```
tumido, line 8: Same as above
```suggestion
2. Add this `Kfdef` to the `kfdefs/overlays/$ENV/$CLUSTER/kustomization.yaml`
```

tumido (changes requested):
Do we want the [`set-odh-release`](https://github.com/operate-first/apps/blob/master/scripts/set-odh-release.sh) script to touch these resources?

We should probably change the script to change only resources in the `odh` folder.

docs/odh/adding_kfdefs.md
HumairAK: If we put this on the site then it may not be clear what repo we're in. What about adding a "clone this repo" set above instead?

HumairAK:
> Do we want the set-odh-release script to touch these resources? 

Good question, I have no idea. I'm not exactly sure how older manifests would interact with changes made to newer releases of the operator. AFAIK the operator it self will be updating every z-release as well right? So I'd imagine we would want to keep all kfdefs in sync. Otherwise we can't guarantee the user odh deployed services to work as intended.

But then we may be causing redeployments for users' services upon every z-release.

docs/odh/adding_kfdefs.md
HumairAK: Updated, let me know if this works.

harshad16 (commented):
/lgtm 
thanks for including our stuff.

4n4nd (commented):
/lgtm

tumido (approved):
/approve

PR #403, opened 2021-03-18
Move ODH deployment to its own ns.

HumairAK:
Given that odh is manually deployed w/o OLM, it makes more sense to be consistent with our own philosophy and deploy it to its own namespace.

### DIFF
--- a/cluster-scope/base/namespaces/opendatahub-operator/kustomization.yaml
+++ b/cluster-scope/base/namespaces/opendatahub-operator/kustomization.yaml
@@ -1,0 +1,11 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: opendatahub-operator
+
+resources:
+  - namespace.yaml
+
+components:
+  - ../../../components/monitoring-rbac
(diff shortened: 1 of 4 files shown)


tumido (changes requested):
You need to create an `OperatorGroup` resource for this namespace.

cluster-scope/base/namespaces/opendatahub-operator/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: opendatahub-operator
+
+resources:
+  - namespace.yaml
+
+components:
+  - ../../../components/project-admin-rolebindings/operate-first
```
4n4nd, line 11: let's remove the operate-first group from this? since this would let anyone in the group become cluster-admin

tumido (commented):
Don't mind me, you don't need `OperatorGroup` since it's not OLM anymore.

HumairAK:
Re-review please

4n4nd (commented):
/lgtm

tumido (approved):
/lgmt
/approve

PR #405, opened 2021-03-18
add metallb

larsks:
the cnv cluster was running metallb to provide public ip for services
that don't work with the ingress router (e.g. for running ssh in a pod
or virtual machine).

This commit adds the basic manifests.

### DIFF
--- a/cluster-scope/overlays/moc/zero/kustomization.yaml
+++ b/cluster-scope/overlays/moc/zero/kustomization.yaml
@@ -5,6 +5,7 @@ kind: Kustomization
 resources:
   - ../common
   - ../../../base/clusterroles/cluster-logs-reader
+  - ../../../base/clusterroles/metallb
   - ../../../base/clusterroles/metrics
   - ../../../base/clusterroles/observatorium-operator
   - ../../../base/clusterroles/opendatahub-operator
@@ -16,6 +17,7 @@ resources:
   - ../../../base/clusterrolebindings/cluster-admins-rb
   - ../../../base/clusterrolebindings/cluster-logs-readers
   - ../../../base/clusterrolebindings/cluster-metrics-federation
+  - ../../../base/clusterrolebindings/metallb
   - ../../../base/clusterrolebindings/metrics
   - ../../../base/clusterrolebindings/observatorium-operator
   - ../../../base/clusterrolebindings/opendatahub-operator
--- a/cluster-scope/overlays/moc/zero/kustomization.yaml
+++ b/cluster-scope/overlays/moc/zero/kustomization.yaml
@@ -5,7 +5,8 @@ kind: Kustomization
 resources:
   - ../common
   - ../../../base/clusterroles/cluster-logs-reader
-  - ../../../base/clusterroles/metallb
+  - ../../../base/clusterroles/metallb-system-controller
+  - ../../../base/clusterroles/metallb-system-speaker
   - ../../../base/clusterroles/metrics
   - ../../../base/clusterroles/observatorium-operator
   - ../../../base/clusterroles/opendatahub-operator
@@ -17,7 +18,9 @@ resources:
   - ../../../base/clusterrolebindings/cluster-admins-rb
   - ../../../base/clusterrolebindings/cluster-logs-readers
   - ../../../base/clusterrolebindings/cluster-metrics-federation
-  - ../../../base/clusterrolebindings/metallb
+  - ../../../base/clusterrolebindings/metallb-allow-privileged
+  - ../../../base/clusterrolebindings/metallb-system-controller
+  - ../../../base/clusterrolebindings/metallb-system-speaker
   - ../../../base/clusterrolebindings/metrics
   - ../../../base/clusterrolebindings/observatorium-operator
   - ../../../base/clusterrolebindings/opendatahub-operator
@@ -42,6 +45,7 @@ resources:
   - ../../../base/namespaces/local-storage
   - ../../../base/namespaces/m4d-blueprints
   - ../../../base/namespaces/m4d-system
+  - ../../../base/namespaces/metallb-system
   - ../../../base/namespaces/mesh-for-data
   - ../../../base/namespaces/observatorium-operator
   - ../../../base/namespaces/open-aiops
@@ -87,6 +91,8 @@ resources:
   - ../../../base/operatorgroups/openshift-metering
   - ../../../base/operatorgroups/openshift-operators-redhat
   - ../../../base/operatorgroups/openshift-storage
+  - ../../../base/podsecuritypolicy/metallb-system-controller
+  - ../../../base/podsecuritypolicy/metallb-system-speaker
   - ../../../base/storageclasses/ocs-storagecluster-cephfs
   - ../../../base/subscriptions/cluster-logging-operator
   - ../../../base/subscriptions/elastic-operator
--- a/metallb/base/kustomization.yaml
+++ b/metallb/base/kustomization.yaml
@@ -1,0 +1,11 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: metallb-system
+
+resources:
+  - daemonset.yaml
+  - deployment.yaml
+  - rolebinding.yaml
+  - role.yaml
+  - serviceaccount.yaml
--- a/metallb/overlays/moc/zero/kustomization.yaml
+++ b/metallb/overlays/moc/zero/kustomization.yaml
@@ -1,6 +1,8 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
+namespace: metallb-system
+
 resources:
   - ../../../base
 
--- a/metallb/overlays/moc/zero/kustomization.yaml
+++ b/metallb/overlays/moc/zero/kustomization.yaml
@@ -1,0 +1,16 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - ../../../base
+
+configMapGenerator:
+  - name: config
+    files:
+      - config=config.yaml
+
+generators:
+  - secret-generator.yaml
+
+generatorOptions:
+  disableNameSuffixHash: true
--- a/cluster-scope/base/clusterrolebindings/metallb-allow-privileged/clusterrolebinding.yaml
+++ b/cluster-scope/base/clusterrolebindings/metallb-allow-privileged/clusterrolebinding.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: metallb-allow-privileged
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:openshift:scc:privileged
+subjects:
+  - kind: ServiceAccount
+    name: speaker
+    namespace: metallb-system
--- a/cluster-scope/base/clusterrolebindings/metallb-allow-privileged/kustomization.yaml
+++ b/cluster-scope/base/clusterrolebindings/metallb-allow-privileged/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - clusterrolebinding.yaml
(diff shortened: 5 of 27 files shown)


larsks:
I'm modeling this after the existing configuration for CNV (e.g., the top-level `kubevirt-hyperconverged` directory in the `apps` repository, with a corresponding entry in the `argocd-apps` repository). 

- First question: Is that the right structure?
- Second question: Should cluster-specific configuration live in `metallb/overlays/moc/zero`, or somewhere else?

tumido (changes requested):
@larsks please move `ClusterRole` and `ClusterRoleBinding` into `cluster-scope` app. These resources have to be unique by name in the whole cluster and are not namespaced, therefore they belong to a cluster-resources.

tumido:
@larsks :

> First question: Is that the right structure?

Not quite, any cluster scoped resource like `ClusterRole`, `Namespace` etc has to go to the `cluster-scoped` app so we avoid duplicity in deployments, conflicting specs for the same name and so on. Cluster scoped resources should be never referenced as a remote manifest, since that can be abused to elevate permissions on the cluster.

> Second question: Should cluster-specific configuration live in metallb/overlays/moc/zero, or somewhere else?

cluster specific resources should live in the overlay, that is correct

metallb/base/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: metallb
+
+resources:
+  - https://raw.githubusercontent.com/metallb/metallb/v0.9.5/manifests/namespace.yaml
```
tumido, line 7: Namespace has to go to `cluster-scoped` as well, not via a remote reference, but as a direct raw manifest.

cluster-scope/base/clusterrolebindings/metallb/clusterrolebinding.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: metallb-allow-privileged-scc
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: allow-privileged
```
tumido, line 8: You can do this without a separate cluster role actually, you can assign `system:openshift:scc:privileged` cluster role here instead.

See, for example:
https://github.com/operate-first/apps/blob/c3e5c619dccc19c44d4de2217f5735d26537f4fc/cluster-scope/base/clusterrolebindings/scc-privileged-opf-ci-pipelines-aicoe-ci/clusterrolebinding.yaml#L9

metallb/base/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - https://raw.githubusercontent.com/metallb/metallb/v0.9.5/manifests/metallb.yaml
+  - clusterrole.yaml
+  - rolebinding.yaml
```
tumido, line 10: Not present here anymore.

metallb/base/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: metallb
+
+resources:
+  - https://raw.githubusercontent.com/metallb/metallb/v0.9.5/manifests/namespace.yaml
+  - https://raw.githubusercontent.com/metallb/metallb/v0.9.5/manifests/metallb.yaml
```
tumido, line 8: This manifest is deploying `ClusterRole`s `ClusterRoleBinding`s: These resources can't be part of the app and have to go to the `cluster-scope`.

Not sure what privileges are required to deploy `PodSecurityPolicy` though.. This probably also require cluster admin, right?

tumido (changes requested):

metallb/base/kustomization.yaml
larsks: Yes, I realize that. Still making changes.

metallb/base/kustomization.yaml
larsks: So there's really no way to consume an upstream resource directly with our current repository structure, I guess? We have to pull the manifests locally and split things up by hand.

tumido: If they mix cluster scoped and namespaced resources, nope, there's no way to do that - because you don't want any remote references for the cluster scoped ones - remote can change and we don't want to automatically pull in different manifests and apply them with cluster admin permissions. It's neither deterministic nor secure. 

If you can get the cluster scoped/namespaced manifests as separate remote manifests, you can use the namespaced as a remote reference, there's no big risk in that, since you can only compromise your own namespace.

The cluster scoped resources have to be parsed out and adopted to our directory structure in `cluster-scope` any way.

larsks: Is the fact that this splits the serviceaccount (which is namespaced) and the clusterrolebinding that references it going to cause any problems?

tumido: It doesn't. Because of additive nature of OpenShift RBAC. `ClusterRoleBinding` is allowing something to somebody. Therefore the user/sa has to identify first in order for crb to be used - and if there's no sa to identify it just dosn't get used... That's why you can really put anything in the subject spec and it doesn't cause troubles until the user identifies. :smile: It's weird sometimes, but handy in other times. :slightly_smiling_face:

metallb/base/kustomization.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: metallb
```
tumido, line 4: Here you're using `metallb` namespace, while here it originally wants to deploy to `metallb-system`.
https://github.com/operate-first/apps/pull/405/files#diff-cf2e704bae45f65353f3bcc6dceaa69b150875d472fd9a5322253c00006c550bR8
and you're defining `metallb-system` namespace here:
https://github.com/operate-first/apps/pull/405/files#diff-0a4f86bc5dd47a741254b8a22c8c8e58196d4858a93a75c95c470d4f1958eea7R4

cluster-scope/overlays/moc/zero/kustomization.yaml
```diff
@@ -5,6 +5,7 @@ kind: Kustomization
 resources:
```
tumido, line 5: Missing the `metallb-system` namespace in this file

cluster-scope/base/namespaces/metallb/namespace.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: metallb-system
```
tumido, line 4: If the namespace should be named `metallb-system` than the folder shall be named the same. Or the namespace should be `metallb`, then the resource name has to change.

tumido (changes requested):
There's some confusion on namespaces you use here. You try to deploy to `metallb` (set in kustomization), resources use `metallb-system` internally, cluster role bindings use `metallb-system` and the folder for `metallb-system` namespace resource is named `metallb`. Can you please clear that confusion?

larsks:
I spotted the namespace problem; I was just submitting updates in both repos. Should be all set.

cluster-scope/overlays/moc/zero/kustomization.yaml
tumido: @larsks I'm still missing the namespace imported in here from the base. Am I looking right? Sorry getting lost in the PR a bit :smile:  

Also the podsecuritypolicies are missing from here.

tumido (changes requested):

larsks:
@tumido good catch, the namespace import was still missing. Fixed.

cluster-scope/overlays/moc/zero/kustomization.yaml
tumido: @larsks ^ the podsecuritypolicies are missing as well

tumido (changes requested):

cluster-scope/overlays/moc/zero/kustomization.yaml
larsks: Aaaand I've added those, also.

cluster-scope/base/clusterrolebindings/metallb/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - metallb-allow-privileged-scc.yaml
```
HumairAK, line 5: @larsks all of these should be in their separate folders like so: 

cluster-scope/base/clusterrolebindings/metallb-allow-privileged
cluster-scope/base/clusterrolebindings/metallb-system-controller
cluster-scope/base/clusterrolebindings/metallb-system-speaker

With each folder having the actual manifest named `clusterrolebinding.yaml` and one kustomization.yaml per manifest. 

This should apply to the `clusterroles` added with this PR as well. The value of `<resource>` in `cluster-scope/base/clusterrolebindings/<resource>` should match the `metadata.name` as much as it can. 

This is how the rest of the resources are also structured. Hope that makes sense.

cluster-scope/base/podsecuritypolicy/metallb/metallb-system-controller.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  labels:
+    app: metallb
+  name: controller
+  namespace: metallb-system
```
HumairAK, line 7: Should this resource have `metadata.namespace` ? I'm reading that this is a cluster-wide resource.

larsks: Actually, that's a good question. Are PodSecurityPolicies cluster-wide?

larsks: All the examples I can find suggest that yes, they are.

cluster-scope/base/podsecuritypolicy/metallb/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
tumido, line 2: No resources in this file.

Each folder in `cluster-scope/base/RESOURCE_PLURAL/` should be named the same as the resource `.metadata.name` and implement exactly one resource. See:
https://github.com/operate-first/apps/blob/master/cluster-scope/README.md

cluster-scope/base/clusterroles/metallb/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - metallb-system-controller.yaml
+  - metallb-system-speaker.yaml
```
tumido, line 6: Only a single resource per folder please. See:
https://github.com/operate-first/apps/blob/master/cluster-scope/README.md

metallb/base/kustomization.yaml
```diff
@@ -0,0 +1,19 @@
...
+namespace: metallb-system
+
+resources:
+  - daemonset.yaml
+  - deployment.yaml
+  - rolebinding.yaml
+  - role.yaml
+  - serviceaccount.yaml
+
+patches:
```
tumido, line 13: Why are we patching in a base? You should reflect that directly in the base manifests instead

metallb/overlays/moc/zero/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+namespace: metallb-system
+
+resources:
+  - ../../../base
+
+configMapGenerator:
+  - name: config
+    files:
+      - config=config.yaml
```
tumido, line 12: Are you generating a configmap from a file which is already a configmap? Why don't you use the `configmap.yaml` directly as a resource?

tumido (dismissed):

metallb/base/kustomization.yaml
larsks: The reason we're patching in a base is that when this all started the goal was to consume an upstream package without modification and then patch in the changes necessary for our environment, much like we handle RPMs by starting with the upstream source and then applying patches.

As we've discovered through this PR it's not really possible to consume an upstream package like that, so perhaps this patch is no longer necessary.

metallb/overlays/moc/zero/kustomization.yaml
larsks: I was, but not intentionally. Thanks for catching that.

tumido (commented):
/lgtm

PR #417, opened 2021-03-22
feat: Onboard RH Labs - CI/CD

tumido:
Part of: https://github.com/operate-first/support/issues/113

@acidonper, sorry for the delay, let me know if this PR is what you need. It creates an user group `lab-cicd` which you are set to be the sole user for now. This group is the project admin on all the namespaces.

Additionally it creates namespaces:
- `lab-cicd-1-jump-app-cicd`
- `lab-cicd-1-jump-app-dev`
- `lab-cicd-2-jump-app-cicd`
- `lab-cicd-2-jump-app-dev`
- `lab-cicd-3-jump-app-cicd`
- `lab-cicd-3-jump-app-dev`
- `lab-cicd-4-jump-app-cicd`
- `lab-cicd-4-jump-app-dev`

Each of this namespaces is available to be viewed (`view` cluster role) by any authenticated or unauthenticated user

Will this work for you or are any changes needed?

### DIFF
--- a/cluster-scope/base/groups/lab-cicd/group.yaml
+++ b/cluster-scope/base/groups/lab-cicd/group.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: user.openshift.io/v1
+kind: Group
+metadata:
+  name: lab-cicd
+users: []
--- a/cluster-scope/base/groups/lab-cicd/kustomization.yaml
+++ b/cluster-scope/base/groups/lab-cicd/kustomization.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- ./group.yaml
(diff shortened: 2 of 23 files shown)


acidonper:
@tumido thank you so much!

Just a little change, we need a namespace (E.g lab-cicd-gitops) to deploy ArgoCD by ourselves in order to provide the demo's attendees a tool to deploy the cicd objects, the demo applicacion and a managing tool to implement continuous deployment workflows based on Helm Charts. Do you think that it make sense or it is better to use another approach?

tumido:
@acidonper hm.. and would you mind using our ArgoCD? :) We can manage this namespace from our ArgoCD which you can get access to - We can create a separate `AppProject` that you would manage and own. It's scope would be limited to the ^ namespaces + would allow anybody to create apps in there via UI. We can also arrange you write access to [operate-first/workshop-apps](https://github.com/operate-first/workshop-apps) where you could be storing the app manifest created during the workshops declaratively - and have users open a PR with the manifests instead of doing it in the UI. (App manifests would be fetched automatically from that `workshop-apps` repository). What do you think?

acidonper:
@tumido, It would be awesome to have both possibilities. With this approach, I can introduce the attendees in "full" gitops procedures... I really love it!

tumido:
In that case, lets merge this PR and I'll follow up with an ArgoCD project and app-of-apps for the workshops repo... Additionally I've sent you an invite for that repo. :slightly_smiling_face: 

/cc @HumairAK @4n4nd review please

4n4nd (commented):
/lgtm

HumairAK (approved):
/lgtm

PR #421, opened 2021-03-23
Upgrade zero cluster to 4.7.2

larsks:
Update the zero cluster to 4.7.2.

The `ClusterVersion` resource includes a `clusterID` attribute that I have omitted from this update, since if we were to reinstall the cluster the ID would be different.

### DIFF
--- a/cluster-scope/overlays/moc/zero/clusterversion.yaml
+++ b/cluster-scope/overlays/moc/zero/clusterversion.yaml
@@ -1,0 +1,10 @@
+apiVersion: config.openshift.io/v1
+kind: ClusterVersion
+metadata:
+  name: version
+spec:
+  channel: stable-4.7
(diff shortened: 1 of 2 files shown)


HumairAK:
@larsks  should we announce this ahead of time w/ a date/time set, before attempting to deploy this?

tumido (commented):
/lgtm
/hold until we're ready

larsks:
@HumairAK  I think we should, yes.

larsks:
How about we schedule the upgrade for Monday (3/29) morning?

tumido:
@larsks sounds good to me :+1: Yep, let's do it

HumairAK (approved):
/lgtm

tumido (approved):
/approve

PR #422, opened 2021-03-23
Upgrade infra cluster to 4.6.21

larsks:
How do we want to handle updates on the infra cluster w/r/t scheduling?

### DIFF
--- a/cluster-scope/overlays/moc/infra/clusterversion.yaml
+++ b/cluster-scope/overlays/moc/infra/clusterversion.yaml
@@ -1,0 +1,10 @@
+apiVersion: config.openshift.io/v1
+kind: ClusterVersion
+metadata:
+  name: version
+spec:
+  channel: stable-4.6
+  upstream: https://api.openshift.com/api/upgrades_info/v1/graph
+  desiredUpdate:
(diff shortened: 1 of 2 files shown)


tumido (commented):
/lgtm
/hold

tumido:
@larsks I think we can add this to the Monday management window and do it along with the zero cluster upgrade. WDYT?

tumido (approved):
/approve

harshad16:
There is hold label added to this PR. 
@tumido just letting you know.

tumido:
Yup, we should plan on a new maintenance window for this, we missed the planned one. Thank @harshad16 

Wanna do it today/tomorrow @larsks @HumairAK ?

HumairAK:
@tumido let's do it tomorrow if @larsks can confirm. Let's send an email out as well: https://github.com/operate-first/SRE/blob/master/incident-management/incident-management-procedure.md#cluster-version-upgrade-notice-template

tumido:
i thought the same, thanks @HumairAK :+1:

larsks:
Tomorrow is fine. I'm not sure I would even bother with an email, since this isn't an end-user facing cluster. The only people who need to be aware of the state of the infra cluster are on `#ops`.

HumairAK:
@4n4nd as oncall do you want to send the e-mail out?

HumairAK (approved):

PR #431, opened 2021-03-24
Add asanmukh as Jupyterhub admin

4n4nd:
only one user can be added as admin for some reason, so I will manually add @HumairAK and @tumido as admins.


HumairAK:
I think this means you're the official admin and hence responsible for all future jupyterhub issues. 

/lgtm

HumairAK:
actually is this something better done in an overlay for zero environment? and don't you want to encrypt it?

anishasthana (approved):
/lgtm

tumido (approved):
/approve

Issue #438, opened 2021-03-25
Implement Resource Quotas for the different namespaces 


HumairAK:
I think it's time we start thinking about a strategy for implementing resource quotas for our namespaces, currently anyoen can request as much as they want and this is kinda of a disaster waiting to happen imo. 

I was thinking of having a standard tier like how PSI has done: bronze/silver/gold, with increasing resources permissions, we can also allow for custom resourcequotas for more niche situations. 

We will also need a [LimitRange](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/#create-a-limitrange-and-a-pod) that's reasonable, so that current pod creations that do not specify limits are not affected.

tumido:
Yup. Sounds good to me. I've found the culprit blocking our systems right now but in general yes, definitely - we need quota and rather asap.

/priority critical-urgent

tumido:
Related: This is the offender: https://github.com/operate-first/support/issues/159

tumido:
We need an ADR and some communication before we start enforcing this. Friday is a bad day for such changes also :slightly_smiling_face:

HumairAK:
@tumido I think we can add the resource quota tiers + limit ranges but not enforce it until next week. We can reference them in the ADRs as well.

HumairAK:
Adr issue: https://github.com/operate-first/blueprint/issues/64

tumido:
> @tumido I think we can add the resource quota tiers + limit ranges but not enforce it until next week. We can reference them in the ADRs as well.

> Edit: nvm looks like we already did merge it in.

precisely, that can is what I though about

HumairAK:
Implemented in #439

PR #439, opened 2021-03-25
Add resourcequotas and limitrange components.

HumairAK:
Related: #438 

The `resourcequotas` aren't included in any of the namespaces yet, this is one way we can go about this. The idea is to include these like we do every other component within the cluster-scope/base/namespaces.

We don't have to use tiers if there are good arguments against it, this is a suggestion. 

I just willy nilly picked amounts for each tier and `limitrange`. We can discuss what amounts are reasonable. 

WDYT?

### DIFF
--- a/cluster-scope/components/limitranges/default/limitrange.yaml
+++ b/cluster-scope/components/limitranges/default/limitrange.yaml
@@ -1,0 +1,13 @@
+kind: LimitRange
+apiVersion: v1
+metadata:
+  name: default
+spec:
+  limits:
+    - type: Container
+      default:
+        cpu: 500m
+        memory: 1000Mi
+      defaultRequest:
+        cpu: 300m
+        memory: 400Mi
--- a/cluster-scope/components/resourcequotas/large/resourcequotas.yaml
+++ b/cluster-scope/components/resourcequotas/large/resourcequotas.yaml
@@ -1,0 +1,11 @@
+kind: ResourceQuota
+apiVersion: v1
+metadata:
+  name: large
+spec:
+  hard:
+    requests.cpu: '4'
+    requests.memory: 16Gi
+    limits.cpu: '4'
+    limits.memory: 16Gi
+    requests.storage: 80Gi
--- a/cluster-scope/components/limitranges/default/kustomization.yaml
+++ b/cluster-scope/components/limitranges/default/kustomization.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+
+resources:
+- ./limitrange.yaml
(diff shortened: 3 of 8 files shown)


cluster-scope/components/limitranges/default/limitrange.yaml
```diff
@@ -0,0 +1,13 @@
+kind: LimitRange
+apiVersion: v1
+metadata:
+  name: limits
```
tumido, line 4: Can you please rename this to `default`?

cluster-scope/components/resourcequotas/bronze/resourcequotas.yaml
```diff
@@ -0,0 +1,12 @@
+kind: ResourceQuota
+apiVersion: v1
+metadata:
+  name: quota
```
tumido, line 4: Please rename to `bronze` (and the rest of quotas accordingly) it we're sticking with these ugly names...

User should be able to tell what tier they've got.

tumido (changes requested):
I very much dislike the names - what we're playing on here? Like.. `gold`, `silver`, `bronze` lol. :smile: 

What about renaming those to idk, `small`, `medium`, `large`, and so on? I like that better.

cluster-scope/components/resourcequotas/bronze/resourcequotas.yaml
HumairAK: > I very much dislike the names - what we're playing on here? Like.. gold, silver, bronze lol. smile 
I believe that's what PSI uses (that's where I got it from). 

I don't really care what we name them, I just want quotas -- I'll update them to small/med/large.

HumairAK:
Thoughts on the actual resource values for each tier?

cluster-scope/components/resourcequotas/large/resourcequotas.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: large
+spec:
+  hard:
+    requests.cpu: '8'
+    requests.memory: 40Gi
+    limits.cpu: '8'
+    limits.memory: 40Gi
+    persistentvolumeclaims: '8'
+    requests.storage: 80Gi
```
4n4nd, line 12: @HumairAK I think we can give out more storage and less memory

HumairAK: Any suggestions on specific values? And are you referring to just `large` or all?

larsks:
@HumairAK does it make sense to look at the current resource utilization of the cluster to help determine reasonable quota values?

HumairAK:
I've updated the values based on discussions in the `#ops` channel in slack.

4n4nd (commented):
/lgtm

tumido (approved):
/lgtm
/approve

PR #442, opened 2021-03-26
Add `system:authenticated` group in the allowed groups for JH access

4n4nd:
Resolves https://github.com/operate-first/support/issues/161


tumido (approved):
/approve

PR #449, opened 2021-03-26
subscribe curator cluster to container storage operator

larsks:
we are going to try getting the curator cluster connected to an
external ceph cluster for storage.

### DIFF
--- a/cluster-scope/overlays/moc/curator/kustomization.yaml
+++ b/cluster-scope/overlays/moc/curator/kustomization.yaml
@@ -6,7 +6,10 @@ resources:
   - ../../../base/clusterrolebindings/cluster-admins-rb
   - ../../../base/groups/cluster-admins
   - ../../../base/ingresscontrollers/default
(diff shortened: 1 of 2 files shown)


tumido (changes requested):
You need to include the `OperatorGroup` as well, otherwise the subscription wouldn't be picked up and operator won't install.

See the zero spec:
https://github.com/operate-first/apps/blob/2c27ba83acfe957037a1b967139d3d6238491d10/cluster-scope/overlays/moc/zero/kustomization.yaml#L102

larsks:
@tumido thanks, I added the operatorgroup, and as long as I was making changes I added the openshift-storage namespace as well, which we're going to need as soon as we get around to creating a storagecluster (that will probably happen manually first, as we work out exactly what we need to do for external ceph connectivity).

tumido (commented):
/lgtm

HumairAK (approved):

Issue #232, opened 2021-02-10
CLO LogCollector/LogForwarding for longterm storage


tumido:
https://github.com/operate-first/apps/pull/230 will limit the stored logs to 7 days. DeCorus wants to analyze logs from much longer periods of time. We need to figure out a scalable storage and forward the logs in there.

/sig devops

tumido:
Related to https://github.com/operate-first/support/issues/83

HumairAK:
>from much longer periods of time

Do we know how much longer?

tumido:
yes, the requirement is to have at least half a year worth of data

drbwa:
And we can even put 12 or 24 months of raw logs to good use. 

Compressed JSON files or maybe parquet (with a bit of investment to understand the log schema and making sure it saves significant space over compressed JSON files) in object storage might be an option.

drbwa:
I have looked into using the [ES Snapshot](https://www.elastic.co/guide/en/elasticsearch/reference/7.11/snapshot-restore.html) feature to back up ES indices to object storage as a potential solution to long-term storage of logs on MOC.

So far, I have come to the conclusion that this is not a suitable approach for our needs. It seems easy to set up and has some nice features (e.g., filter on snapshot and restore, existing integration with S3). However, in order to then use historic data, we would need to restore into a similar ES cluster (replicas, shards), which means that a) we need twice the resources to host such a cluster and b) can still only afford to restore and process seven days of data at a time (using a similarly configured cluster). Seems like too much of a headache.

It might be better to just store the logs in object storage to enable clients to download and transform as they see fit.

Two options:

* We could look into using the [fluentd S3 plugin](https://docs.fluentd.org/output/s3) to write logs to Ceph in addition to ES.
* We could write a custom script that runs hourly and reads the latest data from ES, compresses it and writes it to a bucket in Ceph.

How much space could we get in Ceph for logs?

Looking at a recent snapshot of `infra` and `app` indices, it looks the size of these indices with seven days of data is around 300 GB (in total). We could test how well this compresses using different approaches and would then want to store a multiple of that in object storage.

What do you think @HumairAK @tumido ?

/cc @ronensc

drbwa:
And of course, not to forget [Loki](https://github.com/grafana/loki) might be an option as well, as mentioned in <https://github.com/open-infrastructure-labs/telemetrywg/issues/4>.

4n4nd:
[...]

PR #460, opened 2021-03-30
Add topic for zero cluster app logs

4n4nd:
Related https://github.com/operate-first/apps/issues/232#issuecomment-809645674

### DIFF
--- a/odh/overlays/moc/zero/kafka/overrides/kafka/overlays/topics/zero-cluster-app-logs.yaml
+++ b/odh/overlays/moc/zero/kafka/overrides/kafka/overlays/topics/zero-cluster-app-logs.yaml
@@ -1,0 +1,9 @@
+apiVersion: kafka.strimzi.io/v1beta1
+kind: KafkaTopic
+metadata:
+  name: zero-prod.cluster.app-logs
+  labels:
+    strimzi.io/cluster: odh-message-bus
+spec:
+  partitions: 3
+  replicas: 2
--- a/odh/overlays/moc/zero/kafka/overrides/kafka/overlays/topics/kustomization.yaml
+++ b/odh/overlays/moc/zero/kafka/overrides/kafka/overlays/topics/kustomization.yaml
@@ -17,3 +17,4 @@ resources:
 - thoth-unresolved-package.yaml
 - thoth-unrevsolved-package.yaml
 - thoth-update-provides-source-distro.yaml
+- zero-cluster-app-logs.yaml


harshad16 (commented):
/lgtm

odh/overlays/moc/zero/kafka/overrides/kafka/overlays/topics/zero-cluster-app-logs.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: kafka.strimzi.io/v1beta1
+kind: KafkaTopic
+metadata:
+  name: zero-prod.cluster.app-logs
+  labels:
+    strimzi.io/cluster: odh-message-bus
+spec:
+  partitions: 1
```
anishasthana, line 8: This doesn't seem very resilient.

anishasthana (changes requested):

odh/overlays/moc/zero/kafka/overrides/kafka/overlays/topics/zero-cluster-app-logs.yaml
4n4nd: How many partitions would you recommend?

anishasthana: I'd also go with at least 2 replicas

4n4nd: updated :+1:

anishasthana (approved):
/lgtm

odh/overlays/moc/zero/kafka/overrides/kafka/overlays/topics/zero-cluster-app-logs.yaml
tumido: I don't think this effect resilience as much as it effects performance. You need to keep in mind how many message producers and how many consumers you want to have for that topic. Also mind that messages on different partition are not ordered. So in case you want to ensure message ordering and you expect to have a single consumer, it doesn't make sense for it to listen on 3 different partitions.

What are you expectations @4n4nd ? How many pods you expect to produce messages here and how many pods will be consuming the messages?

tumido (commented):
/hold let's discuss the replicas and the scale of this above a bit...

4n4nd:
@tumido since we have 3 brokers set, it makes sense to have 3 partitions and 2 replicas so that one copy of the message is stored with each broker.

tumido (commented):
Thank you @4n4nd for explanation

/lgtm

HumairAK (approved):

anishasthana (approved):
/unhold

Issue #471, opened 2021-04-01
Revisit `requests` for Elastic search


tumido:
CLO's Elastic never reaches the reservation (CPU request), set value is huge compared to the actual usage:

![image](https://user-images.githubusercontent.com/7453394/113296171-69888a00-92f9-11eb-91b3-b21e70158677.png)

There's no need to reserve 20 cores (4 cores per each of 5 pods) for Elastic only. A request of 4 in total is more than enough I think. We can set limit to `1200m` per pod.

Memory seems to be very consistent no matter what we do. There's no need to reserve 16Gi if it consumes ~10Gi only. Let's request 10Gi and limit to 16Gi
![image](https://user-images.githubusercontent.com/7453394/113296983-6510a100-92fa-11eb-9ef1-dbb80de5fe01.png)

PR #472, opened 2021-04-01
fix: Lower CPU request on openshift-logging

tumido:
Fixes: https://github.com/operate-first/apps/issues/471

### DIFF
--- a/cluster-logging/base/clusterlogging.yaml
+++ b/cluster-logging/base/clusterlogging.yaml
@@ -17,7 +17,10 @@ spec:
     elasticsearch:
       resources:
         requests:
-          cpu: 4
+          cpu: 800m
+          memory: 10Gi
+        limits:
+          cpu: 2
           memory: 16Gi
       # it's better to keep nodeCount to a odd number for leader election issues
       nodeCount: 5


cluster-logging/base/clusterlogging.yaml
```diff
@@ -17,7 +17,10 @@ spec:
     elasticsearch:
       resources:
         requests:
-          cpu: 4
+          cpu: 800m
+          memory: 10Gi
+        limits:
+          cpu: 1200m
```
4n4nd, line 23: Can we set the cpu limit to 2 instead?
or I think we can remove the limits completely because in case it wants to use more cpu and it's available we want it to use it.

tumido: Sure, why not.. I don't actually think Elastic is that effective in multicore, so anything above 1 should be rather scaled to multiple replicas.. But why not. :+1:

4n4nd (approved):
/lgtm

4n4nd:
applied and the pods were redeployed :+1:

tumido:
Seems better than before:
image

Issue #483, opened 2021-04-06
Deploy slack-first bot


tumido:
Deploy https://github.com/tumido/slack-first with "our" Operate-First configuration

tumido:
ArgoCD app missing

PR #484, opened 2021-04-06
feat: Add manifests for slack-first bot

tumido:
Resolves: https://github.com/operate-first/apps/issues/483

### DIFF
--- a/slack-first/base/deployment.yaml
+++ b/slack-first/base/deployment.yaml
@@ -1,0 +1,61 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: slack-first
+  labels:
+    app: slack-first
+spec:
+  selector:
+    matchLabels:
+      app: slack-first
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: slack-first
+    spec:
+      containers:
+        - name: slack-first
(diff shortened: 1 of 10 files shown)


4n4nd (commented):
/lgtm

slack-first/base/deployment.yaml
```diff
@@ -0,0 +1,61 @@
...
+      app: slack-first
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: slack-first
+    spec:
+      containers:
+        - name: slack-first
+          image: quay.io/tcoufal/slack-first:latest
```
anishasthana, line 20: Is it too early to move the image over to the operate-first umbrella?

anishasthana (commented):
Just had one question. Once that is addressed I'm happy to :ship:

slack-first/base/deployment.yaml
tumido: I think the devel on it is not happening in `operate-first` umbrella - I have my own CI set up on this and I'm releasing to quay from GitHub Actions during A GH release process (see here for example: https://github.com/tumido/slack-first/issues/29). I probably don't want to be releasing to operate-first from a my private (yet public) repo.

HumairAK (approved):

PR #488, opened 2021-04-07
Add instructions on adding resource quotas.

HumairAK:
Related: https://github.com/operate-first/blueprint/pull/65

Will follow up with a doc on the different tiers in the `operate-first/support` repo, since that's more user facing.

### DIFF
--- a/docs/cluster-scope/add_resource_quotas.md
+++ b/docs/cluster-scope/add_resource_quotas.md
@@ -1,0 +1,39 @@
+# Adding Resource Quotas
+
+To add a resource quota follow the steps outlined below:
+
+# Adding a resource quota tier to a namespace
+
+Select a tier from [this list][quotas].
+
+Now add this quota to the namespace. For example, if you want to add `small` tier quota to the namespace `argocd` run:
+
+```bash
+# From the root of this repo
+$ cd cluster-scope/base/namespaces/argocd
+$ kustomize edit add component ../../../components/resourcequotas/small
+```
+
+# Adding a custom resource quota to a namespace
+
+Create your [own resource][quotas_doc] in the appropriate namespace folder in `cluster-scope/base/namespaces/<your_namespace>` name this file `resourcequota.yaml`, ensure that the `.metadata.name` is set to `custom`. Then if adding a custom resource to the namespace `argocd` you would run:
+
+```bash
+# From the root of this repo
+$ cd cluster-scope/base/namespaces/argocd
+$ kustomize edit add resource ./resourcequota.yaml
+```
+
+# Add Limit Range to the namespace
+
+Ensure that a default `LimitRange` is configured if adding a resource quota, otherwise your pods may encounter issues during creation. You can include a `LimitRange` from [here][limit-range]. For example, if adding a `LimitRange` to the `argocd` namespace run the following:
+
+```bash
+# From the root of this repo
+$ cd cluster-scope/base/namespaces/argocd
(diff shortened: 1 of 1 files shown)


docs/cluster-scope/add_resource_quotas.md
```diff
@@ -0,0 +1,40 @@
+# Adding Resource Quotas
+
+To add a resource quota follow the steps outlined below:
+
+# Adding a resource quota tier to a namespace
+
+Select a tier from [this list][quotas].
+
+Now add this quota to the namespace. For example, if you want to add `small` tier quota to the namespace `argocd` then run:
```
anishasthana, line 9: remove the last then from this sentence

anishasthana (changes requested):

anishasthana (approved):
/lgtm

docs/cluster-scope/add_resource_quotas.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+# Adding a custom resource quota to a namespace
+
+Create your [own resource][quotas_doc]. Then add it to your namespace, for example if adding `${QUOTA_LOCATION}` to namespace `argocd` you would run:
+
+```bash
+# From the root of this repo
+$ cd cluster-scope/base/namespaces/argocd
+$ kustomize edit add resource ${QUOTA_LOCATION}
```
tumido, line 24: Let's be more specific here. Let's require users to name their quota file `resourcequota.yaml` and use `.metadata.name = custom`. That way we know, even when viewed from console, what the quota "type" it is. It will also keep the repo tidy.

WDYT?

HumairAK: Good point -- this actually poses a question I didn't think of, where should these custom quotas be stored? Should they be in `cluster-scope/components/resourcequotas/` ? If they are named `resourcequota.yaml` for every one, then they would have to be in some path that distinguishes them by namespace, or team name, etc. like: 

`cluster-scope/components/resourcequotas/custom/<namespace>/resourcequota.yaml`

tumido: I actually think we should keep these custom quotas together with the namespace. Unless we want a separate `kustomization.yaml` for each - because as outlined in this PR, we basically want them to be imported into the `namespaces/ABC/kustomization.yaml`. I don't think it should look like:

```yaml
---
kind: Kustomization

namespace: argocd

resources:
  - namespace.yaml
  - ../../resourcequotas/custom/argocd/resourceqouta.yaml

components:
  - ../../../components/project-admin-rolebindings/operate-first
  - ../../../components/monitoring-rbac

```

That doesn't look right to me.

1. Either keep them within the `namespaces/ABC` folder
2. Or do it like we do with `operatorgroups` and have a distinct folder and a `kustomization.yaml` and everything and import it to an overlay separately.

I think I kinda prefer the first option, because I don't want to risk not importing the quota that is a separate to the namespace and then not enforcing it. On the other hand that's not consistent with the repo layout...

HumairAK: > On the other hand that's not consistent with the repo layout... 
This is my issue, consistency is nice for when we add automation. But I think making an exception in this case for namespaces is fine. Let's go with the first option and see where it takes us, we can always change it up.

HumairAK: updated

tumido (commented):
/lgtm

PR #490, opened 2021-04-07
Add clusterlogforwarder to forward app-logs to kafka

4n4nd:
Related https://github.com/operate-first/apps/issues/232

This will forward app-logs to kafka topic `zero-prod.cluster.app-logs`
Update max message size bytes for that topic to 10MB ([slack discussion on size](https://coreos.slack.com/archives/CB3HXM2QK/p1617720645002700))
More docs on log forwarding to kafka: https://docs.openshift.com/container-platform/4.7/logging/cluster-logging-external.html#cluster-logging-collector-log-forward-kafka_cluster-logging-external

### DIFF
--- a/cluster-logging/overlays/moc/zero/clusterlogforwarders/clusterlogforwarder.yaml
+++ b/cluster-logging/overlays/moc/zero/clusterlogforwarders/clusterlogforwarder.yaml
@@ -10,8 +10,20 @@ spec:
       type: kafka
       url: tcp://odh-message-bus-kafka-bootstrap.opf-kafka.svc:9092/zero-prod.cluster.app-logs
   pipelines:
-    - name: forward-app-logs-to-kafka
+    - name: forward-app-logs
       inputRefs:
         - application
       outputRefs:
         - app-logs
+        # default outputref sends logs to the default CLO ES
+        - default
+    - name: forward-infra-logs
+      inputRefs:
+        - infrastructure
+      outputRefs:
+        - default
+    - name: forward-audit-logs
+      inputRefs:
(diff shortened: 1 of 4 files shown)


cluster-logging/overlays/moc/zero/clusterlogforwarders/clusterlogforwarder.yaml
```diff
@@ -0,0 +1,17 @@
+---
+apiVersion: logging.openshift.io/v1
+kind: ClusterLogForwarder
+metadata:
+  name: instance
```
HumairAK, line 5: Can this be named to something less generic? like `kafka-log-forwarder` maybe?

4n4nd: This is required to be `instance` ([docs](https://docs.openshift.com/container-platform/4.7/logging/cluster-logging-external.html#cluster-logging-collector-log-forward-kafka_cluster-logging-external))

HumairAK: Nevermind I see that it has to be `instance`:

From the docs: 
> The namespace for the ClusterLogForwarder CR must be openshift-logging.

HumairAK:
As per the docs: 

> If a ClusterLogForwarder object exists, logs are not forwarded to the default Elasticsearch instance, unless there is a pipeline with the default output. 

We'll need to add the appropriate inputrefs to the default output so that current es ingestion is not effected.

PR #485, opened 2021-04-07
Introduce Thoth Test and required changes

harshad16:
Introduce Thoth Test and required changes
Signed-off-by: Harshad Reddy Nalla <hnalla@redhat.com>

Introduce one new group and modified older group
- thoth: all members
- thoth-devops: only devops team

modified namespace interaction based on these groups.

PR #494, opened 2021-04-08
fix: Thoth devops group not propagated properly

tumido:
Follow up on #485

I've missed it in the review

/cc @harshad16 @HumairAK @4n4nd


harshad16 (commented):
thank you 
/lgtm

HumairAK (approved):

PR #496, opened 2021-04-08
Do not instantiate operatorgroup with hashed name

larsks:
The OCS operator creates its own operatorgroup with a hashed name. If
we include the operatorgroup manifest in this repository, that results
in multiple operatorgroups in the namespace, which causes the operator
install to fail.


tumido (changes requested):
I think we want to keep track of all operator groups. Removing these doesn't seem like a good solution to me. Let's remove the *hash* from the name instead and cleanup the autocreated ones. 
WDYT?

larsks:
@tumido I'm not sure I follow your suggestion exactly: we can't add the operator-group with the unhashed name, because we'll still end up with multiple operator groups.

Is your suggestion to remove the operatorgroup manifest from the `overlay/moc/*/kustomization.yaml`, but keep the `base/operatorgroups/...` manifest around for now even though we're not using it?

tumido:
Nono, I think we should:

- rename the operator group `s/openshift-storage-dsjl6/openshift-storage-operatorgroup/` in git
- deploy our operator group
- if another operator group gets created by the subscription, remove it from the cluster manually

I think upstream has an undocumented name preference. They advise to create the operator group manually for declarative setups as well:
https://github.com/openshift/ocs-operator/blob/e4c49f073e5ccf106b3d7ad7ac3ed73fdae0d078/deploy/deploy-with-olm.yaml

Let's try using the `openshift-storage-operatorgroup` name and see if we get a new group created next time, WDYT?

tumido (commented):
/lgtm

harshad16 (commented):
/lgtm

HumairAK (approved):

PR #507, opened 2021-04-09
Add quotas to opf-* namespaces

HumairAK:
Adding quotas to most `opf-*` namespaces, the ones currently omitted are: 

- opf-ci-pipelines
- opf-ci-prow
- opf-jupyterhub
- opf-jupyterhub-stage
- opf-kafka
- opf-monitoring
- opf-observatorium

I am open to suggestions if we should add quotas to these namespace, and what they should be set to. 

I have also added a nother tier "x-small" because for some of these, "small" seemed overkill, and I suspect we will see more of such situations. We will need to update the support doc accordingly.

### DIFF
--- a/cluster-scope/components/resourcequotas/x-small/resourcequotas.yaml
+++ b/cluster-scope/components/resourcequotas/x-small/resourcequotas.yaml
@@ -1,0 +1,11 @@
+kind: ResourceQuota
+apiVersion: v1
+metadata:
+  name: x-small
+spec:
+  hard:
+    requests.cpu: '500m'
+    requests.memory: 2Gi
+    limits.cpu: '500m'
+    limits.memory: 2Gi
+    requests.storage: 10Gi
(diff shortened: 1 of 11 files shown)


4n4nd (commented):
/lgtm

HumairAK:
@harshad16 @tumido

harshad16:
These take care of the ci stuff opf-ci-pipelines and opf-ci-prow, and they need more resource than the large quota, l can create custom one for them. however dont restrict them with large quota , as then or services would downgrade.

HumairAK:
@harshad16 sounds good I'll let you add the custom ones!

cluster-scope/components/resourcequotas/x-small/resourcequotas.yaml
```diff
@@ -0,0 +1,11 @@
+kind: ResourceQuota
+apiVersion: v1
+metadata:
+  name: x-small
```
tumido, line 4: Can you add this new resource quota to the docs as well please?

tumido (commented):
One thing to ~change~ document, other than that it looks good

cluster-scope/components/resourcequotas/x-small/resourcequotas.yaml
HumairAK: Done: https://github.com/operate-first/support/pull/200

tumido (commented):
/lgtm

4n4nd (approved):

PR #516, opened 2021-04-12
set quota limits for thoth-* namespaces

harshad16:
set quota limits for thoth-* namespaces

Related-to: https://github.com/operate-first/support/issues/196

### DIFF
--- a/cluster-scope/base/namespaces/thoth-amun-api-prod/kustomization.yaml
+++ b/cluster-scope/base/namespaces/thoth-amun-api-prod/kustomization.yaml
@@ -10,3 +10,5 @@ resources:
 components:
   - ../../../components/project-admin-rolebindings/thoth-devops
   - ../../../components/monitoring-rbac
+  - ../../../components/limitranges/default
+  - ../../../components/resourcequotas/medium
--- a/cluster-scope/base/namespaces/thoth-amun-inspection-prod/kustomization.yaml
+++ b/cluster-scope/base/namespaces/thoth-amun-inspection-prod/kustomization.yaml
@@ -6,7 +6,8 @@ namespace: thoth-amun-inspection-prod
 
 resources:
   - namespace.yaml
-
+  - resourcequota.yaml
 components:
   - ../../../components/project-admin-rolebindings/thoth-devops
   - ../../../components/monitoring-rbac
+  - ../../../components/limitranges/default
--- a/cluster-scope/base/namespaces/thoth-amun-inspection-prod/resourcequota.yaml
+++ b/cluster-scope/base/namespaces/thoth-amun-inspection-prod/resourcequota.yaml
@@ -1,0 +1,9 @@
+kind: ResourceQuota
+apiVersion: v1
+metadata:
+  name: custom
+spec:
+  hard:
+    limits.cpu: '16'
+    limits.memory: 32Gi
+    requests.storage: 40Gi
--- a/cluster-scope/base/namespaces/thoth-backend-prod/kustomization.yaml
+++ b/cluster-scope/base/namespaces/thoth-backend-prod/kustomization.yaml
@@ -6,7 +6,9 @@ namespace: thoth-backend-prod
 
 resources:
   - namespace.yaml
+  - resourcequota.yaml
 
 components:
   - ../../../components/project-admin-rolebindings/thoth-devops
   - ../../../components/monitoring-rbac
+  - ../../../components/limitranges/default
--- a/cluster-scope/base/namespaces/thoth-backend-prod/resourcequota.yaml
+++ b/cluster-scope/base/namespaces/thoth-backend-prod/resourcequota.yaml
@@ -1,0 +1,9 @@
+kind: ResourceQuota
+apiVersion: v1
+metadata:
+  name: custom
+spec:
+  hard:
+    limits.cpu: '24'
+    limits.memory: 48Gi
+    requests.storage: 30Gi
--- a/cluster-scope/base/namespaces/thoth-bots-prod/kustomization.yaml
+++ b/cluster-scope/base/namespaces/thoth-bots-prod/kustomization.yaml
@@ -10,3 +10,5 @@ resources:
 components:
   - ../../../components/project-admin-rolebindings/thoth-devops
   - ../../../components/monitoring-rbac
+  - ../../../components/limitranges/default
+  - ../../../components/resourcequotas/medium
--- a/cluster-scope/base/namespaces/thoth-deployment-examples/kustomization.yaml
+++ b/cluster-scope/base/namespaces/thoth-deployment-examples/kustomization.yaml
@@ -10,3 +10,5 @@ resources:
 components:
   - ../../../components/project-admin-rolebindings/thoth
   - ../../../components/monitoring-rbac
+  - ../../../components/limitranges/default
+  - ../../../components/resourcequotas/medium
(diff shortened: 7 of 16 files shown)


tumido (changes requested):
@harshad16 can you lower the CPU quota please? I don't think Thoth should be setting it's CPU limit at more than half of the whole cluster capacity (you're setting the total limit for all Thoth namespaces summed up at 152 cores of 284 cores available in the whole cluster) while your actual usage is never even close to 10 CPU in total for all the namespaces together.

I actually don't think you need a custom quota, you will be totally fine with one of the tiers. If there's a problem you should tune your CPU requests on deployments/pods to match the actual usage more closely.

Let me know if I'm wrong about this. I just think that enabling Thoth via quota to reserve the same amount of CPU as we're already having troubles to handle at https://github.com/operate-first/support/issues/196 doesn't solve the issue.  

image

https://grafana-route-opf-monitoring.apps.zero.massopen.cloud/explore?orgId=1&left=%5B%22now-2d%22,%22now%22,%22opf-observatorium-thanos%22,%7B%22expr%22:%22sum(pod:container_cpu_usage:sum%7Bnamespace%3D~%5C%22thoth-.*%5C%22%7D)%22,%22hide%22:false%7D,%7B%22ui%22:%5Btrue,true,true,%22none%22%5D%7D%5D

harshad16:
@tumido  thanks for the review and comments.
**question**: limit resource is not requesting resource, right? i.e we can reach that limit and not extend it but not necessarily consume that much all the time.

the custom resource quota is added to the user request serving namespace, so it would depend on the number of user requests incoming. I can reduce the resource limit request if this seems high. 
please let me know.

tumido:
> question: limit resource is not requesting resource, right? i.e we can reach that limit and not extend it but not necessarily consume that much all the time.

Correct. On the other hand it also allows you to reserve up to this limit and not actually use it. And we want the CPUs to be used, not just reserved.

It also doesn't make much sense to me "just" to set the limits sky high. Ideally the limits are slightly above the application's actual consumption at its max load. And if it reaches those limits due to increased traffic, more users, you name it, you just request more. But you have it backed by actually using the resources.

I'm just trying to point out that it doesn't make sense to have a limit at 152 if metrics and usage tracking say you never reach using 10  in total ever. Your limit shouldn't be 15x the usage.

What about we do this: Let's set your limit to some low value and if you start hitting it due to actual usage, we will increase it. If you start hitting the limit because of your reservation (but no usage), you lower the reservation.

harshad16:
> Correct. On the other hand it also allows you to reserve up to this limit and not actually use it. And we want the CPUs to be used, not just reserved.

So according to my understanding `limits` just state that a resource can be reached till this point, and doesn't **reserves** any resource, unlike the `request`. that the reason i have set these limits, our service can reach till here, but will not block anyone by reserving it. If my understanding is not right this is good to know.

I completely agree with you what you have said below. i have no issues in reducing the limits, however my thoughts were that it **is not** blocking any resources.  

> It also doesn't make much sense to me "just" to set the limits sky high. Ideally the limits are slightly above the application's actual consumption at its max load. And if it reaches those limits due to increased traffic, more users, you name it, you just request more. But you have it backed by actually using the resources.
> 
> I'm just trying to point out that it doesn't make sense to have a limit at 152 if metrics and usage tracking say you never reach using 10 in total ever. Your limit shouldn't be 15x the usage.

> What about we do this: Let's set your limit to some low value and if you start hitting it due to actual usage, we will increase it. If you start hitting the limit because of your reservation (but no usage), you lower the reservation.

i will check on the application and reduces the limits as you suggested.
however please correct me if my thoughts above are incorrect, this would be good for my learning as well.
thanks again for reviewing.

HumairAK:
> So according to my understanding limits just state that a resource can be reached till this point, and doesn't reserves any resource, unlike the request. that the reason i have set these limits, our service can reach till here, but will not block anyone by reserving it. If my understanding is not right this is good to know. 

Yes you are right: 

> When you specify a resource limit for a Container, the kubelet enforces those limits so that the running container is not allowed to use more of that resource than the limit you set. The kubelet also reserves at least the request amount of that system resource specifically for that container to use. 

From [k8s docs](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#how-pods-with-resource-requests-are-scheduled). We still want limits to be close to requests, so that we don't get blind sided by users using too many resources, and we can assure some guarantees on how many resources are available on the cluster.

tumido:
> So according to my understanding limits just state that a resource can be reached till this point, and doesn't reserves any resource, unlike the request. that the reason i have set these limits, our service can reach till here, but will not block anyone by reserving it. If my understanding is not right this is good to know.

Yes that is true, however additionally to that you have a mapping between Pod resource requests/limits and the total resource requests/limits specified in the quota.
https://kubernetes.io/docs/concepts/policy/resource-quotas/#compute-resource-quota

Setting a very high limit on the quota still means you *can* make your pods to reserve all the CPU up to this limit if you set your pod requests to be the same as limits.

Your quota basically says that the namespaces can limit up to 152 cores in total, while the actual reservation can be anywhere below this value, like 150 for example. That makes the whole quota quite ineffective.

Our intention with the quota is to close the gap on this nonsense:
  
image

We're currently using 15% of the cluster capabilities while additional **75%** is reserved and forced to idle. 3/4 of the whole cpu power on the cluster is idle. That's **6.75** nodes are doing nothing but can't be used either. :exploding_head: 

That's why the cluster is in constant CPU pressure, because users force the system to idle. And only if we set the quota close the actual usage, we can mitigate that.

HumairAK:
any update on this?

harshad16:
apologies for the delay.
I wanted to document all these resource requirements on Project thoth side.
I have updated a little bit here: https://github.com/thoth-station/thoth-application/blob/master/docs/environments.md#resource-consumptions , will complete this by this week.

I have reduced the resources, after talking to the team.
we run workflow-based resources in 3 namespaces (thoth-amun-inspection-prod, thoth-backend-prod, thoth-middletier-prod), out of which 2 are used for user-request responses (thoth-amun-inspection-prod, thoth-backend-prod). we would retain the currently defined limits for now and another namespace (thoth-middletier-prod) is used for data-ingestion, we have reduced the resource for it, we would use the stage environment for extensive ingestion.

As the request comes in random periodic order, we would like to have the defined resources so we could at least serve based on it. (Doesn't mean we would block the request resources).

For the other namespace, all resources are set based on the current consumption. 
we are working on reducing the resources required for the application, so we can further reduce it (however it would be a gradual process).
Thanks for the patience.

tumido (commented):
Thank you @harshad16 This is a huge improvement. :+1: 

I hope we can continue in this trend and push the limits even further down in the future. Maybe if we can integrate your bots an prow with our logging stores (Loki), we don't have to retain the Prow and other CI pods after they've exited - that should allow us to push it a lot more.

Do you have any umbrella tracking issue for this whole adventure in Thoth?

/lgtm

harshad16:
> I hope we can continue in this trend and push the limits even further down in the future. Maybe if we can integrate your bots an prow with our logging stores (Loki), we don't have to retain the Prow and other CI pods after they've exited - that should allow us to push it a lot more.
> 
> Do you have any umbrella tracking issue for this whole adventure in Thoth?

we have an issue with it https://github.com/thoth-station/thoth-application/issues/675
we can refine and create a specific one for the prow job. 
Thanks for the patience.

HumairAK:
Awesome @harshad16 thanks for the update, looks like this is progressing well, I like the documentation!

tumido (approved):
/lgtm
/approve

harshad16:
the check is taking time , as the component are unable to pull the image
https://console-openshift-console.apps.zero.massopen.cloud/k8s/ns/opf-ci-prow/pods/prow-controller-manager-84778d569d-xwjnc/events

tumido:
That is fine @harshad16 that's probably due to the cluster upgrade. Don't worry about it right now :slightly_smiling_face:

tumido:
Merged and applied @harshad16  :tada: :rocket:

PR #523, opened 2021-04-12
Onboard codait-advo team

4n4nd:
Resolves https://github.com/operate-first/support/issues/202

### DIFF
--- a/cluster-scope/base/namespaces/codait-advo/kustomization.yaml
+++ b/cluster-scope/base/namespaces/codait-advo/kustomization.yaml
@@ -1,0 +1,13 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: codait-advo
+
+resources:
+- namespace.yaml
+
+components:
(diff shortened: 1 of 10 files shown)


cluster-scope/base/namespaces/codait-advo/kustomization.yaml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: codait-advo
```
harshad16, line 5: is this to be added to zero cluster?
missing reference here: https://github.com/operate-first/apps/blob/master/cluster-scope/overlays/moc/zero/kustomization.yaml

harshad16 (commented):
/lgtm

just saw that namespace was not added to zero cluster kustomize file.

cluster-scope/base/namespaces/codait-advo/kustomization.yaml
4n4nd: good catch :+1:

tumido (approved):
/lgtm
/approve

PR #526, opened 2021-04-13
feat: Add hostpath SCC for UKY

tumido:
Resolves: https://github.com/operate-first/support/issues/201

/hold waiting for service account name from the team


HumairAK:
Added context: this should allow the requester to use `hostPath` volume mounts in their Pod spec, it should allow them only read access.

https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

> The files or directories created on the underlying hosts are only writable by root. You either need to run your process as root in a privileged Container or modify the file permissions on the host to be able to write to a hostPath volume

4n4nd (approved):
/lgtm

PR #530, opened 2021-04-13
feat: Upgrade Zero to 4.7.7

tumido:
SSIA

/hold

PR #531, opened 2021-04-13
feat: Add a basic script for cluster version updates

tumido:
Part of: https://github.com/operate-first/SRE/issues/216

### DIFF
--- a/scripts/update-cluster-version.sh
+++ b/scripts/update-cluster-version.sh
@@ -1,0 +1,30 @@
+#!/bin/sh
+
+set -o errexit
+trap 'echo "Aborting due to errexit on line $LINENO. Exit code: $?" >&2' ERR
+set -o errtrace
+set -o pipefail
+
+if [ $# -lt 1 ] || [ "$#" == "--help" ] || [ "$#" == "-h" ] || [ ! -f "$1" ]; then
+    echo "Change ClusterVersion manifest based on availableUpdates"
+    echo -e "\nUsage: $0 PATH"
+    echo -e "PATH\tPath to the ClusterVersion manifest to upgrade\n"
+    exit 0
+fi
+
+echo "Updating resource file '$1'"
+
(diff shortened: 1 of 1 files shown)


4n4nd (commented):
/lgtm

tumido:
Related to #530

scripts/update-cluster-version.sh
```diff
@@ -0,0 +1,30 @@
...
+    echo -e "\nUsage: $0 PATH"
+    echo -e "PATH\tPath to the ClusterVersion manifest to upgrade\n"
+    exit 0
+fi
+
+echo "Updating resource file '$1'"
+
+cluster_version=$(oc get -o yaml clusterversion/version)
+
+available_updates=$(echo "$cluster_version" | yq e ".status.availableUpdates | length" -)
```
larsks, line 19: What version of `yq` are you using? I've got https://github.com/kislyuk/yq and it fails with the above expressions. Can we use `-o json` and `jq` instead because there's only a single version of `jq` floating around?

tumido: @larsks we use the Golang version https://github.com/mikefarah/yq (it's standalone and easier to bundle to automation etc.) 

It's included in our toolbox as well: https://github.com/operate-first/toolbox/blob/05526cfe02c6844c94d7250c7131ad78bcab021d/Dockerfile#L45-L47

We kinda rely on `yq` for updates to ODH `KfDef`s and we will need it here, later in the script as well -  `jq` can't serialize into YAML, and we track resources as YAMLs in Git.

tumido: Btw, ideally this can be automated and executed from some CI/bot/operator which would open the PR instead of us.

This script is just an interim workaround so we don't have to fiddle with the manifests all the time.

larsks (approved):

4n4nd (approved):

HumairAK (approved):

PR #538, opened 2021-04-15
Add machineconfig to manage selinux context for /var/hpvolumes

larsks:
Ensure that the directory used by the hostpath-provisioner (a)
exists and (b) has an appropriate selinux context.

Deploying a machineconfig will trigger a rolling reboot of cluster
nodes.


larsks:
Because deploying a new `MachineConfig` resource will trigger a rolling reboot of all cluster nodes, I've asked Surbhi and Gagan if they'd like to schedule this explicitly or if we can just deploy it at will.

larsks:
Scheduled for Monday, 4/19, around 10AM.

HumairAK (approved):

tumido (approved):
/lgtm
/approve

PR #542, opened 2021-04-19
Move Prometheus docs from support repo to apps repo

4n4nd:
Add a basic README similar to other components ([here](https://github.com/operate-first/apps/tree/master/docs))

This will make it easier to add this doc on the operatefirst website. (Just need to link the docs [here](https://github.com/operate-first/operate-first.github.io/blob/master/config/users.yaml))


tumido (commented):
/lgtm

HumairAK (approved):

PR #543, opened 2021-04-19
Fix Prometheus doc formatting on operate-first.cloud

4n4nd:
Improper formatting on the operate-first website:
Screenshot from 2021-04-19 13-58-47


HumairAK (approved):

HumairAK:
the markdown logic looks identical to me, approving the pr -- but I wonder if there's some weirdness happening on the renderer side.

PR #539, opened 2021-04-16
Remove JH dashboard CRs

hemajv:
Since we now have the dashboards available in `ODH v1.0.9` we can remove the old CRs

(related: https://github.com/operate-first/SRE/pull/257)

Issue #466, opened 2021-03-31
re-organizing odh "app" folder


HumairAK:
I think in hindsight it may not have been a good idea to mix our odh kustomize manifests with the overrides, because the folder [here][1] has become a bit of a mess. I think we can possibly untangle this by doing the following: 

- Move kfdefs from `apps/odh` to `apps/kfdef` 
- Any resources added via kustomize in `apps/odh` but not through `/odh` should go in a separate app at the root of `apps`
- Leave only overrides in the `odh` folder so it's purpose is clear, and we can better identify changes to ODH that we have made for operate-first vs upstream `odh-manifests` 

WDYT?

[1]:https://github.com/operate-first/apps/tree/master/odh

tumido:
I agree with the apps reorganization for ODH. I also think we can also go a bit further than this:

Initially ODH was our only supported workload. Right now the focus shifted and ODH is one of many. I think it would be only relevant to rename all the `opf-*` namespaces which are ODH specific to use `odh-*` prefix instead and create `odh` user group that can admin those namespaces including upstream ODH and downstream RHODS folks. WDYT? 

Keep in mind that namespaces can't be renamed so we would need to recreate the namespaces and/or remove the workloads if possible. We can also do this in case we will migrate ODH to any other than Zero cluster in the future.

HumairAK:
Agreed
How about we break this into 2 chunks then: 
1) Re-organize structure
2) Move to odh-* namespaces 

2, will probably need to be done in smaller chunks per component, with announcements for services like JH etc.

tumido:
Yup, let's start with the first point.

Once we're done with that we can slowly migrate the namespaces one by one. I'm worried about moving PVCs between namespaces. This needs to be tested first.

HumairAK:
I don't think we're moving to `odh-*` namespaces anytime soon, I actually think there might be some benefit still to having `opf-*` namespaces, to differentiate what _we_ as the admin team manage vs regular users. Anyways I think the current state is fine, going to close -- if someone disagrees please re-open.

PR #544, opened 2021-04-19
Re organize odh kfdefs to separate app

HumairAK:
Furthers: https://github.com/operate-first/apps/issues/466 

This PR does the following: 
1. Moves all non override manifests to the `kfdef` app 
2. Also removes the grafana dashboards overlay from the grafana kfdef (since it was removed from [here][1]) 

I haven't added/removed the way things are deployed, just moved them around for now. I'm not sure if we _only_ want `kfdefs` in this kfdef app, or also the accompanying manifests (like jh imagestreams, pvcs, routes etc.). But we can discuss options (and save that for the later pr possibly). 

You'll notice now that it's a lot more clear how we are extending `odh` in the `odh-manifests` folder. The structure is identical to the source repo, any additionas would emulate this structure. The paths for overrides are also easier to reason about when adding them to the kfdef, for example: [old][2] vs [new][3]. 

The newer version matches the default `kfdef` [version here][4] with a prefix of 'odh-manifests/' added.

[2]: https://github.com/operate-first/apps/blob/master/odh/base/monitoring/kfdef.yaml#L31
[3]: https://github.com/HumairAK/apps/blob/reorg_kfdefs/kfdefs/base/monitoring/kfdef.yaml#L31
[4]: https://github.com/opendatahub-io/odh-manifests/blob/master/kfdef/kfctl_openshift.yaml#L75

[1]: https://github.com/operate-first/apps/pull/539/files


HumairAK:
hrmm, this is going to require us to update https://github.com/operate-first/toolbox/blob/master/scripts/test-kustomize-build#L20 

I'll first let folks review this before sending that pr in.

tumido (commented):
hey.. this is huge! This is a huge cleanup work! :+1: Awesome stuff. :+1: 

/lgmt

4n4nd (commented):
/lgtm

4n4nd (commented):
/lgtm

PR #546, opened 2021-04-20
Add ssh keys to zero cluster

larsks:
Add ssh keys for the operate-first operations team to the `core` user
on the Zero cluster.

NB: Submitting a machineconfig update triggers a rolling reboot of all
the affected cluster nodes.

Closes operate-first/SRE#260


larsks:
Put on hold because machineconfig updates cause a reboot of all cluster nodes. One option for handling this is that (a) I make the changes manually right now, and then (b) we schedule this for whenever we're performing the next openshift upgrade.

cluster-scope/overlays/moc/zero/kustomization.yaml
```diff
@@ -34,6 +34,8 @@ resources:
   - ../../../base/groups/cluster-admins
   - ../../../base/imageregistryconfigs/cluster
   - ../../../base/ingresscontrollers/default
+  - ../../../base/machineconfigs/ssh-keys-controller
+  - ../../../base/machineconfigs/ssh-keys-worker
```
tumido, line 38: Can we move this into `moc/common` so both clusters are accessible maybe?

larsks: Sure.

tumido (commented):
Thank you!

/lgtm

larsks:
@tumido There are two machineconfig manifests (one for controllers, one for workers) split into two separate machineconfig resource directories because I was figuring we would apply this per cluster, and in some environments (e.g., the curator cluster) there are no worker nodes. Since we've moved this into the common overlay, should we just merge this into a single resource in `base/machineconfigs/ssh-keys`?

tumido:
I think you're right @larsks, I think we probably don't expect some people to have access to worker nodes or master nodes only.. If somebody has access to SSH to the nodes, then they should be able to access them all I think. I'd keep it simple. If we ever need to have different SSH keys for workers than master nodes, we can split those manifests...

> in some environments (e.g., the curator cluster) there are no worker nodes

Even `moc-infra` doesn't have worker node, yeah...

Let's merge it. That sounds simpler to maintain to me and it is less confusing.

larsks:
Done. I wish there were a way to avoid having the two duplicate lists, but I don't think this is worth writing a generator.

tumido (commented):
/lgtm

tumido (approved):
/approve

PR #551, opened 2021-04-20
Transfer cd repo docs to apps repo.

HumairAK:
As Per: https://github.com/operate-first/continuous-deployment/issues/113 
Related: https://github.com/operate-first/continuous-deployment/pull/141

So I've moved all the docs from: https://github.com/operate-first/continuous-deployment  that I think belong here instead. I've updated the paths where I could find them. I didn't include any of the scripts for deploying argocd/odh to quicklab/crc, because I don't think 1) anyone uses them, 2) we need them -- but I don't have a hard stance. 

- I moved the PV set up playbook to be under scripts (didn't make sense for it to be in docs imo). 
- You'll alos see a `cluster-resources` kustomization.yaml, this is to make it easier to manually deploy argocd to dev environment w/ cluster wide access. (See the dev setup docs in this pr). 
- I also created a separate section for setting up dev environments, and put the crc/quicklab docs there.

FYI only dev and docs folders are being updated here, no changes to live clusters are added here.


tumido (commented):
/lgtm

4n4nd (commented):
/lgtm

4n4nd (approved):

PR #559, opened 2021-04-23
fix: Turn subscriptions in openshift-operators to automatic

tumido:
`Manual` install plan strategy is messing with other operators with automated install plans installed to the same namespace.

`Automatic` strategy is respected only if all operators use it. Otherwise it acts as manual.

This is causing troubles updating Strimzi provided via ODH as `Automatic`.  

image (3)


4n4nd (commented):
/lgtm

HumairAK (approved):
@larsks fyi

larsks:
Well, I guess that solves the question of how we manage operator updates...

PR #563, opened 2021-04-23
Add explicit allow list for cluster management argocd proj.

HumairAK:
As per argocd 1.8.7 notes: 

```
This release fixed a regression regarding which cluster resources are permitted on the AppProject level. Previous to this fix, after #3960 has been merged, all cluster resources were allowed on project level when neither of the allow or deny lists was defined. However, the correct behavior is to block all resources in this case.

If you have Projects with empty allow and deny lists, but want the associated applications be able to sync cluster resources, you will have to adapt your cluster resources allow lists to explicitly allow the resources.
```


tumido (approved):
/lgtm
/approve

PR #572, opened 2021-04-26
Enable monitoring on a namespace

ipolonsk:
Part of: [operate-first/SRE#216](https://github.com/operate-first/SRE/issues/216 )
fix: https://github.com/operate-first/SRE/issues/277 

The script accepts a namespace and applies '- ../../../components/monitoring-rbac' line to 'components' field with a kustomize cli.

On success print nothing, and on failure print relevant error from kustomize cli. 

Notice that kustomize editing the file and uses specific annotation while listing the components, the '-' is at the first char and not the second on each line.

### DIFF
--- a/scripts/enable-monitoring.sh
+++ b/scripts/enable-monitoring.sh
@@ -1,0 +1,28 @@
+ #!/bin/sh
+
+set -o errexit
+trap 'echo "Aborting due to errexit on line $LINENO. Exit code: $?" >&2' ERR
+set -o errtrace
+set -o pipefail
+
+if [ $# -ne 1 ] || [ "$#" == "--help" ] || [ "$#" == "-h" ]; then
+    echo "Enable monitoring on a namespace"
+    echo -e "\nUsage: $0 NAMESPACE\n"
+    exit 0
+fi
+
+APP_NAME="cluster-scope"
+NAMESPACE=$1
+
+add_rbac_component() {
+	pushd $APP_NAME/base/namespaces/$NAMESPACE > /dev/null
+	kustomize edit add component ../../../components/monitoring-rbac
+	popd > /dev/null
+}
+
+if [ ! -d $APP_NAME/base/namespaces/$NAMESPACE ]; then
+	echo "Namespace '$NAMESPACE' does not exist in $APP_NAME/base/namespace/. Exiting."
+	exit 1
+fi
+
+add_rbac_component
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 ---
-exclude: scripts/onboarding.sh|argocd/overlays/dev/configs|argocd/overlays/moc-infra/configs
+exclude: scripts|argocd/overlays/dev/configs|argocd/overlays/moc-infra/configs
 
 repos:
   - repo: git://github.com/Lucas-C/pre-commit-hooks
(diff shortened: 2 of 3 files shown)


tumido (commented):
/lgtm :rocket: :+1:

4n4nd (commented):
/lgtm

ipolonsk:
Modified the script to be more consistent with the other scripts.
- Changed file name `enable_monitoring.sh` -> `enable-monitoring.sh`
- Changed PATH used from ''../$APP_NAME/.." -> ''$APP_NAME/.."
(So we can run it like the other scripts with `bash scripts/enable-monitoring.sh` directly from the main folder.)
- Changed the check on the number of arguments, now it gets only one argument. (other arguments will not be used)
- Added check if a namespace exists in base, prevent it from trying to navigate to a not existing folder. (return an error for the user to notice).
- Used function to wrap the commands. 

* since I changed the file name git shows this commit as a new file, sorry for the inconvenience.

scripts/enable-monitoring.sh
```diff
@@ -0,0 +1,28 @@
...
+NAMESPACE=$1
+
+add_rbac_component() {
+	pushd $APP_NAME/base/namespaces/$NAMESPACE > /dev/null
+	kustomize edit add component ../../../components/monitoring-rbac
+	popd > /dev/null
+}
+
+if [ ! -d $APP_NAME/base/namespaces/$NAMESPACE ]; then
+	echo "Namespace '$NAMESPACE' is not exists in $APP_NAME/base/namespace/. Exiting."
```
larsks: Should be "does not exist" rather than "is not exists".

larsks (approved):

HumairAK:
awesome, can you also update the docs [here](https://github.com/operate-first/apps/blob/master/docs/odh/prometheus/add_service_monitoring.md#steps) (the first step) to use this script instead?

tumido (changes requested):
requesting changes for science! :thinking:

HumairAK:
can you squash these to 1 commit

scripts/enable-monitoring.sh
```diff
@@ -0,0 +1,28 @@
...
+
+set -o errexit
+trap 'echo "Aborting due to errexit on line $LINENO. Exit code: $?" >&2' ERR
+set -o errtrace
+set -o pipefail
+
+if [ $# -ne 1 ] || [ "$#" == "--help" ] || [ "$#" == "-h" ]; then
+    echo "Enable monitoring on a namespace"
+    echo -e "\nUsage: $0 NAMESPACE\n"
+	exit 0
```
tumido, line 11: can you please indent this the same?
image

tumido (changes requested):
Please change this line `scripts/onboarding.sh` to `scripts`:

https://github.com/operate-first/apps/blob/999d2cacb8efe9f06c6018ae0461d0fda3724d6b/.pre-commit-config.yaml#L2

ipolonsk:
Hey all,
Sorry for some reason I pushed all the latest commits together with my last commit, working on fixing it.

tumido (approved):
/approve
/lgtm
Let's have one more pair of eyes. I like this! Great job @ipolonsk ! :slightly_smiling_face: 

/hold

larsks (approved):

PR #584, opened 2021-04-29
Add ClusterRole/Binding to enable Opf prometheus oauth token review

4n4nd:
The current Prometheus proxy does not allow access to the Prometheus API because the Proxy does not have permissions to check the validity of oauth tokens.
This will give the `prometheus-k8s` serviceaccount in `opf-monitoring` namespace permissions to review oauth tokens.


HumairAK (approved):

tumido (approved):
/approve

PR #589, opened 2021-05-03
update the thoth quota resources with requests setting

harshad16:
update the thoth quota resources with requests setting


tumido (commented):
/lgtm
/approve

Thank you sir! :1st_place_medal:

PR #591, opened 2021-05-03
fix typo in metallb configuration

larsks:
There was a typo in the metallb configuration that rendered the
default address range invalid.


tumido (approved):
/approve

PR #595, opened 2021-05-04
Onboard user.

HumairAK:
Resolves: https://github.com/operate-first/support/issues/220

PR #596, opened 2021-05-04
Add team-pixel grp to common.

HumairAK:
part of: https://github.com/operate-first/apps/pull/595

not including it broke the kustomize build, it should have been caught


tumido (approved):
/lgtm
/approve

PR #588, opened 2021-04-29
[WIP] Add m4d cluster resources to zero cluster.

HumairAK:
Resolves https://github.com/operate-first/support/issues/192

PR #609, opened 2021-05-05
[WIP] Add script for adding cluster resources to cluster

HumairAK:
This PR is broken in to 2 commits: 

1. Fix the structure in cluster-scope/base app, and apply `kustomize edit fix` to the kustomizations in `common` `infra` and `zero` overlays. No actual changes are made to the output of the final `kustomize builds`, and thus no changes to live state. You can test these by comparing the outputs of the builds in this pr with what's on master. 
2. automation script for adding resource + docs 

The script is really minimal, and can definitely be improved/iterated upon (like adding checks for existing manifests, better clean up upon failures, etc.)

### DIFF
--- a/docs/cluster-scope/add_resource_to_cluster.md
+++ b/docs/cluster-scope/add_resource_to_cluster.md
@@ -1,0 +1,22 @@
+# Adding Cluster Resources
+
+To add a cluster resource to a cluster, you will need to know the `${ENVIRONMENT}` and `${CLUSTER}` you would like to add this resource to.
+
+You will also need the url to the raw file (e.g. a raw file from github) or a path to the resource manifest on your local file system.
+
+Once you have that, you can use the following script from this repo to add this resource:
+
+```bash
+# from the root of this repo
+$ ./scripts/add_cluster_resource.sh ${ENVIRONMENT} ${CLUSTER} -f ${PATH}
+
+# OR
+
+$ ./scripts/add_cluster_resource.sh ${ENVIRONMENT} ${CLUSTER} -u ${URL}
+```
+
+Example:
+
+```bash
+$ ./scripts/add_cluster_resource.sh moc zero -f /path/to/file.yaml
+```
--- a/scripts/add_cluster_resource.sh
+++ b/scripts/add_cluster_resource.sh
@@ -1,0 +1,132 @@
+#!/bin/sh
+
+set -o errexit
+trap 'echo "Aborting due to errexit on line $LINENO. Exit code: $?" >&2' ERR
+set -o errtrace
+set -o pipefail
+
+_BN="$(basename "${0}")"
+
+_usage() {
+cat <<HEREDOC
+
+Fetch a cluster resource from github and add it to cluster-scope app.
+
+Usage:
+  ${_BN} [env] [cluster] [flags]
+
+Mandatory flags (mutually exclusive):
+  -f, --file    path to file on local filesystem
+  -u, --url     remote git url for the raw file
+
+Optional Flags:
+  -h, --help    help for this script
+
+Example:
+  ${_BN} moc zero -f /tmp/manifest.yaml
+  ${_BN} moc zero -u https://raw.githubusercontent.com/<github_user>/<repo>/master/<path-to-manifest>/manifest.yaml
+
+HEREDOC
+}
+
+# Keep values sorted alphabetically since `kustomize edit resource add` just appends.
+_sort_kustomize_values(){
+  tmpfile=$(mktemp)
+  yq e -j kustomization.yaml | jq 'map_values(select(. >= []) |= sort)' | yq e -P > ${tmpfile}
+  mv ${tmpfile} ./kustomization.yaml
+  kustomize edit fix
+}
+
+_get_plural(){
+  kind=${1}
+  apiVersion=${2}
+  cat ./scripts/api-resources.txt  | grep "\b${kind}\b" | grep ${apiVersion} | awk '{print $1}'
+}
+
+_add_resource(){
+  env=${1}
+  cluster=${2}
+  src=${3}
+  kind=`yq e '.kind' ${src}`
+  apiVersion=`yq e '.apiVersion' ${src}`
+  resource_dir=`_get_plural ${kind} ${apiVersion}`
+  filename=$(echo "$kind" | awk '{print tolower($0)}').yaml
+  resource_name=`yq e '.metadata.name' ${src}`
+  base_dest="cluster-scope/base/${resource_dir}/${resource_name}"
+  overlay_dest="cluster-scope/overlays/${env}/${cluster}"
+
+  mkdir -p ${base_dest}
+  pushd ${base_dest} > /dev/null
+  kustomize init
+  cp ${src} ./${filename}
+  kustomize edit add resource ./${filename}
+
+  echo "Resource ${kind}/${resource_name} added to ${base_dest}"
+
+  popd > /dev/null
+  pushd ${overlay_dest} > /dev/null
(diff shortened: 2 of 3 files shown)


scripts/add_cluster_resource.sh
```diff
@@ -0,0 +1,140 @@
...
+  tmpfile=$(mktemp)
+  yq e -j kustomization.yaml | jq 'map_values(select(. >= []) |= sort)' | yq e -P > ${tmpfile}
+  mv ${tmpfile} ./kustomization.yaml
+  kustomize edit fix
+}
+
+_get_plural(){
+  kind=${1}
+  apiVersion=${2}
+  kubectl api-resources  | grep "\b${kind}\b" | grep ${apiVersion} | awk '{print $1}'
```
HumairAK, line 45: Got a little lazy here, since I didn't want to add a lot of logic for finding plurals for resources that end with words like `policy -> policies`. Open to suggestions.

scripts/add_cluster_resource.sh
```diff
@@ -0,0 +1,140 @@
...
+
+Example:
+  ${_BN} moc zero -f /tmp/manifest.yaml
+  ${_BN} moc zero -u https://raw.githubusercontent.com/<github_user>/<repo>/master/<path-to-manifest>/manifest.yaml
+
+HEREDOC
+}
+
+# Keep values sorted alphabetically since `kustomize edit resource add` just appends.
+_sort_kustomize_values(){
```
HumairAK, line 35: `kustomize edit add ....` doesn't care about alphabetical sorting, so this is a potentially convoluted way of ensuring it stays sorted, open to better methods.

larsks:
I'm a little confused by some of the renames. E.g.:

- cluster-scope/base/imageregistryconfigs/cluster/config.yaml `->` cluster-scope/base/configs/cluster/config.yaml
- cluster-scope/base/podsecuritypolicy/metallb-system-controller/podsecuritypolicy.yaml `->` cluster-scope/base/podsecuritypolicies/controller/podsecuritypolicy.yaml

These seem to be removing critical information from the path. What if we need another `podsecuritypolicy` for a `controller` that isn't related to MetalLB?

docs/cluster-scope/add_resource_to_cluster.md
```diff
@@ -0,0 +1,24 @@
+# Adding Cluster Resources
+
+To add a cluster resource to a cluster, you will need to know the `${ENVIRONMENT}` and `${CLUSTER}` you would like to add this resource to.
+
+You will also need the url to the raw file (e.g. a raw file from github) or a path to the resource manifest on your local file system.
+
+You will also need to be logged in to the cluster that supports this api resource. This is done to fetch the plural name of this resource from the cluster, no actual changes are made to the live environment.
```
larsks, line 7: I am a little sad that this solution requires:

1. A pre-existing file
1. A network connection
1. Active cluster authentication

That's a lot of prerequisites. The first item, in particular, really upends the typical model (create the file in the appropriate location, then `git add` when done).

I guess my preference here would be to see the organization enforced by a precommit check, and allow the script to be run without requiring remote connectivity (so if I give it the wrong metadata and create a file in the wrong place, the precommit check will yell at me either locally or when I submit the pr).

larsks: If we're only using the remote connection to determine the plural of a resource name, what if just cache that in the repository? The available api resources isn't going to change all that often, so we can update the cached version when necessary.

scripts/add_cluster_resource.sh
larsks: This is really beginning to smell like "we should write this in a language with native support for YAML".

docs/cluster-scope/add_resource_to_cluster.md
HumairAK: > That's a lot of prerequisites. The first item, in particular, really upends the typical model (create the file in the appropriate location, then git add when done).

This wouldn't really be for that particular usecase, but for more situations where we get requests from users to add cluster-resources, like: https://github.com/operate-first/apps/pull/588

Here we are [asked by this particular team](https://github.com/operate-first/support/issues/192#issuecomment-825206787) to add a bunch of resources found in their git repo, having to create kustomization/new files for all these files are made easier if I can just feed the link and have it be auto created. 

The manual process of adding your own file would still be an option. I agree we can also add ci tests to ensure integrity, and we should. 

> If we're only using the remote connection to determine the plural of a resource name, what if just cache that in the repository? The available api resources isn't going to change all that often, so we can update the cached version when necessary. 

Yeah we can do that.

HumairAK:
> These seem to be removing critical information from the path. What if we need another podsecuritypolicy for a controller that isn't related to MetalLB? 

@larsks if we keep it the way it is currently, then these become exceptions / snowflakes that affect the overall integrity of the structure, making it to add future automations. But I understand the concerns...

Can these resource names be changed to be more specific, or do they have be named as such? (i.e. controller/speaker/cluster etc.) If we can rename them, then that should be sufficient no?

HumairAK:
I think this has introduced some gaping holes in our repo structure and requires a broader discussion so as not to rush things. Let's pause this PR for now and discuss these points in a different thread.

scripts/add_cluster_resource.sh
tumido: I don't have an better idea. We're sorting arrays here.. which are usually ordered :confused: I agree that we might want to start the golang cli finally. :smile:

tumido (commented):
Just a partial review for now from me. I'll get back to this later. Good effort though! :+1:

scripts/add_cluster_resource.sh
```diff
@@ -0,0 +1,140 @@
...
+  apiVersion=${2}
+  kubectl api-resources  | grep "\b${kind}\b" | grep ${apiVersion} | awk '{print $1}'
+}
+
+_add_resource(){
+  env=${1}
+  cluster=${2}
+  src=${3}
+  kind=`cat ${src} | yq e '.kind' -`
+  apiVersion=`cat ${src} | yq e '.apiVersion' -`
```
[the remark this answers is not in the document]
HumairAK: duh

larsks:
https://github.com/operate-first/apps/issues/610 for discussion on some of the issues turned up by this pr.

HumairAK:
Updated the pr to just include the script, and we'll make this PR dependent on the issue/resolution that comes from #610

scripts/add_cluster_resource.sh
HumairAK: I am not against doing this

tumido: totally agree

tumido:
> I'm a little confused by some of the renames. E.g.:
> 
> * cluster-scope/base/imageregistryconfigs/cluster/config.yaml `->` cluster-scope/base/configs/cluster/config.yaml
> * cluster-scope/base/podsecuritypolicy/metallb-system-controller/podsecuritypolicy.yaml `->` cluster-scope/base/podsecuritypolicies/controller/podsecuritypolicy.yaml
> 
> These seem to be removing critical information from the path. What if we need another `podsecuritypolicy` for a `controller` that isn't related to MetalLB?

Then the resource has to have a different name. If the podsecuritypolicy has name `controller` there can be only one such resource on the cluster. Resource name usually doesn't effect its function, but when a resource kind is cluster-scoped the name has to be unique.

When it comes to the `config` kind, this is a bit troublesome, since there are multiple default OpenShift API groups with a kind named `Config` enforcing resource to be named `cluster` so.. we really need to get the API group somehow into the name here.

HumairAK:
@larsks has a python version that does this a lot better, I think we should opt to use this: https://github.com/larsks/opf_organizer 
instead

PR #613, opened 2021-05-05
Add network interfaces for access to nfs server

larsks:
Add network interfaces for access to nfs server

This adds vlan interfaces to provide access to an nfs server for
additional storage. We use node labels to select the appropriate
interface for a given node.

### DIFF
--- a/cluster-scope/overlays/moc/zero/kustomization.yaml
+++ b/cluster-scope/overlays/moc/zero/kustomization.yaml
@@ -131,6 +131,7 @@ resources:
   - ../../../base/subscriptions/serverless-operator
   - ../../../base/subscriptions/web-terminal
   - clusterversion.yaml
+  - nodenetworkconfigurationpolicies/moc-nfs-network.yaml
 
 generators:
   - secret-generator.yaml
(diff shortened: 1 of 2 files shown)


larsks:
I wasn't sure how to handle the node labels in the context of our management framework. For now I've added them automatically by running the following playbook:

```
- hosts: openshift
  tasks:
    - delegate_to: localhost
      k8s:
        state: present
        definition:
          apiVersion: v1
          kind: Node
          metadata:
            labels:
              moc/primary-interface: "{{ ansible_default_ipv4.interface }}"
            name: "{{ ansible_hostname }}"
```

I wasn't sure we wanted to import entire `Node` resources into our repository for this change (I feel like these may be managed dynamically by OpenShift more than other resources). I wish we had a way to provide a patch to an existing resource (e.g., something that could be applied using `oc patch` instead of `oc apply`).

tumido:
@larsks seems like we may want to consider something like https://github.com/redhat-cop/resource-locker-operator for locking partial manifests along with ArgoCD. I've started an issue for that here: https://github.com/operate-first/support/issues/235

tumido:
Related to: https://github.com/operate-first/SRE/issues/280

tumido (commented):
/lgtm

HumairAK (approved):
/lgtm

PR #617, opened 2021-05-06
swap interfaces for storage connection

larsks:
on second thought, we'd like to move the storage connection to the
non-public interface in order to avoid competition with other traffic
on the public interfaces.


HumairAK (approved):
/lgtm

larsks:
Pausing just a second while I double check that the nodes all have the interface config I think they do.

PR #619, opened 2021-05-08
Add jupyterhub-idle-culler 

HumairAK:
Resolves: https://github.com/operate-first/SRE/issues/229

Using buildconfig/deployment configs here, which may be controversial. Not opposed to adding a Dockerfile and pushing to quay, though we'd need to decide where that would go. This just seemed like a quick and easy way to get it this out the door. 

I've set idle timeout to 3 days, open to suggestions here. 

The jupyterhub token is created via Jupyterhub UI using my (admin) account. The docs for jupyterhub-idle-culler also seemed to suggest this. 

I think we should try to encourage upstream to introduce this as a [jupyterhub managed service](https://jupyterhub.readthedocs.io/en/stable/reference/services.html#launching-a-hub-managed-service) so we don't have to rely on our api tokens.

### DIFF
--- a/jupyterhub-idle-culler/base/buildconfig.yaml
+++ b/jupyterhub-idle-culler/base/buildconfig.yaml
@@ -3,15 +3,15 @@ apiVersion: build.openshift.io/v1
 metadata:
   name: jupyterhub-idle-culler
 spec:
-    source:
-      dockerfile: |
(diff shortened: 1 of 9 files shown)


tumido (approved):
/lgtm
/approve

PR #620, opened 2021-05-08
default storageclass should always be set per-cluster

larsks:
we were marking ocs-storagecluster-cephfs as the default storageclass
in base, which makes it impossible to select a different default
storageclass in an overlay if that storageclass is included in
the overlay configuration.


tumido (commented):
Do we want to be extra pedantic about the file names? If yes, we may consider renaming `set-default-storageclass_patch.yaml` to `<storage_class_name>_patch.yaml`

tumido:
@larsks I don't have a preference either, that's why it was an open question. :slightly_smiling_face: 

Maybe name it: `resource_purpose_patch.yaml` or am I overthinking it? @HumairAK do you have any preference?

HumairAK:
I think having the purpose is useful. But the other manifests in the overlays have been named using the format `<resource_name>_patch.yaml` so I think we should be consistent, and if we want to have the purpose, that should be a separate PR reflecting the change on all the patch manifests.

tumido:
On the other hand.. if we're patching a different resource with the same purpose as in another overlay it might get confusing as well. I think the resource name is also important...

tumido (approved):
It doesn't make it impossible to select a different one as default in an overlay, but it's way much harder than it should be.

/lgtm
/approve

PR #621, opened 2021-05-09
fix/resourcequota unique names

larsks:
- fix openshift-cnv namespace
- give custom resourcequotas unique names

PR #622, opened 2021-05-10
refactor/adr0015

larsks:
- default storageclass should always be set per-cluster
- give custom resourcequotas unique names
- Reorganize cluster-scope/base per adr 0015
- reorganize dev overlay per adr 0015
- reorganize moc/common overlay per adr 0015
- reorganize moc/curator overlay per adr 0015
- reorganize moc/infra overlay per adr 0015
- reorganize moc/zero overlay per adr 0015
- Restore resource quotas


cluster-scope/base/apiextensions.k8s.io/customresourcedefinitions/kfdefs.kfdef.apps.kubeflow.org/customresourcedefinition.yaml
```diff
@@ -17,10 +17,14 @@ spec:
       description: KfDef is the Schema for the kfdefs API
       properties:
         apiVersion:
-          description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
+          description: 'APIVersion defines the versioned schema of this representation
```
tumido, line 20: I think this was formatted by accident. Can we please not change the copies of upstream CRD manifests?

cluster-scope/base/core/namespaces/argocd/namespace.yaml
```diff
@@ -1,9 +1,8 @@
----
 apiVersion: v1
 kind: Namespace
 metadata:
   annotations:
-    openshift.io/display-name: "Operate First: ACM"
+    openshift.io/display-name: 'Operate First: Observatorium operator'
```
tumido, line 5: This is not true

tumido (changes requested):
I've stopped reviewing after a bit.

I have few outstanding problems with this PR:

- Re-formatting and reordering CRDs is changing them when compared to the upstream manifest we've copied it from - raw manifest diff is not 0 anymore + we can't use checksums to check for updates
- Python has obviously different YAML standard preference compared to Golang and others:
    - single vs double quotes, not a big deal, but totally unnecessary
    - using quotes for multiline strings instead of `|` and `>` making manifests much harder to read and format (requirement for blank line to do a new line, etc)


Diff on this PR is not zero. It's introducing new changes that if left unnoticed and merged may cause problems:
- Renames namespace on infra cluster
- Creates new role and rolebinding in `argocd` namespace
- Renames `OperatorGroup`

Diffs: 

Zero
```diff
290a291,302
>   name: openshift-monitoring
> spec: {}
> ---
> apiVersion: v1
> kind: Namespace
> metadata:
>   name: openshift-operators
> spec: {}
> ---
> apiVersion: v1
> kind: Namespace
> metadata:
749c761
<   name: custom
---
>   name: opf-datacatalog-custom
853c865
<   name: custom
---
>   name: thoth-amun-inspection-prod-custom
866c878
<   name: custom
---
>   name: thoth-backend-prod-custom
905c917
<   name: custom
---
>   name: thoth-frontend-prod-custom
918c930
<   name: custom
---
>   name: thoth-graph-prod-custom
944c956
<   name: custom
---
>   name: thoth-middletier-prod-custom
959c971
<   name: custom
---
>   name: thoth-test-core-custom
11926c11938
<   name: openshift-cnv-rn5jf
---
>   name: openshift-cnv
```
Infra
```diff
5c5
<     openshift.io/display-name: 'Operate First: ACM'
---
>     openshift.io/display-name: 'Operate First: Observatorium operator'
7c7
<   name: acm
---
>   name: argocd
2511a2512,2534
> kind: Role
> metadata:
>   name: prometheus-monitoring
>   namespace: argocd
> rules:
> - apiGroups:
>   - ""
>   resources:
>   - services
>   - endpoints
>   - pods
>   verbs:
>   - get
>   - list
>   - watch
> - apiGroups:
>   - ""
>   resources:
>   - configmaps
>   verbs:
>   - get
> ---
> apiVersion: rbac.authorization.k8s.io/v1
2602c2625
<   namespace: acm
---
>   namespace: argocd
2614a2638,2651
>   name: prometheus-monitoring
>   namespace: argocd
> roleRef:
>   apiGroup: rbac.authorization.k8s.io
>   kind: Role
>   name: prometheus-monitoring
> subjects:
> - kind: ServiceAccount
>   name: prometheus-k8s
>   namespace: opf-monitoring
> ---
> apiVersion: rbac.authorization.k8s.io/v1
> kind: RoleBinding
> metadata:
3126c3163
<   name: openshift-cnv-rn5jf
---
>   name: openshift-cnv
```
Curator - OK

larsks:
> Diff on this PR is not zero. It's introducing new changes that if left unnoticed and merged may cause problems:

Diff on this PR is *explicitly* not zero. Some of the cases you highlighted stem from the PRs on which this one is based, which were called out in the description.

Some of the other changes I'm not sure where they crept in, but the easy fix is just not to rewrite manifests when relocating them.

larsks:
> but the easy fix is just not to rewrite manifests when relocating them.

...except, now that I recall, that's not possible. In situations where there are multiple manifests in a single file, we need to unpack those to get them in the right place. This requires de-serializing the manifests and then re-serializing them when writing if we're going to automate the process.

The alternative would be to just do everything by hand, I guess.

cluster-scope/base/core/namespaces/argocd/namespace.yaml
larsks: That appears to be something you committed back in January.

```
$ git blame cluster-scope/base/namespaces/argocd/namespace.yaml
[...]
caf8ebe9 namespaces/base/observatorium-operator/namespace.yaml (<7453394+tumido@users.noreply.github.com> 2021-01-12 19:08:08 +0100 6)     openshift.io/display-name: "Operate First: Observatorium operator"
[...]
```

larsks: Ah, interesting, `master` has both `namespaces/argocd.yaml` and `namespaces/acm.yaml`, and both directories have a `namespace.yaml` with the same namespace name defined.

larsks:
Diff between master and this PR for...

## dev

This changes goes away when we land #621.

```
--- out-master.yaml	2021-05-10 13:35:30.677084320 -0400
+++ out-refactored.yaml	2021-05-10 13:35:32.109114587 -0400
@@ -108,7 +108,7 @@
 apiVersion: v1
 kind: ResourceQuota
 metadata:
-  name: custom
+  name: opf-datacatalog-custom
   namespace: opf-datacatalog
 spec:
   hard:
```

## moc/curator

No changes.

## moc/infra

No changes.

## moc/zero

Most of these changes go away when we land #621.

The remaining changes are due to missing namespace resources in our
current configuration.

```
--- out-master.yaml	2021-05-10 13:31:00.198367635 -0400
+++ out-refactored.yaml	2021-05-10 13:31:22.105830659 -0400
@@ -288,6 +288,18 @@
 apiVersion: v1
 kind: Namespace
 metadata:
+  name: openshift-monitoring
+spec: {}
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-operators
+spec: {}
+---
+apiVersion: v1
+kind: Namespace
+metadata:
   annotations:
     openshift.io/display-name: Openshift Operators Redhat
     openshift.io/node-selector: ""
@@ -746,7 +758,7 @@
 apiVersion: v1
 kind: ResourceQuota
 metadata:
-  name: custom
+  name: opf-datacatalog-custom
   namespace: opf-datacatalog
 spec:
   hard:
@@ -850,7 +862,7 @@
 apiVersion: v1
 kind: ResourceQuota
 metadata:
-  name: custom
+  name: thoth-amun-inspection-prod-custom
   namespace: thoth-amun-inspection-prod
 spec:
   hard:
@@ -863,7 +875,7 @@
 apiVersion: v1
 kind: ResourceQuota
 metadata:
-  name: custom
+  name: thoth-backend-prod-custom
   namespace: thoth-backend-prod
 spec:
   hard:
@@ -902,7 +914,7 @@
 apiVersion: v1
 kind: ResourceQuota
 metadata:
-  name: custom
+  name: thoth-frontend-prod-custom
   namespace: thoth-frontend-prod
 spec:
   hard:
@@ -915,7 +927,7 @@
 apiVersion: v1
 kind: ResourceQuota
 metadata:
-  name: custom
+  name: thoth-graph-prod-custom
   namespace: thoth-graph-prod
 spec:
   hard:
@@ -941,7 +953,7 @@
 apiVersion: v1
 kind: ResourceQuota
 metadata:
-  name: custom
+  name: thoth-middletier-prod-custom
   namespace: thoth-middletier-prod
 spec:
   hard:
@@ -956,7 +968,7 @@
 apiVersion: v1
 kind: ResourceQuota
 metadata:
-  name: custom
+  name: thoth-test-core-custom
   namespace: thoth-test-core
 spec:
   hard:
```

tumido (commented):
/lgtm

This is huge change set. Tested locally, looks good to me.

HumairAK:
/lgtm 

also tested locally, diff yielded no change (comparing kustomize build outputs from master and this pr).

tumido (approved):
/approve

PR #628, opened 2021-05-11
Acme operator

tumido:
Resolves: https://github.com/operate-first/support/issues/237
Part of: https://github.com/operate-first/support/issues/199

I've renamed most of the resources from `openshift-acme` to `acme-operator` since most things with `openshift-*` prefix usually refer to OpenShift-owned resources.Hopefully this lowers the confusion here.

It turned out that we can't use CertManager to secure OpenShift routes, since they don't want to support vendor specific resource (`Route`) and OpenShift can't make the routes to load certs from secrets for many years, so we need 2 different solutions in place - one for routes and another one for the rest.

References in the issue above.

### DIFF
--- a/docs/issuing_certificates.md
+++ b/docs/issuing_certificates.md
@@ -1,0 +1,19 @@
+# Issuing certificates
+
+## Securing OpenShift `Route`s with ACME (Let's Encrypt)
+
+We manage and deploy cluster wide deployment of [OpenShift ACME](https://github.com/tnozicka/openshift-acme) controller. This operator facilitates [Let's Encrypt](https://letsencrypt.org/) certificate provisioning in any namespace on the cluster. Any `Route` resource is eligible, please use the following annotation to mark the route as managed by this operator:
+
+```yaml
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  annotations:
+    kubernetes.io/tls-acme: "true"
+```
+
+Annotating a route will result with an additional temporary route and 2 pods being created in your namespace. Once the certificate is issued these pods will be removed. This process is repeated to renew certificates later on.
+
+## Issuing certificates via Cert Manager
+
+For other workloads that rely on SSL/TLS certificates, we will be able to provision certificates via Cert Manager shortly. This is currently work in progress.
--- a/acme-operator/base/deployment.yaml
+++ b/acme-operator/base/deployment.yaml
@@ -1,0 +1,24 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: openshift-acme
+spec:
+  selector:
+    matchLabels:
+      app: openshift-acme
+  replicas: 2
+  strategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: openshift-acme
+    spec:
+      serviceAccountName: openshift-acme
+      containers:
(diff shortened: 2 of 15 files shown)


larsks (approved):
The inconsistent use of YAML document start markers bothers me, but not in a way that I think we need to change anything.

tumido:
@larsks which inconsistency? You mean the

```
a:
- b
```
vs
```
a:
  - b
```

Yeah.. for some reasons some formatters prefer one over the other, kustomize is stubborn in another way... Both ways are valid according to the YAML standard, while the one with the indentation is more consistent in the end, see cases here:
https://yaml.org/YAML_for_ruby.html#collections

docs/issuing_certificates.md
```diff
@@ -0,0 +1,19 @@
+# Issuing certificates
+
+## Securing OpenShift `Route`s with ACME (Let's Encrypt)
+
+We manage and deploy cluster wide deployment of [OpenShift ACME](https://github.com/tnozicka/openshift-acme) controller. This operator facilitates you Let's Encrypt certificate provisioning in any namespace on the cluster. Any `Route` resource is eligible, please use the following annotation to mark the route as managed by this operator:
```
HumairAK, line 5: > This operator facilitates you Let's Encrypt certificate provisioning  
change to: 
> This operator facilitates Let's Encrypt certificate provisioning  

And maybe add a link for let's encrypt.

docs/issuing_certificates.md
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  annotations:
+    kubernetes.io/tls-acme: "true"
+```
+
+Annotating a route will result with an additional temporary route and 2 pods being created in your namespace. Once the certificate is issued these pods will be removed. This process is repeated to renew certificates later on.
+
+## Issuing certificates via Cert Manager
+
+For other workloads that rely on SSL/TLS certificates will be able to provision certificates via Cert Manager shortly. This is currently work in progress.
```
HumairAK, line 19: > For other workloads that rely on SSL/TLS certificates will be able to provision certificates via Cert Manager shortly. This is currently work in progress.

change to: 

> For other workloads that rely on SSL/TLS certificates, we will be able to provision certificates via Cert Manager shortly. This is currently work in progress.

HumairAK:
few nitpicks on documentation, otherwise lgtm

HumairAK:
> @tumido No, the document start maker, ---. 

I think we  can capture this in linting checks

tumido:
The `---` issue should be decided separatelly and not block this PR https://github.com/operate-first/blueprint/issues/89

Since the topic is open and changes would be needed to be made in the repo no matter which way we decide, I don't think I need to be consistent here. Can we put that matter aside in this PR and don't require consistency here?

docs/issuing_certificates.md
tumido: Updated

docs/issuing_certificates.md
tumido: Updated

gastaldi:
Any chance this can be merged soon? Need this to update the certs for https://registry.quarkus.io/.

Thanks!

HumairAK (approved):
/approve

tumido:
@HumairAK can we get https://github.com/operate-first/argocd-apps/pull/141 in as well?

PR #629, opened 2021-05-11
add missing namespace resources

larsks:


tumido:
I think we were debating this a while back and decided **not** to include those namespace resources in here similarly as we don't include other `openshift-*` namespaces and the rest of OpenShift-owned resources. We have kept these `kustomization.yaml` files in the `cluster-scope/.../namespaces` since we wanted to add the monitoring components in there.

The namespace resource is not created and managed by us - Openshift is adding annotations in there as it pleases and we don't want to interfere with that much.

We're either storing partial manifests here (along with complete namespace manifests) which is neither ideal nor consistent.
Or we're not storing these manifests at all, which is not consistent with the rest of the namespaces, but there's a reason to it.

Any ideas how to solve that?

I think we can get rid of these `kustomization.yaml` files once I scope the Monitoring epic for user workload monitoring - we won't need the components here at all, hence we don't need to track the "namespace partials" at all..

larsks:
Based on our discussion we determined these changes are unnecessary.

PR #633, opened 2021-05-11
fix/scripts for adr0015

larsks:
- update scripts for recent repository reorganization
- Add some testing for scripts


tumido (approved):
/approve
/lgtm

Thank you @larsks :+1: :rocket:

PR #634, opened 2021-05-11
Change location of the namespace manifests

ipolonsk:
Update the path to match the new folder structure in the `enable-monitoring.sh` script.


larsks:
This change is already ins #633.

PR #638, opened 2021-05-12
Add Oauth proxy for Thanos Query Frontend

4n4nd:
Related https://github.com/operate-first/SRE/issues/270

Just to be clear, with this proxy the user will need to be able to "get" the `opf-monitoring` namespace even though this proxy is being deployed in `opf-observatorium` namespace. This is because, thanos has the same data which can be accessed using Prometheus/Grafana.


tumido (commented):
/lgtm

HumairAK (approved):
/lgtm

PR #640, opened 2021-05-12
fix serviceaccount name for thanos oauth proxy

4n4nd:
Related #638 
The current serviceaccount being used does not have enough permissions


HumairAK (approved):
/lgtm

PR #642, opened 2021-05-13
add spark operator and sdap customer resource definition

tloubrieu-jpl:
Hello,

I created this pull request as a follow up on ticket https://github.com/operate-first/support/issues/206

This is the first time I am doing that, let me know how that works for you.

Note I used a script to update your repo, see https://github.com/tloubrieu-jpl/split_yaml_for_moc/


cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/config-operator-role-binding/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: config-operator-role-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
+subjects:
+- kind: ServiceAccount
+  name: config-operator
+  namespace: sdap
```
tumido, line 12: Hm.. I don't think we want to host an cluster scoped operator in a namespace we don't control. This makes the `config-operator` SA in `sdap` namespace of full control over all pods in the cluster. I don't think we should allow that, sorry.

cluster-scope/base/rbac.authorization.k8s.io/clusterroles/RELEASE-NAME-spark-operator/clusterrole.yaml
```diff
@@ -0,0 +1,87 @@
...
+    app.kubernetes.io/version: v1beta2-1.2.3-3.1.1
+    helm.sh/chart: spark-operator-1.1.0
+  name: RELEASE-NAME-spark-operator
+rules:
+- apiGroups:
+  - ''
+  resources:
+  - pods
+  verbs:
+  - '*'
```
tumido, line 17: This is a huge risk

cluster-scope/base/rbac.authorization.k8s.io/clusterroles/RELEASE-NAME-spark-operator/clusterrole.yaml
```diff
@@ -0,0 +1,87 @@
...
+  - watch
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - create
+  - get
+  - update
+  - delete
```
tumido, line 68: This is a huge risk

tumido (changes requested):
Hey @tloubrieu-jpl, thanks for your contribution! I think we have some concerns about security implications of this PR. I don't think we want to have a cluster scoped operator installed in a namespace that we have no control over. We can't grant you all access to many resources across the whole cluster.

Way forward: Can you please remove the Google Spark operator from this PR and separate the SDAP customer resource? For the Google Spark operator there's been some recent works for bringing that to OpenShift happening here https://github.com/guimou/spark-on-openshift  We'll collaborate on that to test it and bring it into Operate First properly as a separate workstream.

tumido:
FYI: I've filed https://github.com/operate-first/support/issues/241 for the Spark Operator - I hope @guimou can help us with that :slightly_smiling_face:

cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/config-operator-role-binding/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: config-operator-role-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
```
tumido, line 8: Do you want us to grant you a `cluster-admin` access here? Can you please scope what your service account needs better? I think cluster admin is a bit to generous here.

cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/config-operator-role-binding/clusterrolebinding.yaml
tloubrieu-jpl: @tumido, I removed the spark updates from the PR.
Regarding your comment here, we don't need actually to have this operator with a cluster scope, how should I change that ?

tumido (changes requested):
Thank you. I still have some permissions problem with this. I would rather see you deploy a cluster role with a given set of permissions than a total grant me access to all kind of role :slightly_smiling_face: 

 If there's any upstream to SDAP, you can just copy the cluster role and point us to the original location.I don't think we

cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/config-operator-role-binding/clusterrolebinding.yaml
tloubrieu-jpl: Hi @tumido , the operator need to read/update/create configmap in a namespace. Should I create a role for that ?

Thanks,

Thomas

tumido: Hey! Ok.. I think we need to take a step back first. What is your intention with this operator? Is this SDAP operator a thing you develop and maintain or is this something you just want to use in your namespace?

If you want to provide this operator as a service to others, you'd need to define a cluster role, sure.

On the other hand, if you want to deploy it just for yourself and keep the operator visibility to your namespace only, you don't need to define any cluster-scoped RBAC. Configmaps are namespaced resources so you can grant this via a role locally within your namespace.

Either way is fine with us, it just needs to be defined before we decide what manifests to use.

I've looked into the sdap sources and they indeed assign `cluster-admin` role, which is quite overkill for any operator in my eyes. We need to narrow the scoped down a bit...

I think you want to go the cluster-scoped operator route (making this operator to work on resources in any namespace on the cluster). If that's the case I think you want a cluster role like this:

```
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: sdap
rules:
  - apiGroups:
      - sdap.apache.org
    resources:
      - gitbasedconfigs
    verbs:
      - create
      - delete
      - deletecollection
      - get
      - list
      - patch
      - update
      - watch
  - apiGroups:
      - ""
    resources:
      - configmaps
    verbs:
      - get
      - list
      - watch
      - create
      - update
      - patch
      - delete
```

I'm unable to find any github repo for nexusjpl/config-operator, can you please point me to the right location? If I can see the actual operator code I can help scope the role better for you. 

If you **don't** want to make this operator available at cluster scope, you can create a namespaced `RoleBinding` resource within your `sdap` namespace and grant the operator any access to your namespace you want - it's your sandboxed playground in the end :slightly_smiling_face: (like even assigning the `admin` `ClusterRole` which grants the operator full project admin access)

tloubrieu-jpl: Thanks @tumido ,
the config-operator is synchronizing git based configuration files  with kubernetes configmap.

Although that might be useful for other we are not planning to support that development for usage outside SDAP. We are barely using it ourselves !

The repo where it is maintained is https://github.com/apache/incubator-sdap-ingester/tree/dev/config_operator

So I think we will go for the option where the operator is only visible to our namespace.

tumido: Ok, in that case we just need to define aggregate cluster roles so you can access the resource as a namespace admin. All the rest can be set from within your own namespace. :slightly_smiling_face: I'll push an update to your PR with all the necessary manifests.

cluster-scope/base/apiextensions.k8s.io/customresourcedefinitions/gitbasedconfigs.sdap.apache.org/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- customresourcedefinition.yaml
```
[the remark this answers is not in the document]
tumido: updated

HumairAK (approved):
/approve

PR #644, opened 2021-05-13
use curl instead of both wget and curl in the same script

larsks:
While editing the update-cluster-resources script, I noticed that
we're using both `curl` and `wget` in the same script. This commits
settles on using `curl` in all cases.


tumido (approved):
/lgtm
/approve

4n4nd (commented):
/lgtm

PR #646, opened 2021-05-13
Add OAuth Proxy for Loki query frontend endpoint

4n4nd:
Related https://github.com/operate-first/SRE/issues/270

From now on users will need "get" access to the `opf-observatorium` namespace to be able to query logs from the Operate First Loki instance. This can be given using the `opf-observatorium-view` role/rolebinding added in this PR.


tumido (commented):
/lgtm

HumairAK (approved):
/lgtm

PR #647, opened 2021-05-13
fix: ACME quota for HA

tumido:
ACME operator requires 2 pods to be running for HA - where CPU limit is 500m on each. `x-small` doesn't allow for the second pod to start successfully.

Note this is just for HA, doesn't have any effect on basic functionality.

Part of: https://github.com/operate-first/support/issues/237


4n4nd (approved):

HumairAK (approved):
/lgtm

PR #654, opened 2021-05-14
Update Jupyterhub StorageClass to moc-nfs-csi

4n4nd:
Tested this in the staging env and it doesn't affect existing PVCs.
New PVCs created were with the new `moc-nfs-csi` storageclass


HumairAK (approved):
/lgtm

PR #668, opened 2021-05-20
Deploy odh google spark operator to zero cluster.

HumairAK:
Related: https://github.com/operate-first/support/issues/241

### DIFF
--- a/kfdefs/base/google-spark-operator/kfdef.yaml
+++ b/kfdefs/base/google-spark-operator/kfdef.yaml
@@ -1,0 +1,20 @@
+---
+apiVersion: kfdef.apps.kubeflow.org/v1
+kind: KfDef
+metadata:
+  annotations:
+    kfctl.kubeflow.io/force-delete: "false"
+  name: opendatahub
+spec:
+  applications:
+    - kustomizeConfig:
+        repoRef:
+          name: guimou
+          path: google-spark-operator/operator
+      name: google-spark-operator
+  repos:
+    - name: manifests
+      uri: "https://github.com/opendatahub-io/odh-manifests/tarball/v1.0.11"
+    - name: guimou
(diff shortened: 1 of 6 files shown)


HumairAK:
currently using @guimou 's repo, but will switch to odh-manifests when https://github.com/opendatahub-io/odh-manifests/pull/427 is merged

HumairAK:
the readme seems to suggest there may be a need for an additional SA / rbac, but the operator has clusterwide privileges for pods so I'm not sure why that would be necessary...

kfdefs/base/google-spark-operator/kfdef.yaml
```diff
@@ -0,0 +1,20 @@
...
+  name: opendatahub
+spec:
+  applications:
+    - kustomizeConfig:
+        repoRef:
+          name: guimou
+          path: google-spark-operator/operator
+      name: google-spark-operator
+  repos:
+    - name: manifests
```
4n4nd, line 16: I don't think we need the "manifests" repo here since it is not being used in the kfdef

HumairAK: the eventual goal is to switch to this repo

tumido: I don;t mid either way. The fact is that keeping it in here makes this manifest eligible for ODH z-stream update PRs  which makes us keep an eye on this manifest more closely. The fact that there's an open PR for this code makes me assume that in 1 or 2 z-stream releases we will be able to switch over to ODH upstream repo here.

tumido (commented):
/approve

tumido (approved):
/lgtm
/approve

4n4nd (approved):
/lgtm

PR #669, opened 2021-05-20
Apply sre-p alertmanager config to openshift-monitoring's prometheus.

HumairAK:
I've updated our alertmanager config to use the one provided by SRE-P, basically everything that goes to pagerduty for them would go to github for us. We are being a bit more selective on what routes go to github rather then sending everything. 

Also looking for more suggestions on what alerts we ought to include/exclude.

### DIFF
--- a/cluster-scope/overlays/moc/common/alertmanager-main-secret.yaml
+++ b/cluster-scope/overlays/moc/common/alertmanager-main-secret.yaml
@@ -9,34 +9,193 @@ stringData:
     global:
       resolve_timeout: 5m
     inhibit_rules:
-      - equal:
-          - namespace
-          - alertname
-        source_match:
-          severity: critical
-        target_match_re:
-          severity: warning|info
-      - equal:
-          - namespace
-          - alertname
-        source_match:
-          severity: warning
-        target_match_re:
-          severity: info
+    - source_match:
+        severity: critical
+      target_match_re:
+        severity: warning|info
+      equal:
+      - namespace
+      - alertname
+    - source_match:
+        severity: warning
+      target_match_re:
+        severity: info
+      equal:
+      - namespace
+      - alertname
+    - source_match:
+        alertname: ClusterOperatorDegraded
+      target_match_re:
+        alertname: ClusterOperatorDown
+      equal:
+      - namespace
+      - name
+    - source_match:
+        alertname: KubeNodeNotReady
+      target_match_re:
+        alertname: KubeNodeUnreachable
+      equal:
+      - node
+      - instance
+    - source_match:
+        alertname: KubeNodeUnreachable
+      target_match_re:
+        alertname: SDNPodNotReady|TargetDown
+    - source_match:
+        alertname: KubeNodeNotReady
+      target_match_re:
+        alertname: KubeDaemonSetRolloutStuck|KubeDaemonSetMisScheduled|KubeDeploymentReplicasMismatch|KubeStatefulSetReplicasMismatch|KubePodNotReady
+      equal:
+      - instance
+    - source_match:
+        alertname: KubeDeploymentReplicasMismatch
+      target_match_re:
+        alertname: KubePodNotReady|KubePodCrashLooping
+      equal:
+      - namespace
+    - source_match:
+        alertname: ElasticsearchOperatorCSVNotSuccessful
+      target_match_re:
+        alertname: ElasticsearchClusterNotHealthy
+      equal:
+      - dummylabel
     receivers:
-      - name: Default
+      - name: Watchdog
+      - name: "null"
+      - name: Github
         webhook_configs:
           - url: 'http://github-receiver-service.opf-alertreceiver.svc.cluster.local:9393/v1/receiver'
-      - name: Watchdog
     route:
+      receiver: "null"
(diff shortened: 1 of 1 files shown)


HumairAK:
related: https://github.com/operate-first/SRE/issues/274

cluster-scope/overlays/moc/common/alertmanager-main-secret.yaml
```diff
@@ -9,34 +9,194 @@ stringData:
...
+            - severity
+          continue: true
+          routes:
+            # Alerts we do not want to receive
+            - receiver: "null"
+              match:
+                alertname: CPUThrottlingHigh
+            - receiver: "null"
+              match:
+                alertname: NodeFilesystemSpaceFillingUp
```
larsks, line 86: If this means what I think it does we probably want to know about it.

cluster-scope/overlays/moc/common/alertmanager-main-secret.yaml
```diff
@@ -9,34 +9,194 @@ stringData:
...
-      repeat_interval: 12h
       routes:
-        - match:
+        - receiver: "null"
+          group_by:
+            - namespace
+            - severity
+          continue: true
+          routes:
+            # Alerts we do not want to receive
```
larsks, line 80: Should the `UpdateAvailable` alert be here?

cluster-scope/overlays/moc/common/alertmanager-main-secret.yaml
HumairAK: I guess they don't care about the "warning" severity, and maybe only "critical", still want to keep it?

cluster-scope/overlays/moc/common/alertmanager-main-secret.yaml
HumairAK: let's add it

HumairAK: added routing of UpdateAvailable to "null"

cluster-scope/overlays/moc/common/alertmanager-main-secret.yaml
larsks: Hard to say without knowing the thresholds. If we're supressing this alert as a warning but passing it as critical, that's probably fine.

larsks:
I think we also need to do something about the `SystemMemoryExceedsReservation`, because those seem constant. I don't know if the solution there is suppressing the notification, or figuring out why these are firing and if there's something we need to configure differently in OpenShift.

larsks:
I think the answer to my previous comment is (b), we should probably [raise the amount of reserved memory](https://docs.openshift.com/container-platform/4.7/nodes/nodes/nodes-nodes-managing.html). I'll open an issue for that, and we should probably keep the alert.

tumido (approved):
/approve

Nice! Thanks for this!

cluster-scope/overlays/moc/common/alertmanager-main-secret.yaml
HumairAK: So these are the alert rules I think: 

```yaml
name: NodeFilesystemSpaceFillingUp
expr: (node_filesystem_avail_bytes{fstype!="",job="node-exporter"} / node_filesystem_size_bytes{fstype!="",job="node-exporter"} * 100 < 40 and predict_linear(node_filesystem_avail_bytes{fstype!="",job="node-exporter"}[6h], 24 * 60 * 60) < 0 and node_filesystem_readonly{fstype!="",job="node-exporter"} == 0)
for: 1h
labels:
  severity: warning
annotations:
  description: Filesystem on {{ $labels.device }} at {{ $labels.instance }} has only {{ printf "%.2f" $value }}% available space left and is filling up.
  summary: Filesystem is predicted to run out of space within the next 24 hours.
```

```yaml
name: NodeFilesystemSpaceFillingUp
expr: (node_filesystem_avail_bytes{fstype!="",job="node-exporter"} / node_filesystem_size_bytes{fstype!="",job="node-exporter"} * 100 < 15 and predict_linear(node_filesystem_avail_bytes{fstype!="",job="node-exporter"}[6h], 4 * 60 * 60) < 0 and node_filesystem_readonly{fstype!="",job="node-exporter"} == 0)
for: 1h
labels:
  severity: critical
annotations:
  description: Filesystem on {{ $labels.device }} at {{ $labels.instance }} has only {{ printf "%.2f" $value }}% available space left and is filling up fast.
  summary: Filesystem is predicted to run out of space within the next 4 hours.
```

The warning says that the expr predicts it the node will run out of space in 24 hours, vs 4 hours for critical -- I feel like I would want to know at the warning stage, wdyt?

larsks: Agreed, let's keep the warning. A 4 hour lead time wouldn't give us much time to respond.

HumairAK: done

HumairAK:
humpty bumpity

tumido (approved):
/lgtm
/approve

PR #673, opened 2021-05-21
Increase system reserved memory for worker nodes

larsks:
We've been seeing a number of errors of the form:

    System memory usage of 1.209G on os-ctrl-1 exceeds 90% of the
    reservation...

See e.g. [1]. According to [2], we can adjust the system reserved
memory by creating a custom Kubeletconfig resource. This commit
adds the necessary resource and applies it to the worker nodes.

According to [3], this may or may not take effect immediately:

> If dynamically updating this field, consider that it may not be
> possible to increase the reserved resources, because this requires
> resizing cgroups. Always look for a NodeAllocatableEnforced event
> after updating this field to ensure that the update was successful.

[1]: https://github.com/operate-first/alerts/issues/7738
[2]: https://docs.openshift.com/container-platform/4.7/nodes/nodes/nodes-nodes-managing.html
[3]: https://github.com/kubernetes/kubelet/blob/master/config/v1beta1/types.go#L744

### DIFF
--- a/cluster-scope/overlays/moc/zero/kubeletconfigs/increase-worker-system-reserved-memory.yaml
+++ b/cluster-scope/overlays/moc/zero/kubeletconfigs/increase-worker-system-reserved-memory.yaml
@@ -1,0 +1,11 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: KubeletConfig
+metadata:
+  name: increase-worker-system-reserved-memory
+spec:
+  machineConfigPoolSelector:
+    matchLabels:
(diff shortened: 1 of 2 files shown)


4n4nd:
@HumairAK will love this

HumairAK:
(there seems to be a linting error you need to fix)

otherwise: 
/lgtm

tumido (approved):
/lgtm

HumairAK:
tested the kustomize build locally since ci seems stuck here, 

/lgtm

PR #675, opened 2021-05-24
switch the targetport with port name for thoth service monitor

harshad16:
switch the targetport with port name for thoth service monitor


4n4nd (approved):
/lgtm

PR #679, opened 2021-05-25
Switch all Subscriptions to automatic approval

larsks:
I believe we had previously determined to switch all our operator
subscriptions to automatic approval. This takes care of the few cases
that were still configured for Manual approval.


4n4nd (commented):
/lgtm

tumido (approved):
/lgtm
/approve

If I remember correctly we needed the Automatic to be set only if the operators are bound to a common namespace like `openshift-operators` or are dependent one on another.

I agree, I think it's safer for now to use the Automatic strategy for all. If we encounter issues we should revert to Manual and finally put the install plan approver to use.

Issue #671, opened 2021-05-20
Increase node reserved memory


larsks:
We frequently see `SystemMemoryExceedsReservation` alerts. We should probably increase the system memory reservation on the workers; it looks like instructions for doing that are [here](https://docs.openshift.com/container-platform/4.7/nodes/nodes/nodes-nodes-managing.html).

(from #669)

4n4nd:
Related https://github.com/operate-first/apps/pull/673

tumido:
Verify if fixed @tumido

tumido:
@larsks @4n4nd we still have this alert here: https://github.com/operate-first/alerts/issues/3544 It's comming from the MOC Infra cluster.

Can we please repeat https://github.com/operate-first/apps/pull/673 against MOC Infra?

PR #685, opened 2021-05-27
overlays/infra: Increase system reserved memory for worker nodes

larsks:
This is the same as #673, but for the moc-infra cluster.

Part of #671.


larsks:
Reminder that merging this will cause a cluster reboot.

4n4nd (commented):
/lgtm

tumido (approved):
/approve
/lgtm

tumido:
@larsks @4n4nd we can unhold whenever ready. I don't think we should see any problems with node restarts on infra so the impact should be minimal - yet we should announce an possible outage.

tumido:
Resolves: https://github.com/operate-first/apps/issues/671

4n4nd (approved):

PR #578, opened 2021-04-27
Add blackbox exporter

hemajv:
This addresses https://github.com/operate-first/SRE/issues/250

Adding the blackbox exporter for our prometheus to allow blackbox probing of endpoints over http/https etc. This will help us to scrape the probe service endpoints making it easier to monitor the availability of our services. The setup implemented is based on the IDH team's [setup](https://github.com/AICoE/internal-data-hub/tree/main/prometheus/bases/blackbox).

PR #686, opened 2021-05-27
fix(blackbox-exporter): Secret generator is not a resource

tumido:
Fixes #578 that broke the `kfdef-zero` ArgoCD app


4n4nd (approved):
/lgtm

4n4nd:
@tumido shouldn't this have been caught by the `kustomize-build` test that prow runs?

tumido:
@4n4nd nope.. the `ksops` resource can render via `kustomize` successfully, it's the API validation on OCP that rejects it during `oc apply`

PR #688, opened 2021-05-27
expose ray dashboard

erikerlandson:
Adds a `Route` for the Ray dashboard, and exposes an environment var `RAY_DASH_URL` that jupyter users can display to get the dashboard URL for their browser.


4n4nd (approved):
/lgtm

PR #690, opened 2021-05-31
Add spark3.0.1 notebook image

oshritf:
Current spark base notebook image is 2.4.5 which is not up to date.
Adds spark3.0.1 base notebook image. Spark version configurable Image created by @anishasthana and Landon (https://github.com/opendatahub-io/odh-manifests/issues/293)
Next up is adding xskipper sample use ipynb to it, and perhaps configure the pyspark jars to include needed jars automatically.

* I wonder if it should be added to https://github.com/aicoe-aiops or to opendatahub-io/odh-manifests instead


HumairAK (approved):
/lgtm

oshritf:
Humair Khan: so it was merged/deployed, but it's not showing up, I believe it's because we're missing a tag name.
https://github.com/operate-first/apps/pull/692

PR #693, opened 2021-06-01
Add Jupyterhub to osc playground.

HumairAK:
Add Jupyterhub for OS-C playground. A separate environment is needed to allow for experimentation with JH itself if needed, and some flexibility.

### DIFF
--- a/kfdefs/overlays/moc/zero/osc-playground/kustomization.yaml
+++ b/kfdefs/overlays/moc/zero/osc-playground/kustomization.yaml
@@ -1,0 +1,23 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: osc-playground
+
+resources:
+  - ../../../../base/jupyterhub
+
+patchesJson6902:
+  - patch: |
+      - op: add
+        path: /spec/applications/2/kustomizeConfig/overlays/-
+        value: storage-class
+      - op: add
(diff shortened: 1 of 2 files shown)


kfdefs/overlays/moc/zero/osc-playground/kustomization.yaml
```diff
@@ -0,0 +1,23 @@
...
+patchesJson6902:
+  - patch: |
+      - op: add
+        path: /spec/applications/2/kustomizeConfig/overlays/-
+        value: storage-class
+      - op: add
+        path: /spec/applications/2/kustomizeConfig/parameters/-
+        value:
+          name: storage_class
+          value: ocs-storagecluster-cephfs
```
tumido, line 18: Just noticed.. maybe you wanna use NFS?

HumairAK: yes I do good catch

HumairAK: I copied it from stage...which should also probably be nfs

HumairAK: updated

4n4nd: we can just remove it and let it use the default storageclass (which currently is the nfs)

4n4nd (approved):
/lgtm

PR #695, opened 2021-06-02
Update osc quota to large.

HumairAK:
Medium is apparently not enough for the full JH deployment we define in the base kfdef. CPU seems to require >2 cores.


4n4nd (approved):
/lgtm

PR #696, opened 2021-06-02
Add custom resourcequota for osc.

HumairAK:
Large quota is not enough, need at least 8 + 2 cores for a large notebook + the jh pods.


4n4nd (approved):
/lgtm

PR #702, opened 2021-06-03
feat: Switch Zero OCP channel to `fast`

tumido:
As a preparation for https://github.com/operate-first/SRE/issues/330 we need to switch from `stable-4.7` to `fast-4.7`.


4n4nd (approved):
/lgtm

PR #705, opened 2021-06-03
chore: Update cluster version updater script when multiple updates are available

tumido:
Just a quick update to the cluster version updater script, before it gets eventually converted to the golang version.

Offer a choice if there are multiple updates available:
Peek 2021-06-03 16-10


HumairAK (approved):
/lgtm

Issue #641, opened 2021-05-12
Opf-prometheus metrics are not being scraped since the proxy was added


4n4nd:
Relevant Alert https://github.com/operate-first/alerts/issues/5890

This is resulting in an alert being fired for "Prometheus Service Down" (https://prometheus-portal-opf-monitoring.apps.zero.massopen.cloud/alerts)

PR #714, opened 2021-06-07
Update odh servicemonitors

4n4nd:
The [upstream directory](https://github.com/opendatahub-io/odh-manifests/tree/master/prometheus/operator/base/service-monitors) for servicemonitor manifests has changed, they need to be changed here as well.
This has resulted into multiple servicemonitors looking for same services.

Resolves https://github.com/operate-first/apps/issues/641


HumairAK (approved):
/lgtm

PR #715, opened 2021-06-07
fix: change scheme for prometheus serviemonitor

4n4nd:
After the oauth proxy was added, the service was changed to https from http, this needs to be reflected in the servicemonitor as well.
Resolves https://github.com/operate-first/apps/issues/641


tumido (approved):
/lgtm
/approve

PR #719, opened 2021-06-08
feat: Install Koku operator to Zero cluster

tumido:
Created via https://github.com/operate-first/opfcli/pull/22
Part of: https://github.com/operate-first/support/issues/265

Install and enables operator on Curator cluster

/cc @skanthed

### DIFF
--- a/cluster-scope/base/core/namespaces/koku-metrics-operator/kustomization.yaml
+++ b/cluster-scope/base/core/namespaces/koku-metrics-operator/kustomization.yaml
@@ -1,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+    - namespace.yaml
+components:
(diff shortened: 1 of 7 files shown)


tumido:
/unhold 

this goes to the Zero cluster instead of the Curator cluster. A consecutive PR adding a new curator team to the Zero cluster + granting them access to koku namespace will follow.

HumairAK:
aside: we don't typically want quotas on operator namespaces yes? 

/lgtm

4n4nd (commented):
/lgtm

tumido:
> aside: we don't typically want quotas on operator namespaces yes?

@HumairAK yeah.. I don't know the resource consumption of the operator yet.. but it seems this is a namespace scoped operator and the curator team wants to have access to it.. so.. we may want to set some quota to prevent starving... We'll need to monitor the workloads there and add quota based on the usage. :shrug:

PR #721, opened 2021-06-08
Prevent usage of rgw storage class for buckets.

HumairAK:
~Resolve: https://github.com/operate-first/SRE/issues/324~

I also added a cap of 1 objectbucketclaim for each quota. I also updated existing user custom quotas, a brief skim through indicated they already respected these restrictions so it should not affect them. 

Reference: https://kubernetes.io/docs/concepts/policy/resource-quotas/#storage-resource-quota

### DIFF
--- a/cluster-scope/base/core/namespaces/opf-ci-pipelines/resourcequota.yaml
+++ b/cluster-scope/base/core/namespaces/opf-ci-pipelines/resourcequota.yaml
@@ -10,3 +10,4 @@ spec:
     requests.memory: 80Gi
     requests.storage: 30Gi
     count/taskruns.tekton.dev: 3k
+    count/objectbucketclaims.objectbucket.io: 1
--- a/cluster-scope/base/core/namespaces/opf-ci-prow/resourcequota.yaml
+++ b/cluster-scope/base/core/namespaces/opf-ci-prow/resourcequota.yaml
@@ -9,3 +9,4 @@ spec:
     requests.cpu: '16'
     requests.memory: 36Gi
     requests.storage: 30Gi
+    count/objectbucketclaims.objectbucket.io: 1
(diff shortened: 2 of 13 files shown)


larsks:
> I also added a cap of 1 objectbucketclaim for each quota.

Is it common for applications to only interact with a single bucket?

HumairAK:
@larsks I'm not sure, I can up the numbers for the higher tiers if there's a suggestion. I figured allowing unlimited buckets was something we didn't want to do.

HumairAK:
As @tumido pointed out on slack, `requests.storage` would only affect pvcs, we'll have to think of a different way to prevent ceph rgw buckets from being created.

tumido:
linking it in here as well, there's currently no way to quota storage used by buckets https://github.com/rook/rook/issues/7906

we can still limit the OBC counts though, since that's a kube native mechanism

HumairAK:
@tumido any thoughts on whether we should cap at 1 bucket, or should this number be higher, and dependent on tier size?

4n4nd:
I think 1 is a good start for all the OBC quotas. We haven't seen any project use more than 1 OBC yet?
If someone requests for it, we can adjust the quotas later.

4n4nd (commented):
/lgtm

HumairAK:
I've updated the pr to include only a quota on bucket creation count

tumido (approved):
/lgtm
/approve

I think `1` is a nice starting point. We can increase it any time :slightly_smiling_face:

tumido:
@HumairAK is there a reason for the hold? Or was that just because of the review time window? :slightly_smiling_face:

HumairAK:
Don't remember -- let's merge it. 

/unhold

larsks:
@HumairAK  is this effectively the same as https://github.com/operate-first/SRE/issues/313?

tumido:
@larsks I don't think so. This limits the bucket quantity to 1, but it's still possible to provision that single bucket using the "forbidden" storage class.

Issue #574, opened 2021-04-26
Apply resource quotas to user namespaces that don't yet have one


HumairAK:
As title suggests, we should hunt down/reach out to teams that have not yet implemented a resource quota and apply it.

tumido:
```
$ grep -rL "limitranges" cluster-scope/base/namespaces/*/kustomization.yaml | awk -F/ '{print $4}'

acm
apicurio-apicurio-registry
argocd
as-pushgateway
b4mad-minecraft
cnv-testing
ds-black-flake
ds-example-project
ds-ml-workflows-ws
fde-audio-decoder-demo
hostpath-provisioner
kubeflow
lab-cicd-1-jump-app-cicd
lab-cicd-1-jump-app-dev
lab-cicd-2-jump-app-cicd
lab-cicd-2-jump-app-dev
lab-cicd-3-jump-app-cicd
lab-cicd-3-jump-app-dev
lab-cicd-4-jump-app-cicd
lab-cicd-4-jump-app-dev
local-storage
m4d-blueprints
m4d-system
mesh-for-data
metallb-system
observatorium-operator
open-aiops
opendatahub-operator
openshift-cnv
openshift-logging
openshift-metering
openshift-monitoring
openshift-operators
openshift-operators-redhat
openshift-serverless
openshift-storage
opf-ci-pipelines
opf-ci-prow
opf-jupyterhub
opf-jupyterhub-stage
opf-kafka
opf-monitoring
opf-observatorium
pulp-operator
ray-odh-integration
sa-dach-anwendertreffen
sa-dach-openshift-examples
sdap
tekton-pipelines
tufts-dcc-6
uky-hpc-workload-generator
ws-ml-prague
```

- `opf-ci-pipelines` and `opf-ci-prow` are the most pressing ones - those cause troubles all the time, I'll make a PR for those.
- not sure if we need or should be setting it for `openshift-*` namespaces and other operator-related ones like `local-storage` etc.

HumairAK:
> opf-ci-pipelines and opf-ci-prow are the most pressing ones - those cause troubles all the time, I'll make a PR for those 

@harshad16 ^

harshad16:
> * `opf-ci-pipelines` and `opf-ci-prow` are the most pressing ones - those cause troubles all the time.

@tumido , what kind of trouble is being faced, can you please elaborate. 
As a maintainer of these I would work on them, as you stated these **troubles are faced all the time**.

 
[...]

PR #726, opened 2021-06-09
Apply resource quotas to namespaces that are without one.

HumairAK:
Applied quotas to namespaces as specified: https://github.com/operate-first/apps/issues/574#issuecomment-827497979

I didn't touch `openshift-*` or `opf-*` namespaces. 

I also tried to peek in to some of the namespaces to see if they are active/maintained, and tried to approximate a quota that's better fitting then just `x-small`, if it was either unclear, or a namespace not in use, then I defaulted to `x-small`.


4n4nd (approved):
/lgtm

PR #727, opened 2021-06-09
Update thanos object storage creds

4n4nd:
This will update the thanos Object Storage credentials.
The new storage was created by @larsks and the previous data has been copied to this bucket manually.


HumairAK (approved):
/approve
/lgtm

PR #722, opened 2021-06-09
fix(alerts): ClusterOperatorDown/Degraded inhibit rule dedup

tumido:
https://github.com/operate-first/apps/pull/718 introduced a duplicity in attempt to correct SRE-P rule - this made both `ClusterOperatorDown` and `ClusterOperatorDegraded` alerts suppressed. And we don't want that. I hope this will correct things.

PR #728, opened 2021-06-10
fix: Semi-revert #722

tumido:
Partially reverts #722 

Looking back at the source

https://github.com/openshift/configure-alertmanager-operator/blob/c9032fd13a9dcd7d374d1491db4b23af1120ba11/pkg/controller/secret/secret_controller.go#L349-L350

It seems there's a reason why `ClusterOperatorDegraded` is favored over `ClusterOperatorDown`

/cc @HumairAK


4n4nd (approved):
/lgtm

PR #729, opened 2021-06-10
Create interface on the MOC production openshift provisioning network.

naved001:
Plan is to create the MOC cluster using ACM.


HumairAK:
@larsks

larsks:
On hold while we discuss the naming of things...

naved001:
> New changes are detected. LGTM label has been removed.

good bot.

PR #733, opened 2021-06-10
Increase Apicurio Registry quota

paoloantinori:
Just noticed the new Quota requirements.
We are currently still in active exploration on our namespace so our needs might vary in the future (with a possibility to split namespaces in more specialized ones).

This is a screenshot of today's status. I've picked the Medium tier but please advise if you feel we could calibrate it more accurately.

Screenshot from 2021-06-10 22-30-10


HumairAK (approved):
/lgtm

paoloantinori:
hi @HumairAK , no worries, but I'm afraid I need to bump it again to get back to be operational. We are short on CPU. I will start assigning more restrictive quotas, but I was hoping to do it only after I manage to bring online most of the pods.

PR #736, opened 2021-06-11
Add thoth-station SA in `opf-monitoring`

4n4nd:
give thoth-station access to opf-monitoring

Resolves https://github.com/operate-first/support/issues/267


HumairAK (approved):
/lgtm
/approve

harshad16:
@4n4nd  thank you for the PR
i was wondering based on your comments in the slack thread
would it be more convenient if we create a SA in our namespace and then we provide rolebinding ?

harshad16:
actaully it feel like better way.
/lgtm

PR #738, opened 2021-06-11
Add odh-admins to odh namespaces.

HumairAK:
Adding @rimolive to odh-admins, and also extending odh-admins permissions to other odh namespaces they don't currently have access to.


4n4nd (approved):
/lgtm

PR #741, opened 2021-06-11
Create argocd-readonly group for read-only access to all resources

larsks:
Add naved (who wants to be able to poke around) and my @redhat.com
account (for testing this out).

### DIFF
--- a/argocd/overlays/moc-infra/configs/argo_cm/dex.config
+++ b/argocd/overlays/moc-infra/configs/argo_cm/dex.config
@@ -11,6 +11,7 @@ connectors:
       groups:
         - apicurio
         - argocd-admins
+        - argocd-readonly
         - b4mad
         - thoth
         - operate-first
--- a/argocd/overlays/moc-infra/configs/argo_rbac_cm/policy.csv
+++ b/argocd/overlays/moc-infra/configs/argo_rbac_cm/policy.csv
@@ -12,6 +12,10 @@ p, role:readonly-public, applications, get, workshops/*, allow
 # Give Openshift group (argocd-admins) the argocd admin role with unrestricted argocd access
 g, argocd-admins, role:admin
 
+# Give Openshift group (argocd-readonly) the argocd admin readonly for read access
+# to all resources.
+g, argocd-readonly, role:readonly
+
 # Give read access to argocd resources for standard users
 g, thoth, role:standard-user
 g, operate-first, role:standard-user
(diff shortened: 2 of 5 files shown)


HumairAK:
@larsks can you add this ocp group here: https://github.com/operate-first/apps/blob/master/argocd/overlays/moc-infra/configs/argo_cm/dex.config#L11

this is so people in this group can authenticate via dex.

larsks:
@HumairAK added!

HumairAK:
Just FYI I've encountered some issues in the past with newlines in these config files, though I think that was only for the ENV files, so I think this should be okay. But just a heads up in case things go awry.

Also -- this will require a manual `kustomize build...  | oc apply ... ` etc. to bring changes to the infra cluster. 

/lgtm 
/approve

larsks:
@HumairAK

> Just FYI I've encountered some issues in the past with newlines in these config files

Could you expand on this a bit? Which files in particular, and is there a place where I added newlines in a way that was inconsistent with how the files were already formatted?

> Also -- this will require a manual kustomize build... | oc apply ... etc. to bring changes to the infra cluster.

Why does this require the manual build/apply step? Does argocd not manage itself?

HumairAK:
@larsks currently all of the files in this location: https://github.com/operate-first/apps/tree/master/argocd/overlays/moc-infra/configs 
don't have a newline. These get consumed by the [configmapgenerator](https://github.com/operate-first/apps/blob/master/argocd/overlays/moc-infra/kustomization.yaml#L9).

> Why does this require the manual build/apply step? Does argocd not manage itself? 

Currently no, we can add it, but I'd prefer if we would keep auto sync off (otherwise we would need to turn off auto sync for 3 apps when trying to turn off autosync for any one particular app).

HumairAK:
For more context: Sometime back we had added 1 file per each env var in the configmapgenerator, (like how [these folks](https://github.com/AICoE/aicoe-cd/tree/main/manifests/overlays/prod/configs/argo_cm) do it). and if I recall correctly, adding a new line would get included in the env var, so I had CI ignore newlines in this location and removed them from all these configs to avoid such issues. 

But since then I've moved the env vars in to a while [here](https://github.com/operate-first/apps/blob/master/argocd/overlays/moc-infra/configs/argo_cm/envs), and I don't think these get affected by any newlines in the file.

larsks:
> currently all of the files in this location: https://github.com/operate-first/apps/tree/master/argocd/overlays/moc-infra/configs don't have a newline.

@HumairAK, Now I'm extra confused, because files in that `configs` directory have *lots* of newlines, so I'm not sure what we're talking about.

HumairAK:
bah sorry, I see that I never specified this earlier -- I'm referring to newlines at the end of the file. Woops sorry!

PR #751, opened 2021-06-14
chore: Downsize ACME operator

tumido:
I've figured we don't need a HA for this, it should be fine to run a single replica


HumairAK (approved):
/lgtm
/approve

PR #753, opened 2021-06-14
fix: Use common pull secret in MOC

tumido:
Resolves https://github.com/operate-first/SRE/issues/339

Do we know about any better way how to keep ACM in line?  

ACM synces pull secrets from those defined in cluster namespaces - if we introduce one single pull secret in the `common` we would just make ArgoCD and ACM fight. I'll raise issue upstream


larsks:
@tumido we should probably use a pull secret tied to the MOC subscription.

tumido:
@larsks I've used the one from Zero cluster.. is there a better one?

larsks:
@tumido It looks like on the zero cluster we used my redhat account, which isn't tied to the BU subscriptions. We should probably switch to using one on the BU account. I have a separate redhat uid that's tied to BU; let me grab the pull secret for that account.

larsks:
@tumido I see there are *three* pull secrets modified in this PR. Do the curator- and zero- pull secrets get pushed to their respective clusters?

larsks:
I've updated this PR so they all use my BU credentials.

tumido:
@larsks ACM should push those curator and zero secrets over. I think we want to avoid ArgoCD and ACM fighting over it and ACM can't be forced to not sync those resources.. hence I'm leaving it on ACM to sync it

tumido:
I've noticed you've overwritten wrong namespace (now all goes to `zero` namespace. that's not right). Wait for me to fix it.

4n4nd (commented):
/lgtm

tumido:
bump, I need somebody to approve this @HumairAK @4n4nd. lgtm is not enough.

4n4nd (approved):

PR #740, opened 2021-06-11
Add events creator role to sdap.

HumairAK:
Should resolve: https://github.com/operate-first/support/issues/266

PR #754, opened 2021-06-14
Aggregate events to edit

tumido:
Resolves: https://github.com/operate-first/support/issues/271
Reverts: #740

Upstream fix pending.

Adds write permissions only, since read is aggregated from view already.


tumido:
cc @4n4nd @larsks @HumairAK

HumairAK:
> Adds write permissions only, since read is aggregated from view already.
Wouldn't you want edit to also have read perms so it can work as a standalone role, and not require view?

Also -- why not aggregate to admin (since that's the cluster role we bind to during onboarding)?

cluster-scope/overlays/moc/common/kustomization.yaml
```diff
@@ -38,6 +38,7 @@ resources:
   - ../../../base/user.openshift.io/groups/tufts-dcc-6
   - ../../../base/user.openshift.io/groups/uky-redhat
   - ../../../base/user.openshift.io/groups/workshops
+  - ../../../base/rbac.authorization.k8s.io/clusterroles/events-aggregate-to-edit
```
HumairAK, line 41: mind placing it alphabetically?

tumido:
> Wouldn't you want edit to also have read perms so it can work as a standalone role, and not require view?
>
> Also -- why not aggregate to admin (since that's the cluster role we bind to during onboarding)?

@HumairAK 

That's not how it works. All permissions aggregated to view are aggregated to edit and admin as well. Analogically all permission aggregated to edit are also aggregated/propagated to admin. So if view role has read on events, so does edit and admin. And this PR adds write for events to edit which means it's added to admin as well. See my upstream PR that did the same and got merged into Kubernetes: https://github.com/kubernetes/kubernetes/pull/34729

cluster-scope/overlays/moc/common/kustomization.yaml
tumido: done

HumairAK:
>  this PR adds write for events to edit which means it's added to admin as well.  

Gotcha, then

/lgtm
/approve

PR #756, opened 2021-06-15
In onboarding.sh change display-name annotation to description

4n4nd:
Related https://github.com/operate-first/opfcli/issues/41

### DIFF
--- a/scripts/onboarding.sh
+++ b/scripts/onboarding.sh
@@ -20,7 +20,7 @@ create_namespace() {
 	kind: Namespace
 	metadata:
 	  annotations:
-	    openshift.io/display-name: "$DESCRIPTION"
+	    openshift.io/description: "$DESCRIPTION"
 	    openshift.io/requester: $OWNER
 	  name: $NAMESPACE
 	spec: {}


larsks:
Rather than fixing onboarding.sh we should just remove it.

4n4nd:
unfortunately [some of our docs](https://github.com/operate-first/support/blob/main/docs/onboarding_to_cluster.md#adding-namespaces) still need to be updated. I think until we update the docs to use opfcli, we should keep the script.

larsks:
Then maybe this PR should update the docs *and* remove the script. I am highly in favor of using this as an excuse to get rid of the script, rather than continuing to maintain two separate implementations of this functionality.

4n4nd:
afaik @HumairAK was working on updating the docs, but he isn't around today. I thought until then we could add this minor fix.

tumido:
Let's resolve the discussion at https://github.com/operate-first/opfcli/issues/41 first.

4n4nd:
it was decided to not maintain the `onboarding.sh` script. closing

PR #768, opened 2021-06-17
update retention to match cache time for advises and provenance

KPostOffice:
If messages are still around after cache expires they will no longer provide info to users. Cache retention time is 4hours, I adjusted topic retention time to match this.


4n4nd (approved):
/lgtm

harshad16 (commented):
/lgtm
thanks :100:

PR #774, opened 2021-06-21
Revert "Upgrade ODH to v1.1.0"

4n4nd:
Reverts operate-first/apps#773


tumido:
FTR: Revert was decided because there's no ODH operator released yet. We've decided to wait for couple of minutes to see if upstream can resolve the issue. If not we'll proceed to revert.

4n4nd:
The update was completed successfully so we won't need this PR anymore. Closing.

PR #778, opened 2021-06-21
update ray images to use ray 1.4 and include ibm codeflare

erikerlandson:
Recent versions of ray 1.x (e.g. the latest, 1.4) support the ray client-server connect, and so I am updating all the images to use ray 1.4.0 instead of nightly wheel builds.  I have also included IBM's codeflare, which they released to pypi this week.


4n4nd (approved):
/lgtm

PR #780, opened 2021-06-22
feat: Convert routes with cert to acme-operator

tumido:
Resolves: https://github.com/operate-first/support/issues/280


tumido:
let's hope ArgoCD or odh operator won't be deleting the certs, since they are not part of the spec... :confused:

4n4nd (commented):
/lgtm

tumido:
/unhold

only dashboard should propagate automatically :slightly_smiling_face:

PR #770, opened 2021-06-18
Create new KafkaUser for cluster logging

4n4nd:
Add kafka creds to ClusterLogForwarder
Add Kafka tls creds to vector

PR #785, opened 2021-06-22
Fix Kafka bootstrap url in Vector configmap

4n4nd:
We need to use the public kafka url for tls authentication
Related https://github.com/operate-first/apps/pull/770


tumido (approved):
/lgtm
/approve

PR #788, opened 2021-06-23
feat: Add new on-call members

tumido:
Part of: https://github.com/operate-first/SRE/issues/345
Requires: https://github.com/tumido/slack-first/issues/115

PR #795, opened 2021-06-23
fix: Reorder on-call to keep current schedule

tumido:
Put Humair all the way to the back (from the first place) to restore the rotation schedule (with the previous change #788 ) we've changed the current on call person from me back to Anand.


4n4nd (approved):
/lgtm

HumairAK (approved):
/lgtm
/approve

PR #816, opened 2021-06-29
Update the CPU limit graph

Shreyanand:
In the odh Jupyterhub dashboard, the query for CPU usage percentage should have ```usage / limit``` instead of ```usage / resource```  to be  consistent with the title of the graph and the graph for memory usage percentage.


tumido (approved):
/lgmt
/approve

good catch :+1:

Issue #820, opened 2021-07-01
No Prom metrics for Vector deployment in opf-observatorium


4n4nd:
Docs on how to expose these metrics are available [here](https://vector.dev/docs/reference/configuration/sources/internal_metrics/)

PR #821, opened 2021-07-01
Expose Prometheus metrics for Vector

4n4nd:
Update vector to v0.14.x
Use internal route for kafka sync

Resolves https://github.com/operate-first/apps/issues/820

### DIFF
--- a/vector/overlays/moc/zero/opf-observatorium/service.yaml
+++ b/vector/overlays/moc/zero/opf-observatorium/service.yaml
@@ -1,0 +1,17 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: vector-metrics
+spec:
+  selector:
+    app: vector
+  ports:
+    - name: metrics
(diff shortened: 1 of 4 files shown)


vector/overlays/moc/zero/opf-observatorium/service.yaml
```diff
@@ -0,0 +1,17 @@
...
+  selector:
+    app: vector
+  ports:
+    - name: metrics
+      protocol: TCP
+      port: 80
+      targetPort: metrics
+    - name: graphql-api
+      protocol: TCP
+      port: 8000
```
tumido, line 16: Is there a reason to remap the port numbers? Why not keep it consistent with the actually exposed ports?

4n4nd: updated :+1:

HumairAK (approved):
/lgtm
/approve

PR #833, opened 2021-07-12
Update resource quota for trino namespace.

HumairAK:
To make room for cloudbeaver as per https://github.com/operate-first/support/issues/299


4n4nd (approved):
/lgtm

PR #835, opened 2021-07-13
feat: Onboard debezium-ui

uidoyen:
Resolves: https://github.com/operate-first/support/issues/304


HumairAK:
Thanks for the PR @uidoyen ! We always appreciate it when folks submit their PR's themselves :) 

As per the @harshad16 's comment in the issue, we're waiting on more hardware before onboarding new projects, but we'll keep this open and bring it in once that happens. 

/hold

tumido:
@HumairAK rebased :slightly_smiling_face:

PR #836, opened 2021-07-13
Add cluster-monitoring-config 

4n4nd:
This will attach PVCs to openshift-monitoring prometheus instances.
There are 2 replicas of openshift-monitoring prometheus so this will create 2 PVCs of size 100Gi each.


HumairAK (approved):
/lgtm
/approve

larsks:
@4n4nd the existing instances are already using `>` 200GB for each instance. Have you also reduced the retention time?

PR #843, opened 2021-07-14
added group authorization for kafka user

Gkrumbach07:
I am using KafkaJS which requires kafka consumers to have a consumer group (compared to something like kafka-python). Therefore I am getting an error: `Not authorized to access group: Group authorization failed` when i try to connect a consumer.

So this pull request is adding that group to my user yaml.


4n4nd (approved):
/lgtm

PR #845, opened 2021-07-16
Add superset image and update trino config to match live.

HumairAK:
This should match live state and thus allow us to renable auto sync on odh. 

The config is mostly the same as upstream save for this new addition: 

```
<property>
<name>fs.s3a.path.style.access</name>
<value>true</value>
</property>
```


4n4nd (approved):
/lgtm

Issue #782, opened 2021-06-22
Probe "opf-prober" is not being applied properly causing it to be not in sync in argocd


HumairAK:
See [here](https://argocd.operate-first.cloud/applications/kfdefs-zero?node=monitoring.coreos.com%2FProbe%2Fopf-monitoring%2Fopf-prober%2F0). If we sync it, nothing happens. The offending field is `relabelingConfigs`. Let's ensure the versions / specs match up, and that this field indeed exists for us.

HumairAK:
removed the relabeling configs for now to avoid constant sync loops in argocd, for the record these were removed: 

```yaml
      relabelingConfigs:
      - source_labels: [__address__]
        target_label: __param_target
      - source_labels: [__param_target]
        target_label: instance
      - target_label: __address__
        replacement: blackbox-prober.opf-monitoring.svc.cluster.local:80
      - regex: 'namespace'
        action: labeldrop
```

PR #846, opened 2021-07-16
Remove relabeling commits to match live.

HumairAK:
Not sure why these keep getting removed, the version seems to support it, but let's take it out for now so we can put kfdefs app on auto sync, will delegate resolution of this [to a separate issue](https://github.com/operate-first/apps/issues/782).


tumido (approved):
/lgtm
/approve

PR #848, opened 2021-07-21
Request to add a image to support BU Jupiter-Book based textbook, lecture notes, lab manual and assignments

jappavoo:
Hi All,

Thanks so much for helping with this.


jappavoo:
Opening a new pull request that squashes the unnecessary commits

PR #851, opened 2021-07-22
feat: diverge localstorage for Zero and Rick

tumido:
Part of: https://github.com/operate-first/hetzner-baremetal-openshift/issues/12

It turned out that Rick has local storage also deployed, but in different namespace than zero and uses `LocalVolumeSet` instead of `LocalVolume` used in Zero. This PR is trying to get those resources tracked in git. Further unification will be needed.


4n4nd (commented):
/lgtm

HumairAK (approved):
/lgtm

PR #852, opened 2021-07-22
feat: diverge OCS for Rick and Zero

tumido:
Part of: https://github.com/operate-first/hetzner-baremetal-openshift/issues/12

Rick cluster uses a bit different OCS spec (it also uses a different OCS version). Let's track those and converge some of the fields in the future.


4n4nd (approved):
/lgtm

PR #856, opened 2021-07-22
Update grafana dashbaord for cpu usage in zero cluster.

HumairAK:
The change adds load averages for all nodes on the zero cluster. Need to use thanos datasource for accurate results until we install querier in the `opf-monitoring` namespace.


4n4nd (approved):
/lgtm

PR #862, opened 2021-07-26
Revert "feat: Add Augur as external service to JH"

4n4nd:
Reverts operate-first/apps#860

Some context https://github.com/operate-first/support/issues/306#issuecomment-886978861


tumido (commented):
let's hold this and keep the external service available until we have confirmation it was indeed a MOC network issue and the service can be accessed directly. If the issue gets resolved and the service still can't be accessed directly, it's worth keeping this external service available. On the other hand if it can be accessed directly, there's no need for external service here. :slightly_smiling_face: 

Let's just wait for resolution

/hold

4n4nd:
@tumido this issue seems to have been fixed [here](https://github.com/operate-first/support/issues/309#issuecomment-893826276), so we can continue merging this PR.

tumido (approved):
/unhold
/lgtm
/approve

PR #866, opened 2021-07-27
Increase cluster-monitoring storage size to 150Gi

4n4nd:
Openshift Monitoring Prometheus Instance PVCs are full, we need to increase their size.


HumairAK (approved):
/lgtm
/approve

PR #867, opened 2021-07-27
Increase keycloak resource allotment on infra.

HumairAK:
Expecting to fix the following term signal that's being received: 

```
[0m[0;31mERROR *** JBossAS wrapper process (1) received TERM signal ***[0m
```


4n4nd (approved):
/lgtm

PR #868, opened 2021-07-28
feat: Added group sre with role cluster-reader to all clusters

mimotej:
Created group sre and added it to all cluster and granted it role cluster-reader (implementation of this issue https://github.com/operate-first/support/issues/318).


HumairAK:
Thanks @mimotej this looks good. I would recommend that next time you work off another git `branch` instead of `master` branch, this will help you keep things a bit more organized in your workflow. The pr itself lgtm :) 

/lgtm
/aprove

PR #871, opened 2021-07-29
fix: add sre to secret generator

mimotej:
Added fix because sre group was missing in secret-generator file. (issue https://github.com/operate-first/support/issues/318)


HumairAK (approved):
/lgtm
/approve

PR #872, opened 2021-07-29
add enterprise-neurosystem group and namespace

erikerlandson:
addresses https://github.com/operate-first/support/issues/313

### DIFF
--- a/cluster-scope/base/core/namespaces/enterprise-neurosystem-collaboration/kustomization.yaml
+++ b/cluster-scope/base/core/namespaces/enterprise-neurosystem-collaboration/kustomization.yaml
@@ -1,0 +1,9 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+    - namespace.yaml
+components:
+    - ../../../../components/project-admin-rolebindings/enterprise-neurosystem
+    - ../../../../components/limitranges/default
+    - ../../../../components/resourcequotas/medium
+namespace: enterprise-neurosystem-collaboration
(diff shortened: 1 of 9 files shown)


cluster-scope/base/core/namespaces/enterprise-neurosystem-collaboration/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+    - namespace.yaml
+components:
+    - ../../../../components/project-admin-rolebindings/enterprise-neurosystem
```
4n4nd, line 6: @erikerlandson in your request you requested for a `Medium` namespace quota, can you add the quota in this kustomization? ([Example](https://github.com/operate-first/apps/blob/307af64697d70e9aa85a5806692abda63a505bb6/cluster-scope/base/core/namespaces/ds-example-project/kustomization.yaml#L6))
If you are using the opf cli, you can use [the `-q` flag](https://github.com/operate-first/opfcli#create-project)

4n4nd (changes requested):

cluster-scope/base/core/namespaces/enterprise-neurosystem-collaboration/kustomization.yaml
erikerlandson: can I just use `-q medium` ?

4n4nd (approved):

PR #877, opened 2021-08-04
fix/failing nncp

larsks:
Apparently there are accuracy issues in [1], so fix the case of
Parallel in the nncp resources.

[1]: https://access.redhat.com/solutions/6096871


HumairAK (approved):
/lgtm
/approve

PR #891, opened 2021-08-12
Add auth via dex to grafana on zero.

HumairAK:
Resolve: #37 

Also does a few other things: 

- Remove hue client from dex 
- Fix dex redirect/issuer paths
- Fix dex SA namesspace
- Change Grafana route to match route metadata.name
- Add doc on how to give ocp groups grafana roles 

Those in the ocp groups `cluster-admins` / `operate-first` have Admin in Grafana, while those in `data-science` group have `viewer`

### DIFF
--- a/dex/base/dex-sa.yaml
+++ b/dex/base/dex-sa.yaml
@@ -4,5 +4,6 @@ kind: ServiceAccount
 metadata:
   name: dex
   annotations:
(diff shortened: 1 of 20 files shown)


4n4nd (approved):
/lgtm

4n4nd:
We can create separate groups in each grafana namespace which can be used as dedicated groups for grafana access. This can be done in a separate PR.
Also, thanks for doing this!

PR #894, opened 2021-08-16
Update superset oauth configs and rbac access.

HumairAK:
Should also resolve: https://github.com/operate-first/support/issues/354

Added a new `superset-user` group, fill follow up with docs on how to get onboarded to superset for access. I just gave them 'Gamma' level access, that seemed sufficient. Not sure if we need a more robust mapping set up, but I figured we'd do that if/when the need arises. 

Reference: 
https://superset.apache.org/docs/security#provided-roles
https://superset.apache.org/docs/installation/configuring-superset#custom-oauth2-configuration


HumairAK:
Shameless self merge to fix auth for these services, as no one is around to approve this atm.

Issue #791, opened 2021-06-23
Migrate users to Keycloak


tumido:
Migrate users from MOC SSO to Keycloak by changing and decrypting the OCP `Group` resources in common overlay

- [x] Provide dual identity by enlisting github usernames as well as emails
- [x] Once MOC SSO is disabled, decrypt groups and store github usernames only (in `cluster-scope/base`)

tumido:
Resolved via https://github.com/operate-first/apps/pull/842

PR #905, opened 2021-08-18
feat: Enable Operate First SSO on Curator

tumido:
Part of: #791 

We forgot about the curator cluster this time... let's enable it there as well.


4n4nd (approved):
/lgtm

PR #911, opened 2021-08-19
feat: Set better meteor limit range

tumido:
Tekton pipelines uses default limitranges for default settings. This change will make tekton steps use better and larger limits by default.

/cc @4n4nd @HumairAK


4n4nd (approved):
/lgtm

PR #912, opened 2021-08-19
feat: Up meteor quota limit

tumido:
I think I need to request bigger quota limit for the meteor namespace.

Together with #911 the default cumulative limit went up, because each container has 4 CPU limit by default. So if a pod has 3 containers it's limit is 6 cores now. It will probably never use it but we need to be able to schedule it. This should keep requests low and raise limits for now. I'll look into better way to influnce those limits in future.


4n4nd (approved):
/lgtm

4n4nd (approved):
/lgtm

PR #913, opened 2021-08-19
Schedule trino workers to os-wrk-1 on zero.

HumairAK:
Using overrides here to change resource usages (cpu:8, mem: 16Gi), and adds the toleration/node selectors to run on os-wrk-1


4n4nd (approved):
/lgtm

PR #918, opened 2021-08-23
Add OSC manifests to opf apps repo.

HumairAK:
The PR introduces manifests for os-climate cluster `osc-cl1`, no changes are made to `zero/infra/rick`. See commits for details on what's added. (No actual changes to any live deployment will be made, argocd apps will follow this pr for the respective services being added.) 

In the OSC cluster I've used `odh-*` prefixes for the odh namespaces, instead of `opf-*` hence the addition of those namespaces.


4n4nd (approved):
/lgtm

PR #922, opened 2021-08-25
add external label to moc/zero prometheus

4n4nd:
This will add a label `prometheus: moc/zero` to all the metrics exported using remotewrite


tumido (approved):
/lgtm
/approve

PR #925, opened 2021-08-25
Revert "feat: Set better meteor limit range (#911)"

tumido:
This reverts commit 6a3aa2c835e354dbe5ae489fe728ecdba8030409.

Meteor should scale better soon, hence we need the original limitranges back again.


4n4nd (approved):
/lgtm

PR #941, opened 2021-08-30
feat: enable use user workload monitoring

mimotej:
Resolves https://github.com/operate-first/SRE/issues/294


tumido (commented):
/hold

this will increase load on Zero cluster, maybe we should do this on Smaug instead?

cc @4n4nd thoughts?

PR #943, opened 2021-08-30
feat: Onboard DevConf.US 2021 cluster

tumido:
Part of: https://github.com/operate-first/support/issues/368

We decided to spin up a new cluster for DevConf.US workshop. This PR onboards the cluster to Operate First once available.

PR #944, opened 2021-08-30
update oauth client secret for devconfus2021 cluster

4n4nd:
Part of #943


tumido (changes requested):
Sorry @4n4nd , my changes didn't get saved in sops... Actually we want the UUID in this secret and change it in the `KeycloakClient` (I've copy pasted the `KeycloackClient` from Curator cluster - they both have the same secret there which should not be the case - please take the client secret in this oauth secret resource as the truth and change it in keycloak client.

My bad, sorry.

4n4nd:
updated :+1:

tumido (approved):
/lgtm
/approve

tumido:
Merging without CI

PR #958, opened 2021-09-01
feat: Allow operate-first project apps to be listed by public

tumido:
App of apps lives in operate first project for security reasons (deploys to argocd namespace), we need users to be able to see the app. It probably deserves it's own project, but this is quick and dirty solution for now


4n4nd (approved):
/lgtm

PR #966, opened 2021-09-01
fix oauth token for the grafana datasource in devconfus2021

4n4nd:
the existing token didn't seem to work for some reason


tumido (approved):
/lgtm
/approve

PR #978, opened 2021-09-10
feat: add sre to argocd-group

mimotej:
Add SRE group to policy.csv so they can use ArgoCD web interface.


HumairAK (approved):
/lgtm
/approve

PR #981, opened 2021-09-13
Re-encrypt groups in prod common with all gpg keys.

HumairAK:
These files were not encrypted with all the gpg keys found in their relative sops yaml file, this pr remedies that.


larsks (approved):
Works for me!

4n4nd (approved):
/lgtm

PR #991, opened 2021-09-14
feature/smaug nmstate

larsks:
- Add openshift-nmstate operatorgroup
- Add nmstate deployment


HumairAK:
looks like this needs to be rebased

Issue #952, opened 2021-08-31
ArgoCD diff in ACM app


tumido:
ArgoCD is diffing `KlusterletAddonConfigs` very inconsistently. Leaving the app out of sync for now, we need to revisit later.

In https://github.com/operate-first/apps/pull/950 we defined a bunch of resources. The diff was fine for all of them except 2 - curator and zero clusters:
![image](https://user-images.githubusercontent.com/7453394/131509118-f2cb9b21-ce72-439e-9d19-4b377cd929bf.png)

![image](https://user-images.githubusercontent.com/7453394/131509161-0b3465d8-abf6-459f-a08e-a46e51a940ba.png)

This is total nonsense. Not sure what's wrong.

tumido:
After discussions with @HumairAK we've found out this is a glitch, deleting those resources ArgoCD recreated them just fine.

PR #995, opened 2021-09-15
Sync klusteraddonconfigs with live.

HumairAK:
This should fix https://github.com/operate-first/apps/issues/952


4n4nd (approved):
/lgtm

tumido (changes requested):
can you add EOLs  please

tumido (approved):
/lgtm
/approve

tumido (commented):
any idea why these 2 clusters (compared to the rest) need a full manifest?

I also don't think we should be tracking the version strings if possible, since that is managed by the operator anyways...

PR #996, opened 2021-09-15
Configure NFS vlan on smaug workers

larsks:
This adds a systemd service that executes the necessary nmcli commands
at boot to configure a tagged interface on the br-ex ovs switch.

When this merges it will cause a rolling reboot of worker nodes.


4n4nd (approved):
/lgtm

PR #1003, opened 2021-09-16
Re-add hive.properties to osc trino catalog.

HumairAK:
Seems like this catalog is required for some reason, otherwise workers will not boot up, possibly coordinator as well though not tested.


4n4nd (approved):
/lgtm

PR #1006, opened 2021-09-16
migration of kfdefs: opf-superset, and opf-trino to smaug

Gregory-Pereira:
Not sure about namespaces that go with this, I checked on the cluster and couldn't find any there that related to any of these 3. Also now that PVCs are set up, we should double-check the memory request patches.

### DIFF
--- a/cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
+++ b/cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
@@ -28,6 +28,8 @@ resources:
   - ../../../../base/core/namespaces/opf-monitoring
   - ../../../../base/core/namespaces/opf-seldon
   - ../../../../base/core/namespaces/opf-slack-first
+  - ../../../../base/core/namespaces/opf-superset
+  - ../../../../base/core/namespaces/opf-trino
   - ../../../../base/core/namespaces/tekton-pipelines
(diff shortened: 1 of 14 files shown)


cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
```diff
@@ -14,6 +14,9 @@ resources:
   - ../../../../base/core/namespaces/opf-ci-pipelines
   - ../../../../base/core/namespaces/opf-ci-prow
   - ../../../../base/core/namespaces/opf-ci-test-pods
+  - ../../../../base/core/namespaces/osc-playground
```
HumairAK, line 17: @Gregory-Pereira can you remove this line, this ns is not needed anymore.

kfdefs/overlays/moc/smaug/osc-playground/kustomization.yaml
```diff
@@ -0,0 +1,23 @@
+---
```
HumairAK, line 1: as per the comment above, let's remove this kfdef

HumairAK: and the osc-playground folder entirely within kfdefs/overlays/moc/smaug

HumairAK (approved):
/lgtm
/approve

PR #1010, opened 2021-09-16
Add users and group mappings to osc trino.

HumairAK:
Resolve: https://github.com/os-climate/os_c_data_commons/issues/55


erikerlandson:
`aicoe_osc_demo` should not have blanket `All` access.

I think everyone can have "admin-like" access to the `demo1` catalog.

HumairAK:
@erikerlandson they don't, the catalog "all" is so that it doesn't restrict write access for tables

Issue #855, opened 2021-07-22
chore: Unify manifest for Zero and Rick


tumido:
Unify manifest duplications introduced in:
https://github.com/operate-first/apps/pull/852
https://github.com/operate-first/apps/pull/851
https://github.com/operate-first/apps/pull/854

- [ ] `OperatorGroups`, `Subscriptions` are duplicated and differ in either name or namespace
- [ ] OCS uses different channel on each cluster, upgrade Zero
- [ ] Local storage should use either `LocalVolumes` or `LocalVolumeSets`, not both if possible (just for the sake of unification)

tumido:
Let's adapt the manifests used by Rick also for the Smaug cluster and decomission the Zero manifests together with zero. That way we don't have to migrate existing setup on any of the clusters.

/cc @rbo @larsks

do you see any blocking difference here (like "this thing can't be used on Smaug, because..")

Note: of course we'll have to overlay device specific changes in manifests etc.

PR #1012, opened 2021-09-17
ceph migration to smaug

Gregory-Pereira:
may need to change storage cluster details based on our storage and PVCs on smaug.


tumido (commented):
Because we don't consume any local storage, this PR is not needed. See: https://github.com/operate-first/apps/issues/855#issuecomment-906705560
https://github.com/operate-first/apps/pull/1016#pullrequestreview-757312089

Is that correct, @larsks ?

/hold

larsks:
Correct, while we will be configuring OCS on Smaug, we will be consuming an external ceph cluster rather than using local storage.

PR #1014, opened 2021-09-17
kubeflow components smaug migration

Gregory-Pereira:
may need to change resource quota spec in kubeflow-components/overlays/moc/smaug/profiles/demo-profile.yaml depending on resources available on smaug.


tumido (commented):
/hold

We definitely don't want to pollute Smaug with Kubeflow's istio and certmanager just yet.

HumairAK:
> We definitely don't want to pollute Smaug with Kubeflow's istio and certmanager ~just yet.~

PR #1015, opened 2021-09-17
kubevirt-hyperconverged migration to smaug

Gregory-Pereira:


tumido (changes requested):
Missing subscription:

https://github.com/operate-first/apps/blob/674b89efe636aad473bcec1cbbf1d2a7cf1421f1/cluster-scope/overlays/prod/moc/zero/kustomization.yaml#L117

https://github.com/operate-first/apps/blob/674b89efe636aad473bcec1cbbf1d2a7cf1421f1/cluster-scope/overlays/prod/moc/zero/kustomization.yaml#L183

https://github.com/operate-first/apps/blob/674b89efe636aad473bcec1cbbf1d2a7cf1421f1/cluster-scope/overlays/prod/moc/zero/kustomization.yaml#L107

https://github.com/operate-first/apps/blob/674b89efe636aad473bcec1cbbf1d2a7cf1421f1/cluster-scope/overlays/prod/moc/zero/kustomization.yaml#L50

larsks:
I don't think we want to install the CNV operator on smaug unless there is an actual demand for it.

HumairAK:
> I don't think we want to install the CNV operator on smaug unless there is an actual demand for it.

agreed

PR #1016, opened 2021-09-17
local-storage smaug migration

Gregory-Pereira:
almost certainly will need to change storageClassDevices.devicePaths for smaug cluster. Looking for comments on that.


tumido (commented):
This PR is not needed. See: https://github.com/operate-first/apps/issues/855#issuecomment-906705560

Is that correct, @larsks ?

/hold

PR #1017, opened 2021-09-17
metallb smaug migration

Gregory-Pereira:
Will need to change address pool addresses with Smaug's IP address range.


tumido (commented):
/cc @larsks

larsks:
I don't think we should install metallb until there is an actual demand for it.

Gregory-Pereira:
> I don't think we should install metallb until there is an actual demand for it.

closing then.

PR #1019, opened 2021-09-17
Update curator to use common pull-secret

larsks:
The curator cluster was still configured to use Ilana's expired pull
secret.


4n4nd (approved):
/lgtm

PR #1022, opened 2021-09-17
opf-monitoring datasource, rolebindings, and role migrations to smaug

Gregory-Pereira:


Gregory-Pereira:
No longer needed, due to using user workload monitoring to scrape per namespace, rather than scraping everything from opf-monitoring.

PR #1026, opened 2021-09-19
wip: feat: add grafana to smaug cluster

mimotej:
Adding grafana to smaug grafanadatasource without token, that needs to be generated + without RBAC authentication which I want to add also in this PR. Otherwise rest can be reused from grafana deployment which was done for  DevConf cluster.
/cc @tumido 
Relates to: https://github.com/operate-first/argocd-apps/pull/211


4n4nd:
@mimotej this was added in https://github.com/operate-first/apps/pull/1062, so I think this PR can be closed now.

PR #1027, opened 2021-09-20
feat: Cleanup cert-manager config and sync from live state

tumido:
/cc @Gregory-Pereira @HumairAK @4n4nd 

So, I've synced up the live state from the Smaug cluster and split up the cert manager configs per env. The idea is to make the setup robust for multi-cluster deployments:

1. Have a single `Certificate` spec for all ingresses
2. Overlay the issued dns name per cluster
3. Use the same `Issuer` definition for all clusters in the same region, including the DNS service account secret

This manifests render 0 diff against live state on Smaug.


4n4nd (commented):
/lgtm

HumairAK (approved):
/lgtm
/approve

PR #1034, opened 2021-09-20
add singleuser-profiles-gpu-ui.yaml

erikerlandson:
enable gpu ui on the odh launcher for the os-climate cluster


erikerlandson:
We will also want to set up gpu taints and tolerations, per:
https://github.com/opendatahub-io/jupyterhub-singleuser-profiles/blob/master/docs/gpu_tolerations.md

HumairAK (approved):
/lgtm
/approved

PR #1039, opened 2021-09-21
clean up rules.json

erikerlandson:


odh-manifests/osc-cl1/trino/base/trino-config-secret.yaml
```diff
@@ -131,10 +130,10 @@ stringData:
       {
         "group": "aicoe_osc_demo|data_commons_project",
         "catalog": "demo1",
-        "privileges": ["SELECT", "INSERT", "DELETE"]
+        "privileges": ["SELECT", "INSERT", "DELETE", "OWNERSHIP"]
       },
       {
-        "privileges": ["SELECT"]
```
HumairAK, line 137: this will remove read only access for all the currently assigned groups for the different catalogs, iirc you wanted this no?

HumairAK: nvm -- I was notified this is the intended goal

PR #1040, opened 2021-09-21
Update trino version for osc cluster.

HumairAK:
Updates trino image version so we can get this change: https://github.com/trinodb/trino/pull/8641/commits 

There's some deprecated configs I had to remove to support this upgrade @rimolive for your reference. 

The upgrade also required us to identify the oauth issuer.


4n4nd (approved):
/lgtm

PR #1042, opened 2021-09-21
Update osc trino oauth configs.

HumairAK:
This change will force trino to use github usernames as the "user" identity within trino when using Oauth.


4n4nd (approved):
/lgtm

PR #1043, opened 2021-09-22
Add default catalog to trino and remove original hive catalog.

HumairAK:
The kustomization is to filter out the hive catalog deployment from odh-manifests repo.


tumido (approved):
/lgtm
/approve

PR #1047, opened 2021-09-23
Remove machineconfig that configured the NFS vlan

larsks:
We're going to be handling this with one or more nmstate configs
instead, since we no longer have a convenient bridge on each node to
which we can attach a port.


HumairAK (approved):
/lgtm
/approve

4n4nd (approved):
/lgtm

PR #1048, opened 2021-09-23
Add clustersets and gitops mgmt as per acm 2.3

HumairAK:
Okay so the way they are doing integration with argocd seems to have changed.

Old way: [here](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.2/html/manage_applications/managing-applications#gitops-pattern) where we just update the one field `applicationManager.argocdCluster`. When onboarding balrog, this didn't add the argocd secret. That's because...

The new way: [here](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.3/html/applications/managing-applications#gitops-config) 

Which requires you to have a cluster-set, which is in tech preview, but I set it up on live to test it out and it indeed added balrog to the argocd cluster list (i.e. created the secret). 

This required adding various new manifests as you can find in the PR. Read more about it [here](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.3/html/applications/managing-applications#gitops-config) and [here](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.3/html/clusters/managedclustersets). 

Anyways, the end result is to add a new cluster to argocd you need to just add a label to the corresponding cluster `ManagedCluster` resource. All clusters in this clusterset should be auto added to argocd.

### DIFF
--- a/acm/overlays/moc/infra/gitopsclusters/argocd-managed.yaml
+++ b/acm/overlays/moc/infra/gitopsclusters/argocd-managed.yaml
@@ -1,0 +1,13 @@
+apiVersion: apps.open-cluster-management.io/v1alpha1
+kind: GitOpsCluster
+metadata:
+  name: argocd-managed
+  namespace: argocd
+spec:
+  argoServer:
+    argoNamespace: argocd
+    cluster: local-cluster
+  placementRef:
+    apiVersion: cluster.open-cluster-management.io/v1alpha1
+    kind: Placement
+    name: argocd-managed-clusters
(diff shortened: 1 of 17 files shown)


4n4nd (commented):
/lgtm

HumairAK:
/hold 
for @tumido to review

tumido (approved):
/lgtm
/approve

... sigh.. it used to be so simple and now we have this mess... :slightly_frowning_face: :smile:

PR #1053, opened 2021-09-24
Add network configuration for NESE VLAN

larsks:
And also generate nodenetworkconfigurationpolicies from templates,
rather than performing a bunch of copy-and-paste.

### DIFF
--- a/cluster-scope/overlays/prod/moc/smaug/nodenetworkconfigurationpolicies/Makefile
+++ b/cluster-scope/overlays/prod/moc/smaug/nodenetworkconfigurationpolicies/Makefile
@@ -1,0 +1,20 @@
+INTERFACES = \
+  eno1 \
+  enp4s0f0np0 \
+  enp5s0f0np0 \
+  enp65s0f0np0 \
+  enp65s0f1np1
+
+CONFIGS = \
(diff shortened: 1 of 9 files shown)


larsks:
@HumairAK is this crazy talk? This includes a `Makefile` that generates the interface configurations from templates to avoid copy/paste errors. The output isn't dynamic (that is, we don't attempt to run this at build time); the generate files are added to the repository.

HumairAK:
seems like you're trying to get around templating by using makefiles lol 

if that's the case might as well just get a kustomize helm plugin so we don't have to resort to different makefiles

larsks:
> seems like you're trying to get around templating by using makefiles lol

I'm just trying to reduce the amount of repetitive work involved (as a way of reducing the chance for errors). I'm open to other ways of accomplishing this!

larsks:
> if that's the case might as well just get a kustomize helm plugin so we don't have to resort to different makefiles

Kustomize already has a helm plugin (at least, 4.3.0 does).  E.g., see <https://github.com/larsks/opf-apps/tree/feature/add-nese-vlan-helm/cluster-scope/overlays/prod/moc/smaug/nodenetworkconfigurationpolicies>.

PR #1061, opened 2021-09-28
add memory size options to os-climate cluster

erikerlandson:
xref: https://github.com/os-climate/os_c_data_commons/issues/65

### DIFF
--- a/kfdefs/overlays/osc/osc-cl1/jupyterhub/profiles/singleuser-profiles-container-sizes.yaml
+++ b/kfdefs/overlays/osc/osc-cl1/jupyterhub/profiles/singleuser-profiles-container-sizes.yaml
@@ -31,3 +31,27 @@ data:
         limits:
           memory: 32Gi
           cpu: 4
+    - name: '1 cpu, 2GB mem'
+      resources:
+        requests:
+          memory: '1Gi'
+          cpu: '1'
+        limits:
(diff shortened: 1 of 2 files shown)


erikerlandson:
Note, the original PR requested increased filesystem size: I do not know how to configure the PVC to change file size.

erikerlandson:
I see that there are some per-user specifications for PVCs, for example:
kfdefs/overlays/moc/smaug/opf-jupyterhub/pvcs/michaelclifford.yaml

1) Is there a way to change the default PVC resource request?
2) Is there a way make PVC resource request part of the singleuser profiles, along with memory and cpu?

cc @vpavlin

4n4nd:
@erikerlandson you should be able to set the default pvc size in the JH configmap ([example](https://github.com/operate-first/apps/blob/master/odh-manifests/zero/jupyterhub/base/jupyterhub-configmap.yaml))

erikerlandson:
cc @redmikhail

HumairAK:
@erikerlandson current cpu/mem usage here: https://console-openshift-console.apps.odh-cl1.apps.os-climate.org/k8s/cluster/nodes?orderBy=asc&sortBy=Memory

erikerlandson:
@HumairAK  if looks ok to you, can we merge?

PR #1065, opened 2021-09-28
Add jupyterhub idle culler to smaug

4n4nd:
Update JH idle culler to use upstream built image instead of building images locally


HumairAK (approved):
/lgtm
/aprove

PR #1066, opened 2021-09-28
Add a triage-party application

codificat:
Related to https://github.com/operate-first/SRE/issues/389

This is to add a definition / deployment of [triage-party](https://github.com/google/triage-party) to smaug.


triage-party/base/deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+                  name: triage-party-github-token
+                  key: token
+            - name: PERSIST_PATH
+              value: "/app/pcache"
+            - name: PERSIST_BACKEND
+              value: "disk"
+          volumeMounts:
+            - name: config
+              mountPath: /app/config
+            - name: pcahce
```
4n4nd, line 31: pcache*

triage-party/base/deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+              mountPath: /app/pcache
+      volumes:
+        - name: config
+          configMap:
+            name: triage-party-config
+        - name: pcahce
```
4n4nd, line 37: pcache*

codificat: thanks for the catch, pushing a fixed version

triage-party/base/deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+    matchLabels:
+      app: triage-party
+  template:
+    metadata:
+      labels:
+        app: triage-party
+    spec:
+      containers:
+        - name: triage-party
+          image: triage-party:latest
```
4n4nd, line 17: @codificat any reason we are using an imagestream here? I see that the imagestream is pointing to a fixed image in quay (`quay.io/thoth-station/triage-party:v1.4.0`) anyway. Would it be better to just use that and eliminate the ImageStream?

triage-party/overlays/moc/smaug/secret-generator.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: viaduct.ai/v1
+kind: ksops
+metadata:
+  name: thoth-secret-generator
```
4n4nd, line 4: Is this the thoth specific deployment?

codificat: Nope, that's the op1st deployment - will fix the name, thanks.

triage-party/base/route.yaml
```diff
@@ -0,0 +1,9 @@
+---
+kind: Route
+apiVersion: route.openshift.io/v1
+metadata:
+  name: triage-party
```
tumido, line 5: Can we get the route secured via `acme-controller` please? Just add the same annotation as you can see here: 

https://github.com/operate-first/apps/blob/8b54f97f86e2f9938a55486aa7267af62d22e5f3/cloudbeaver/overlays/emea/ai4ci/secure-route.yaml#L5-L6

triage-party/overlays/moc/smaug/secrets.enc.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: v1
+items:
```
tumido, line 2: There's no need to keep the secret in a `List`, can this resource be of `Secret` kind please?

tumido (changes requested):

tumido:
@codificat yeah, let's create new namespace with `x-small` quota.

triage-party/base/route.yaml
codificat: Sure, I can move the annotation(s) to the base if you prefer, but I added this in the kustomization within the overlay: the `Route` there contains things that are specific to the smaug cluster, like the host name for the route and this annotation (because smaug does have acme-controller).

Makes sense?

triage-party/overlays/moc/smaug/secrets.enc.yaml
codificat: Will do, thanks

codificat:
> @codificat yeah, let's create new namespace with `x-small` quota. I think `opf-triage-party` should be the name (to simplify it for ArgoCD)

Ok, added. I hope I got it right - my first one here so please review with extra :heart:, thanks.

It is associated with the operate-first project, so I don't have access to it within Argo. That's probably fine (?), but if you think I should add myself to the operate-first group just let me know.

codificat:
/hold remove

I believe this now has everything needed (pending review, of course)

codificat:
actually,
/hold cancel

codificat:
> @codificat you can use `/unhold` if this is ready for review

Thanks - it is ready for review, and `/hold cancel` did remove the hold label... I believe it's the same as
/unhold

4n4nd:
> > @codificat you can use `/unhold` if this is ready for review
> 
> Thanks - it is ready for review, and `/hold cancel` did remove the hold label... I believe it's the same as /unhold



> > @codificat you can use `/unhold` if this is ready for review
> 
> Thanks - it is ready for review, and `/hold cancel` did remove the hold label... I believe it's the same as /unhold

Ah my bad, this looks good to me. Could you just rebase?

PR #1067, opened 2021-09-28
Add my keys to smaug nodes

naved001:
* will come in handy if I need to diagnose network/infrastructure things


HumairAK (approved):
Does this mean I can also bother  @naved001 when something goes south with smaug?!? :)

/lgtm
/approve

PR #1070, opened 2021-09-28
Move argocd apps contents to apps repo.

HumairAK:
Resolves: https://github.com/operate-first/argocd-apps/issues/216
And also retires that repo. 

Upon merge, we need to make sure that `app-of-apps` are sync'd, so that they point to the `apps` repo . 

Everything is the same as `argocd-apps` except the following: 

- I have removed auto sync for all `app-of-apps` since we keep them off all the time anyway, let's keep our argocd-infra app in sync at least. 
- I removed the `opf-app-of-apps-zero` from the main kustomization, we can clean up the file it self when we do a thorough clean up of all zero manifests. 
- updated `app-of-apps` files to point to the apps repo instead of `argocd-apps`


tumido (commented):
/lgtm

4n4nd (approved):
/lgtm

PR #1074, opened 2021-09-29
fix: Attempt to ACM's sync for ArgoCD

tumido:
Attempting to fix https://github.com/operate-first/SRE/issues/391

1. The apiversion change should help with those errors
2. The connection secrets changed in format - this should make the AWS credentials sync properly again (not always report out of sync, because ACM migrated it to the newer format)
3. Automated sync does not prune, so the `ManagedClusterInfo` resources can stay


HumairAK (approved):
/lgtm
/approve

PR #1083, opened 2021-09-30
fix: Mark cephfs storageclass default on rick

tumido:
Follow up to https://github.com/operate-first/apps/pull/1075

Without specifying a storageclass, the PVC won't bound, since there's no default.


4n4nd (commented):
/lgtm

HumairAK (approved):
/lgtm
/approve

PR #1093, opened 2021-09-30
Add more rigorous access control to osc trino 

erikerlandson:


odh-manifests/osc-cl1/trino/base/trino-config-secret.yaml
```diff
@@ -92,23 +92,23 @@ stringData:
...
     {
       "catalogs": [
       {
-        "group": "admins|aicoe_osc_demo|data_commons_project",
+        "group": "admins|data_commons_project",
         "allow": "all"
       },
       {
-        "group": "os_climate_corporate_data_project|climate_itr_tool_project",
+        "group": "aicoe_osc_demo|os_climate_corporate_data_project|climate_itr_tool_project",
```
HumairAK, line 111: You might want to take a look at the table here: https://trino.io/docs/current/security/file-system-access-control.html#catalog-schema-and-table-access 

Basically, if you want people from `aicoe_osc_demo` to run `INSERT INTO` or `DELETE FROM` commands, then they need to have `all` access for catalog.

erikerlandson: In that case, all groups will need this.

PR #1096, opened 2021-09-30
Update argocd to 2.1.3, add application sets

HumairAK:
Resolve https://github.com/operate-first/SRE/issues/385
Upgrade for ArgoCD to 2.1.3. 

Not sure why argocd resources were being imported from the cluster-resources-infra app, but I moved them to be bundled by the argocd-infra app. I also added application sets as part of this bundle. 

Also snuck in a change to declarative ocp prod app of apps to sync with live state, which has automated syncs enabled.

### DIFF
--- a/argocd/base/kustomization.yaml
+++ b/argocd/base/kustomization.yaml
@@ -1,7 +1,8 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-- github.com/argoproj/argo-cd/manifests/base?ref=v2.0.1
+- github.com/argoproj/argo-cd/manifests/base?ref=v2.1.3
+- github.com/argoproj-labs/applicationset/manifests/base?ref=v0.2.0
 - routes/argocd_server_route.yaml
 - serviceaccounts/argocd-manager.yaml
 patchesStrategicMerge:
@@ -13,7 +14,7 @@ patchesStrategicMerge:
 images:
 - name: quay.io/argoproj/argocd
   newName: quay.io/operate-first/argocd
-  newTag: v2.0.1-1
+  newTag: v2.1.3-1
 - name: redis
   newName: quay.io/operate-first/redis
-  newTag: 6.2.1-alpine
+  newTag: 6.2.4-alpine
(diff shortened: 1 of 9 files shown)


argocd/base/kustomization.yaml
```diff
@@ -13,7 +14,7 @@ patchesStrategicMerge:
 images:
 - name: quay.io/argoproj/argocd
   newName: quay.io/operate-first/argocd
-  newTag: v2.0.1-1
+  newTag: v2.1.3-1
 - name: redis
   newName: quay.io/operate-first/redis
-  newTag: 6.2.1-alpine
+  newTag: 6.2.4-alpine
```
larsks, line 20: Why do we use a custom redis image?

cluster-scope/overlays/prod/moc/infra/kustomization.yaml
```diff
@@ -4,9 +4,6 @@ kind: Kustomization
 
 resources:
   - ../common
-
```
larsks, line 7: This appears to remove all the argocd resources from the infra cluster, without adding them back anywhere else?

larsks: Or these aren't necessary here, because we add them from `argocd/cluster-resources/kustomization.yaml`, and the entries in the infra overlay were redundant?

argocd/base/kustomization.yaml
tumido: because there was no upstream redis image available on quay or any other catalog than docker hub. And docker hub has rate limits... I don't know if the situation has changed

HumairAK: So we can pull from quay.io, instead of docker (to avoid rate limiting) also described [here](https://github.com/operate-first/apps/blob/master/docs/argocd/update_argocd.md#updating-redis-image)

cluster-scope/overlays/prod/moc/infra/kustomization.yaml
HumairAK: Yep it's the latter -- I added them to the argocd overlay kustomization. It made sense to have them there in my mind, because we need to be able to install argocd fully by using a manual `oc apply`.

4n4nd (approved):
/lgtm

PR #1099, opened 2021-09-30
fix: Diffing in ACM ArgoCD app

tumido:
Next patch to https://github.com/operate-first/SRE/issues/391 

This should fix the `ManagedClusterInfo` pruning. This PR makes ArgoCD ignore those resources completely, so they should never appear on the app screen again.


HumairAK (approved):
/lgtm
/approve

PR #1100, opened 2021-10-01
Add PVC full alerts for Jupyterhub users

4n4nd:
This will add PVC full alerts for jupyterhub users. These alerts will be sent to the github alertreceiver.
In these alerts, the Jupyterhub runbook is linked so that users can try to fix the issue by themselves.
Also, users will be tagged in the alert issue using their github handle.

Example alert issue: https://github.com/operate-first/alerts/issues/22570


HumairAK (approved):
/lgtm
/approve

PR #1116, opened 2021-10-04
Fix AICoE OSC demo ImageStream

chauhankaranraj:
This PR changes the tag from `v.10` to `latest` in the ImageStream for AICoE OSC demo, so that jupyterhub (on the OS-Climate cluster) has the latest image.


4n4nd (approved):
/lgtm

PR #1127, opened 2021-10-06
feature/nese routes

larsks:
- Remove routes from nmstate configuration
- Add static routes for accessing NESE storage (take 2)


larsks:
I think I've finally got the correct syntax on the DHCP server to configure the static routes, so this may not be necessary. Going to let it hang around a bit in case we end up needing it.

PR #1128, opened 2021-10-06
Remove routes from nmstate configuration

larsks:
nmstate won't allow us to add routes to an interface configured using
dhcp [1]. We'll have to set up routes via some other mechanism
(ideally via the DHCP server).

[1]: https://github.com/nmstate/nmstate/blob/a972b09a1d42b00f4195cacc90e372ea61c50244/libnmstate/route.py#L159


HumairAK (approved):
/lgtm
/approve

PR #1169, opened 2021-10-13
Cluster level roles and it's role bindings

skanthed:


HumairAK:
I'm guessing you are getting these Clusterroles/Clusterrolebindings from somewhere...just some feedback having names like `manager-role` and `report-editor-role` where the `kind` is in the `name` is redundant, we know it's a `role` no need to add `*-role` suffix, also `manager` is really vague, _lots_ of applications have _managers_, something like `curator-manager` would be more meaningful. Just my 2 cents :) 

Anyways since this is sourced from somewhere I guess it's not worth diverging from it's source material so I don't think it should block this PR.

PR #1175, opened 2021-10-13
Add VPA operator to neu-cs6620 namespace of smaug cluster

apoorvam:
### Changes

* Adding VerticalPodAutoscaler operator to Smaug cluster's neu-cs6620 namespace.


HumairAK:
In the [OCP Docs](https://docs.openshift.com/container-platform/4.8/nodes/pods/nodes-pods-vertical-autoscaler.html#nodes-pods-vertical-autoscaler-using-about_nodes-pods-vertical-autoscaler) I see the following: 

> On the Install Operator page, ensure that the Operator recommended namespace option is selected. This installs the Operator in the mandatory openshift-vertical-pod-autoscaler namespace, which is automatically created if it does not exist. 

Let's keep this in the recommended namespace.

PR #1186, opened 2021-10-15
add new physical risk trino group to os-climate

erikerlandson:
xref: https://github.com/os-climate/os_c_data_commons/issues/88


erikerlandson:
I think what needs to happen is we need a new "physical risk" user group and add Nikolaos to that.

erikerlandson:
Added a physical risk project group and associated access control

PR #1188, opened 2021-10-18
add low-privilege demo users and groups

erikerlandson:
adds:
- demo_user_a, demo_user_b, demo_user_c
- demo_group_a, demo_group_b, demo_group_c
- each user has same password `os-climate-rocks!`, which is low-value and therefore usable in demonstrations or labs.

the intent of the access control config is:
- `demo_group_*` has full privileges in schema `osc_datacommon_dev.demo`
- For all other schema, `demo_group_*` has no privileges.


erikerlandson:
cc @HumairAK

PR #1213, opened 2021-10-21
fix: Remove nmstate from Smaug, revert to CNV

tumido:
Since CNV is deployed to Smaug, we should remove nmstate operator to avoid both nmstate deployments fighting over it.

/cc @larsks @HumairAK 

/hold


4n4nd (commented):
/lgtm

HumairAK (approved):
/lgtm
/approve

PR #1216, opened 2021-10-21
feat: Autoscale balrog

tumido:
So.. I wanted to try some :dollar: on AWS.. let's see how it goes :smile: 

According to docs we can autoscale like this:

https://docs.openshift.com/container-platform/4.8/post_installation_configuration/cluster-tasks.html#cluster-autoscaler-about_post-install-cluster-tasks

/cc @harshad16 @HumairAK @larsks


tumido:
ACM can't do this as efficiently as OCP autoscaling (not based on load)

/unhold

PR #1250, opened 2021-10-29
Fix opf-datactalog objectbucketName to match auto generated name.

HumairAK:
Seems like this gets replaced by ocs, not sure if it's auto-generated but matching it with what's on live all the same.


4n4nd (approved):
/lgtm

PR #1255, opened 2021-10-30
re-sync trino user groups with github teams

erikerlandson:
Original goal: define a group to confine lseg data access to users known to be covered under NDA
New Goal: resync user groups with github teams, which will include `..._developer` user groups for each project. These `developer` groups should have access to LSEG data.

cc @caldeirav @MichaelTiemannOSC

### DIFF
--- a/odh-manifests/osc-cl1/trino/base/trino-config-secret.yaml
+++ b/odh-manifests/osc-cl1/trino/base/trino-config-secret.yaml
@@ -60,46 +60,24 @@ stringData:
     mikhail:$2y$10$kk/XazcoKmZGRvUwiM6kgex5QUEOTnVCmk5mU290gXTkArmEdAvLC
     ricardo:$2y$10$aLK8b1tH0bvXGk8GjU2EteokQQ4xH3e29PgyUTtiVsDUmCV2nHVpO
     vincent:$2y$10$93Lnt4vXk1sbhftl3fVE8euqyaCHPWmUrEdP99SRmrXK/XJ6UToDG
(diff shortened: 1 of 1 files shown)


caldeirav:
@erikerlandson so how I foresee our access rights management to work is a bit different:
- At Trino layer, access is driven by Github teams / groups, allocating to them all the different data sources they need. In the case of LSEG, we need itr-tool-project-team to have access and also our admin team.
- In addition to the above, we want to drive entitlements to the source data but would be best done not via Trino table but via data lineage / metadata. When a pipeline loads LSEG data, it should tag the source LSEG as metadata and we will use the compliance layer built on Apache Ranger to restrict the access consistently across the whole lineage this way. 

We could very much achieve the control objective currently with your proposal by using itr-tool-project-team and making sure the team itself only has people who have signed the LSEG agreement.

erikerlandson:
the `climate_itr_tool_project` trino usergroup has these members - are any of these NOT covered by the NDA?

climate_itr_tool_project:joriscram,alexanu,caldeirav,DavWinkel,ChristianMeyndt,hbaltzell,ttschmucker,toki8,LeylaJavadova,christian-schellnegger,roland-mf,franz-mf,EWinter21

erikerlandson:
Goal is to create trino user groups that correspond to "developer" sub-teams in os-climate github org, for example
`climate_itr_tool_project_developer`
`physical_risk_developer`
`corporate_data_developer`

HumairAK (approved):
/lgtm
/approve

PR #1256, opened 2021-10-31
Creating balrog datasource to enable smaug grafana to scrape balrog metrics

Gregory-Pereira:
Addresses part of [Thoth-station/Thoth-application issue 2029](https://github.com/thoth-station/thoth-application/issues/2029):

Metrics in Balrog:
    Grafana:
        - [x] update the grafana data source from balrog cluster to [smaug grafana](https://github.com/operate-first/apps/blob/master/grafana/overlays/moc/smaug/grafanadatasource.enc.yaml)


Gregory-Pereira:
Should I include other files such as [those available in the Smaug overlay but for Balrog](https://github.com/operate-first/apps/tree/master/grafana/overlays/moc/smaug)? I would assume not because we are simply pointing a data source to Smaug's existing and already configured grafana instance.

harshad16:
> Should I include other files such as [those available in the Smaug overlay but for Balrog](https://github.com/operate-first/apps/tree/master/grafana/overlays/moc/smaug)? I would assume not because we are simply pointing a data source to Smaug's existing and already configured grafana instance.

we don't need to create another overlay for balrog, so the balrog data source would be used in the smaug
so we need to create an additional datasource in the smaug data source yaml https://github.com/operate-first/apps/blob/46cae4e4f653d6eab1bd9c72298b0ded0fae18a7/grafana/overlays/moc/smaug/grafanadatasource.enc.yaml#L9
However, as it is another cluster we can't use the service URL
https://github.com/operate-first/apps/blob/46cae4e4f653d6eab1bd9c72298b0ded0fae18a7/grafana/overlays/moc/smaug/grafanadatasource.enc.yaml#L18
we would have to create a proper URL. 
/hold

Gregory-Pereira:
> we would have to create a proper URL.

Can this be done by simply modifying [the URL](https://github.com/operate-first/apps/blob/master/grafana/overlays/moc/smaug/grafanadatasource.enc.yaml#L18) to the openshift route: `https://grafana-openshift-monitoring.apps.smaug.na.operate-first.cloud` or would we have to implement a proxy or some other solution to expose the service?

4n4nd (changes requested):
It looks like you are making changes to the existing grafana datasource that is in Smaug cluster. I don't think this is what you want to do

4n4nd (changes requested):
It looks like you are making changes to the existing grafana datasource that is in Smaug cluster. I don't think this is what you want to do

4n4nd:
@Gregory-Pereira is the intention of this PR to be able to see metrics from the Balrog cluster in the Smaug Grafana instance?

Gregory-Pereira:
> @Gregory-Pereira is the intention of this PR to be able to see metrics from the Balrog cluster in the Smaug Grafana instance?

No, to the best of my understanding the intention is to add Smaug Grafana as a data source to Balrog, so it can use all the metrics already available there.

Gregory-Pereira:
Addressed by [operate-first/apps pr 1271](https://github.com/operate-first/apps/pull/1271)

PR #1266, opened 2021-11-02
Change octo aws secret key val to str.

HumairAK:
Should resolve: https://github.com/operate-first/SRE/issues/391#issuecomment-957954879


4n4nd (approved):
/lgtm

PR #1268, opened 2021-11-02
Remove unused jh pvcs from smaug.

HumairAK:
Preventing us from sync'ing JH app: https://argocd.operate-first.cloud/applications/kfdefs-smaug?resource=sync%3AOutOfSync 

As these users have not created a notebook since migration completed, we can increase it again once the need for them arises.


4n4nd (approved):
/lgtm

PR #1275, opened 2021-11-03
Add public bucket to trino on Smaug.

HumairAK:
Furthers: https://github.com/operate-first/support/issues/165#issuecomment-954836520
As per docs [here](https://github.com/operate-first/apps/blob/master/docs/odh/trino/add_more_buckets.md).


hemajv (commented):
/lgtm :ship: :rocket:

PR #1283, opened 2021-11-04
Re-organize smaug grafana dashboards.

HumairAK:
Move all dashboards within `smaug` since that is where they are deployed, no reason to keep them in base, let's keep them together. Also removed inactive dashboards (legacy cpu overview, and flowchart).


harshad16 (commented):
/lgtm
thanks :100:

Gregory-Pereira (commented):
👍  thanks Humair

PR #1284, opened 2021-11-04
Add routes for ipmi access to curator cluster

larsks:
The baremetal operator was complaining because it was unable to reach
the ipmi addresses of the cluster nodes. This adds the necessary
routes for access to the ipmi network.


HumairAK (approved):
/lgtm
/approve

PR #1294, opened 2021-11-05
Add action to sync hitchiker docs between repos.

HumairAK:
This is a suggestion to sync docs from apps to hitchhiker's guide automatically. It uses github actions, but ideally we would use thoth tooling to do this, this should serve as an interim solution.

Basically everytime any change is pushed to `docs` the bot account will push these changes to the `docs` folder in `hitchhiker's` galaxy. 

WDYT?

side note: ignoring workflows in pre-commit as part of this change due to [this issue](https://github.com/adrienverge/yamllint/issues/158), the pr is merged, but I tried it against latest yamllint 1.26.3 and it still failed.

### DIFF
--- a/.github/sync.yml
+++ b/.github/sync.yml
@@ -1,0 +1,3 @@
+operate-first/hitchhikers-guide:
+  - source: docs/
+    dest: docs/
--- a/.github/workflows/sync-action.yaml
+++ b/.github/workflows/sync-action.yaml
@@ -1,0 +1,17 @@
+name: Sync Files
+on:
+  push:
+    paths:
+      - 'docs/**'
+    branches:
+      - 'master'
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@main
+      - name: Run GitHub File Sync
+        uses: BetaHuhn/repo-file-sync-action@v1
+        with:
+          GH_PAT: ${{ secrets.OP1ST_TOKEN }}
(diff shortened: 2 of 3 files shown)


HumairAK:
/hold 
haven't created the token yet -- will create it once it's approved

durandom:
does the sync happen via force-push or as a PR?

HumairAK:
right now it's force push, but it _can_ be a PR (I prefer force push though)

4n4nd (approved):
/lgtm
Seems like a good enough interim solution

tumido:
I don't like this. What's the supposed course of actions then? Update the docs in apps? Update the docs in hitchhikers-guide? Does this job mean that any update to hitchhikers guide is meaningless and will be overwriten because any change in the apps/docs will just copy the files from apps instead of the updated content in hitchhikers guide?

HumairAK:
> I don't like this. What's the supposed course of actions then? Update the docs in apps? Update the docs in hitchhikers-guide? Does this job mean that any update to hitchhikers guide is meaningless and will be overwriten because any change in the apps/docs will just copy the files from apps instead of the updated content in hitchhikers guide?

Well, right now I'm questioning this PR for other reasons, as we are now in the midst of discussing the purpose of HitchHiker's guide vs Handbooks vs Docs. 

But: 

> Update the docs in apps? 
yes

> Does this job mean that any update to hitchhikers guide is meaningless and will be overwriten because any change in the apps/docs will just copy the files from apps instead of the updated content in hitchhikers guide?

Any update to the folder being sync'd will be meaningless from this repo, as soon as docs folder is updated in apps, it will be sync'd with the location specified here.

HumairAK:
No longer needed. Docs are instead added to this repo.

PR #1311, opened 2021-11-10
Add Jupyterhub resource tiers based on usage analysis

Shreyanand:
This PR updates the tiers for single user profiles using result from usage analysis in [notebooks](https://github.com/aicoe-aiops/operate-first-jupyterhub-analysis/blob/master/notebooks/resource_allocation/) here. 
We found three triers: Small, Medium, and Large based on clustering of user consumption. 

Based on observation of just spawned pods, the minimum CPU and memory were set to 100m and 0.5gi respectively. 

Related Issue https://github.com/aicoe-aiops/operate-first-jupyterhub-analysis/issues/12

Continuation of this [PR](https://github.com/operate-first/apps/pull/1156)

PR #1313, opened 2021-11-11
Update configmap

Shreyanand:
To make sure that the changes of this [PR](https://github.com/operate-first/apps/pull/1311) are reflected in the Jupyterhub spawner page, we have to update the config map to point at "smaug" instead of "zero".


4n4nd (approved):
/lgtm

PR #1314, opened 2021-11-12
docs: Updates to OCS documentation

larsks:
Provide some basic OCS troubleshooting documentation.

Closes operate-first/SRE#426

### DIFF
--- a/docs/ocs/claiming_object_store.md
+++ b/docs/ocs/claiming_object_store.md
@@ -7,7 +7,6 @@ This doc brings a brief summary on how to request a hosted S3 buckets within the
 Deploying a following resource within your application will grant you a bucket:
 
 ```yaml
----
 apiVersion: objectbucket.io/v1alpha1
 kind: ObjectBucketClaim
 metadata:
@@ -35,7 +34,7 @@ The auto created secret `CLAIM_NAME` contains 2 properties, which provide access
 
 The auto created configmap `CLAIM_NAME` contains 4 additional properties, which specifies means to access the bucket:
 
-- `BUCKET_HOST` which corresponds to an internal cluster route to the Rook operator deployment (`*.openshift-storage.svc.cluster.local`).
+- `BUCKET_HOST` which corresponds to an internal cluster route to the Noobaa service (usually `s3.openshift-storage.svc`)
 - `BUCKET_NAME` which holds the unique name (in the cluster) of the bucket, prefixed with what was specified in `spec.generateBucketName` of the `ObjectBucketClaim`
 - `BUCKET_PORT`
 - `BUCKET_REGION`
@@ -43,7 +42,7 @@ The auto created configmap `CLAIM_NAME` contains 4 additional properties, which 
 
 ### Usage in deployment
 
-In order to use the bucket within your deployment, you can mount the `Secret` and a `ConfigMap`:
+In order to use the bucket within your deployment, you can import the information from the `Secret` and `ConfigMap` as environment variables:
 
 ```yaml
 ...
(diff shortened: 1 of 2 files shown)


HumairAK:
/gentle-nudge
so it's not forgotten, would be nice to have :)

larsks:
@HumairAK how's this for a start?

docs/ocs/troubleshooting.md
```diff
@@ -0,0 +1,169 @@
...
+- Are existing PVs available?
+- If you create a new `PersistentVolumeClaim`, does it bind immediately to a PV?
+- Are you able to access existing object buckets?
+  - From inside the cluster?
+  - From outside the cluster?
+- If you create a new `ObjectBucketClaim`, does it bind immediately to
+  a new bucket?
+
+If *some* of the above steps work, there is (probably) problem a
+software problem on the cluster. Continue with "Troubleshooting
```
HumairAK, line 14: > there is (probably) problem a software problem on the cluster 

to

> there is (probably) a software problem on the cluster

docs/ocs/troubleshooting.md
```diff
@@ -0,0 +1,169 @@
...
+### Check ceph status
+
+Run the `ceph health` command:
+
+```
+$ ceph --id provisioner-moc-rbd-1 health
+HEALTH_WARN 5 clients failing to respond to cache pressure; 12 nearfull osd(s); 1 pool(s) full; 23 pool(s) nearfull; 1 pgs not deep-scrubbed in time; 19 pgs not scrubbed in time
+```
+
+What we care about here is that we get a valid response back from the
```
HumairAK, line 135: What is a valid response in this case, or what is not a valid response?

docs/ocs/troubleshooting.md
```diff
@@ -0,0 +1,169 @@
+# Troubleshooting OpenShift Container Storage (OCS)
```
HumairAK, line 1: Just to be consistent with out other trouble shooting docs, can we name this: 

`# OpenShift Container Storage (OCS) Runbook` 

and for the filename instead of `troubleshooting.md` name it `runbook.md` ?

docs/ocs/troubleshooting.md
```diff
@@ -0,0 +1,169 @@
...
+
```
HumairAK, line 2: Can we also have a Pre-requisite section? In this case it would be cluster admin to the cluster hosting ocs?

docs/ocs/troubleshooting.md
```diff
@@ -0,0 +1,169 @@
...
+You can log into the `rook-ceph-tools` container to run Ceph commands
+to diagnose Ceph cluster health and connectivity problems. To log into
+the pod:
+
+```
+oc -n openshift-storage rsh deploy/rook-ceph-tools
+```
+
+If this deployment isn't available, you can enable it by [editing the
+`OCSInitialization`
```
HumairAK, line 85: can we mention that to access this you need a red hat account? (if there is another non gated link, even better)

HumairAK (changes requested):

docs/ocs/troubleshooting.md
larsks: That seems odd to me; it's not really a runbook; it's *just* troubleshooting. I would expect a runbook to be a much more general document.

docs/ocs/troubleshooting.md
larsks: A valid response is anything that returns ceph server status.

docs/ocs/troubleshooting.md
larsks: I don't think there's any non-gated link for this.

docs/ocs/troubleshooting.md
HumairAK: @larsks -- then I'm not sure I understand the difference between troubleshooting and the purpose a runbook. Isn't the point of a runbook in large part to help with troubleshooting?

larsks: I don't know, I always thought a run book was "how to operate this thing", of which troubleshooting is only a part. I guess I consider all the documents in this directory to constitute parts of a run book for OCS.

HumairAK:
Awesome thanks @larsks !
/lgtm
/approve

PR #1315, opened 2021-11-12
added camel-k operator

durandom:
to the smaug cluster

I clicked install at https://operatorhub.io/operator/camel-k and added the contents of https://operatorhub.io/install/camel-k.yaml

anything more needed?

fixes https://github.com/operate-first/support/issues/476

### DIFF
--- a/cluster-scope/base/operators.coreos.com/subscriptions/camel-k/subscription.yaml
+++ b/cluster-scope/base/operators.coreos.com/subscriptions/camel-k/subscription.yaml
@@ -1,0 +1,10 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: camel-k
+spec:
+  channel: stable-1.7
+  installPlanApproval: Automatic
+  name: camel-k
+  source: community-operators
+  sourceNamespace: openshift-marketplace
--- a/cluster-scope/base/operators.coreos.com/subscriptions/camel-k/kustomization.yaml
+++ b/cluster-scope/base/operators.coreos.com/subscriptions/camel-k/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: openshift-operators
+resources:
+- subscription.yaml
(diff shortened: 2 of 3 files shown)


4n4nd (changes requested):
1. The subscription needs to be added in the [base](https://github.com/operate-first/apps/tree/master/cluster-scope/base/operators.coreos.com/subscriptions).
2. The patch generally contains the stuff that is different from the base, in case of operator subscriptions it is usually the `spec.channel` that is different.

cluster-scope/base/operators.coreos.com/subscriptions/camel-k/subscription.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: camel-k
+  namespace: operators
```
tumido, line 5: This namespace doesn't exist. I think you wanted to use `openshift-operators` instead. It's also generally better to enforce the namespace via `kustomization.yaml` (see this [Kiali example here](https://github.com/operate-first/apps/blob/master/cluster-scope/base/operators.coreos.com/subscriptions/kiali-operator/kustomization.yaml)).

cluster-scope/base/operators.coreos.com/subscriptions/camel-k/subscription.yaml
```diff
@@ -0,0 +1,11 @@
...
+spec:
+  channel: stable-1.6
+  installPlanApproval: Automatic
+  name: camel-k
+  source: operatorhubio-catalog
```
tumido, line 10: I don't think we have such catalog available here. In operator hub we have 2 options how to install camel-k:

1. comminuty version of the operator named `camel-k` from `community-operators` catalog
2. redhat certified version with name `red-hat-camel-k` from `redhat-operators` catalog

Based on channel selected I think you wanted to use `community-operators` catalog.

You may also consider updating the channel to `stable-1.7`


These are your options (package name and the catalog can be seen in the URL)

Either community operator:
image

Or the Red Hat version of it:
image

tumido (changes requested):

durandom:
@tumido thanks 💯 that was a decent review. I addressed the comments

tumido (approved):
/lgtm
/approve

PR #1316, opened 2021-11-12
Fybrik - adding cluster role binding for Vault

eletonia:
This PR adds the cluster role binding required to deploy Vault on the Smaug cluster. 

See reasoning below:
`vault-server-binding` is part of the Helm deployment of vault when using Kubernetes auth method https://wanwenli.com/kubernetes/2019/08/28/k8s-tokenreview-api-vault.html 

The Vault Kubernetes auth method (https://www.vaultproject.io/docs/auth/kubernetes#kubernetes-auth-method) is used to authenticate Kubernetes services against Vault using their Service Account Token. To enable it Vault needs to access Kubernetes TokenReview API to validate the service JWT.

For this reason Vault Service Account should be granted permissions to access this API with this cluster role binding.

@roee88 @ronenkat @HumairAK

### DIFF
--- a/cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/vault-server-binding/clusterrolebinding.yaml
+++ b/cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/vault-server-binding/clusterrolebinding.yaml
@@ -1,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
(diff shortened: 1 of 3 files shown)


HumairAK:
looks like you haven't rebased, and can you also squash to 1 commit :)

eletonia:
> looks like you haven't rebased with upstream master, and can you also squash to 1 commit :)

@HumairAK Done

HumairAK (approved):
/lgtm
/approve

PR #1325, opened 2021-11-16
Patch the pulp instance with node affinity to allow RDX storage mode

harshad16:

Related-to: https://github.com/operate-first/support/issues/456
Epic: https://github.com/operate-first/support/issues/176

The ceph storage in smaug doesn't allow volume mode: file system with the ReadWriteMany mode.
Alternative this was suggested to use ReadWriteOnce mode, to allow many pods to attach to this kinda storage the pods must be on the same node. Hence added the node affinity which the pulp operator allows
https://github.com/pulp/pulp-operator/blob/c3fb814aa6685e14d5b016830e680c908550b559/config/crd/bases/pulpproject_v1beta1_pulp_crd.yaml#L230

### DIFF
--- a/pulp/overlays/moc/smaug/opf-pulp.yaml
+++ b/pulp/overlays/moc/smaug/opf-pulp.yaml
@@ -7,13 +7,13 @@ spec:
   route_tls_termination_mechanism: Edge
   ingress_type: Route
   loadbalancer_port: 80
-  file_storage_size: 50Gi
   image_pull_policy: IfNotPresent
-  file_storage_storage_class: ocs-external-storagecluster-ceph-rbd
   image_web: 'quay.io/pulp/pulp-web:stable'
+  file_storage_size: 50Gi
+  file_storage_storage_class: ocs-external-storagecluster-ceph-rbd
+  file_storage_access_mode: ReadWriteOnce
   web:
     replicas: 1
-  file_storage_access_mode: ReadWriteMany
   content:
     log_level: INFO
     replicas: 2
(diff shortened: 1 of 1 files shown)


harshad16:
@HumairAK @larsks 
PTAL 
Last time, we discuss about the ReadWriteMany and ReadWriteOnce mode.

HumairAK:
does the node have to be specified explicitly? is there no way to say "schedule these pods together" regardless of node?

harshad16:
> does the node have to be specified explicitly? is there no way to say "schedule these pods together" regardless of node?

Based on the descriptions
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes
Multiple pods to be attached to the same PVC with type ReadWriteOnce require them to be on the same node.

if you have any other solution, please let me know
i m not aware any other way :sweat:

fridex:
@HumairAK your points are very reasonable. Nevertheless, would it be possible to have this merged? This is blocking us from testing the instance and eventually discovering other issues along the way that we could fix in parallel to this one. We will add this to our list and revisit this in the near future. Thanks in advance. 👍🏻

HumairAK (approved):
/lgtm
/approve

fridex:
> /lgtm /approve

Thanks, @HumairAK! 👍🏻

harshad16:
> If `oct-03-31-compute` were to go down, then this deployment is pretty much screwed. Is there no way to just have it "deploy these pods on the same node, regardless of which node it is" ?

i will look for the alternative and open a different pr with that :+1:

PR #1330, opened 2021-11-16
Creating Security Context Constraint required for Fybrik Vault

eletonia:
After seeing these errors when deploying Vault on Smaug using ArgoCD, we think adding this SCC will solve this https://github.com/hashicorp/vault-helm/issues/275  

@HumairAK @ronenkat @revit13

image


HumairAK:
/hold 

Not sure if this is the right way to go about this. Why does the container need to run as `uid: 100` ? That looks like it would be prone to collide with system uids...

Afaik you can instead update the pod/deployment's security context to be more openshift friendly. I'm not super familiar with security context however, maybe @larsks has better insight to offer here.

larsks:
@HumairAK I agree that in general it's better to fix the manifests so that the pod can run with an arbitrary UID rather than changing permissions to allow it to run with a fixed UID. It [looks like](https://github.com/operate-first/apps/blob/master/argocd/overlays/moc-infra/applications/envs/moc/smaug/fybrik/fybrik.yaml) we're currently deploying Fybrik directly from a Helm chart (I thought we didn't do that?), so the first step to fixing it is probably converting the helm chart to Kustomize manifests so that we can patch things.

HumairAK:
@larsks -- users with their own namespaces are welcome to choose how they want to organize their app manifests (they just can't deploy cluster/privileged resources..those need to go through `cluster-scope`), though we encourage the use of kustomize over helm. All the manifests _we_ manage, i.e. those that live in `operate-first` org ought to be done via kustomize

PR #1343, opened 2021-11-19
fix apicurio project

famartinrh:
server in apicurio project is incorrect and that made application to not deploy.

### DIFF
--- a/argocd/overlays/moc-infra/projects/apicurio-registry.yaml
+++ b/argocd/overlays/moc-infra/projects/apicurio-registry.yaml
@@ -7,7 +7,7 @@ metadata:
 spec:
   destinations:
     - namespace: 'apicurio-*'
-      server: 'smaug'
+      name: 'smaug'
   sourceRepos:
     - '*'
   roles:


argocd/overlays/moc-infra/projects/apicurio-registry.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
 spec:
   destinations:
     - namespace: 'apicurio-*'
-      server: 'smaug'
+      server: 'https://api.smaug.na.operate-first.cloud:6443'
```
HumairAK, line 10: whoops apologies I missed this last time, for `server` you need to have the url, but if you want to use the _name_ then you have to use the `destination.name` field, like [here](https://github.com/operate-first/apps/blob/master/argocd/overlays/moc-infra/applications/envs/moc/smaug/b4mad/racing.yml#L7)

famartinrh: np, it's fine now

HumairAK: > it's fine now 

Can you switch this to `name` instead? Like so:
```yaml
  destinations:
    - namespace: 'apicurio-*'
      name: 'smaug'
```

HumairAK (approved):
/lgtm
/approve

PR #1344, opened 2021-11-19
Fix thanos deployment sync issues

4n4nd:
1. Update resourcequota for `opf-observatorium` namespace
2. fix namespaces in thanos query servicemonitors
3. remove thanos query empty securityContext from deployment


HumairAK (approved):
/lgtm
/approve

Issue #1352, opened 2021-11-23
Pods being evicted in Nodes `oct-03-03-compute` and `oct-03-09-compute` on Smaug cluster


4n4nd:
Evicted pod statuses:
```
status:
  phase: Failed
  message: 'Pod The node had condition: [DiskPressure]. '
  reason: Evicted
  startTime: '2021-11-23T14:17:12Z'
```

```
status:
  phase: Failed
  message: >-
    The node was low on resource: ephemeral-storage. Container sparkoperator was
    using 31748Ki, which exceeds its request of 0. 
  reason: Evicted
  startTime: '2021-09-24T18:59:13Z'
```

HumairAK:
<details>

<summary> All evicted pods </summary> 

```
$ oc get pods -A -o=custom-columns=NAME:.metadata.name,Namespace:.metadata.namespace,STATUS:.status.reason,NODE:.spec.nodeName,REASON:.status.message | grep Evicted

meteor-srgls-jupyterbook-5c96b588c6-hcxh4                         aicoe-meteor                                       Evicted   oct-03-09-compute   The node was low on resource: ephemeral-storage. Container meteor-srgls-jupyterbook was using 56Ki, which exceeds its request of 0. 
apicurio-registry-mt-66b8f7b96b-5qhc5                             apicurio-apicurio-registry                         Evicted   oct-03-03-compute   The node was low on resource: ephemeral-storage. Container tenant-manager was using 36Ki, which exceeds its request of 0. Container envoy was using 448Ki, which exceeds its request of 0. Container apicurio-registry was using 12Ki, which exceeds its request of 0. 
apicurio-registry-mt-66b8f7b96b-l6ctr                             apicurio-apicurio-registry                         Evicted   oct-03-09-compute   The node was low on resource: ephemeral-storage. Container apicurio-registry was using 28Ki, which exceeds its request of 0. Container envoy was using 832Ki, which exceeds its request of 0. Container tenant-manager was using 36Ki, which exceeds its request of 0. 
apicurio-registry-mt-ui-mt-68765464c6-75spg                       apicurio-apicurio-registry                         Evicted   oct-03-03-compute   The node was low on resource: ephemeral-storage. Container apicurio-registry-mt-ui was using 52Ki, which exceeds its request of 0. 
[...]

PR #1353, opened 2021-11-23
Attach PVCs to openshift-monitoring and uwm prometheuses in MOC/Smaug

4n4nd:
Docs: https://docs.openshift.com/container-platform/4.8/monitoring/configuring-the-monitoring-stack.html#configuring-persistent-storage
Related https://github.com/operate-first/apps/issues/1352


HumairAK (approved):
/lgtm
/approve

larsks:
I would probably just make these bigger -- I mean, we *know* that metrics are already consuming about 300GB, so presumably we're going to end up there again.

anishasthana (approved):

PR #1364, opened 2021-11-24
remove '#' comments they break trino

erikerlandson:


larsks:
They don't just break trino -- comments in JSON are always invalid.

PR #1369, opened 2021-11-26
Add dev trino instance to osc-cl1.

HumairAK:
Resolves: https://github.com/os-climate/os_c_data_commons/issues/99 

This PR does the following: 
- Creates `odh-trino-dev` namespace on osc cluster
- Deploys a dev trino instance on the osc cluster in namespace `odh-trino-dev` 
- The bucket used as starting catalog is: `ocp-odh-os-demo-s3` (I.e. `OSC_DATACOMMONS_DEV`, which is also configured for the non-dev trino instance currently running, so I'm not sure if this poses as an issue) 

I'm also doing a trial run of pulling the configmaps out of `odh-manifests` and putting them in the `kfdefs` directory as separate files, I think this would make it easier to add validation in the future for files like `rules.json`, and if you have syntax highlighting/validation etc. in your ide, editing these files becomes easier. If it works without issues, I'll bring these changes to the non-dev instance currently on live.

### DIFF
--- a/cloudbeaver/overlays/osc/osc-cl1/configs/initial-data-sources.conf
+++ b/cloudbeaver/overlays/osc/osc-cl1/configs/initial-data-sources.conf
@@ -1,36 +1,6 @@
 {
     "folders": {},
     "connections": {
-        "sqlite_xerial-sample-database": {
-            "provider": "generic",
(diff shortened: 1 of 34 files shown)


redmikhail:
Approving but we need to keep an eye on resource usage

PR #1370, opened 2021-11-26
Correct conflict with configmap name for osc dev trino.

HumairAK:
SSIA

### DIFF
--- a/kfdefs/overlays/osc/osc-cl1/trino-dev/configs/kustomization.yaml
+++ b/kfdefs/overlays/osc/osc-cl1/trino-dev/configs/kustomization.yaml
@@ -8,7 +8,7 @@ configMapGenerator:
     files:
       - jmx.properties
       - catalogs/osc_datacommons_dev.properties
-  - name: trino-config
+  - name: trino-configfiles
     files:
       - config-coordinator.properties
       - config-worker.properties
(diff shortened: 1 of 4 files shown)


kfdefs/overlays/osc/osc-cl1/trino-dev/configs/kustomization.yaml
```diff
@@ -8,7 +8,7 @@ configMapGenerator:
     files:
       - jmx.properties
       - catalogs/osc_datacommons_dev.properties
-  - name: trino-config
+  - name: trino-configfiles
```
erikerlandson, line 11: would it be more clear to resolve conflict by naming with `trino-dev-*` prefix?

HumairAK: no the conflict is with a config map introduced from [here](https://github.com/opendatahub-io/odh-manifests/blob/master/trino/base/kustomization.yaml#L34) 

Basically I'm converting [this secret](https://github.com/opendatahub-io/odh-manifests/blob/master/trino/base/trino-config-secret.yaml#L5) to a `configmap`, but can't keep the name `trino-config`, since that conflicts with the previously existing configmap, which should probably be named `trino-params` instead or something, but that's a separate issue. 

I'll have to eventually do the same for our prod trino instance, when I separate out the files there. These files are in a separate namespace, so it's not related to the prod deployment.

erikerlandson (approved):

PR #1373, opened 2021-11-26
Configure a separate osc trino dev bucket catalog.

HumairAK:
SSIA


erikerlandson:
It seems correct but I don't understand what the purpose of renaming it is

erikerlandson (approved):

PR #1392, opened 2021-11-30
Add manual trigger for jbook action.

HumairAK:
Yaml looks a bit weird, but seems it's valid, not specifically mentioned in the docs, but you can see it in use here: 
https://github.com/operate-first/operate-first.github.io/blob/master/.github/workflows/build_job.yaml#L6


4n4nd (approved):
/lgtm

tumido (approved):
/approve
/lgtm

PR #1393, opened 2021-11-30
Added VPA Custom Resource manifest to opf-trino-stage.

laxmisravyarudraraju:


HumairAK:
Can you squash your commits to 1?

kfdefs/overlays/moc/smaug/opf-trino-stage/vpa/vpa-custom-resource.yaml
```diff
@@ -0,0 +1,22 @@
...
+  labels:
+    app: vpa-trino-stage-v1
+spec:
+  resourcePolicy:
+    containerPolicies:
+      - containerName: trino-worker
+        maxAllowed:
+          cpu: 3000m
+          memory: 3Gi
+        minAllowed:
```
HumairAK, line 15: Any particular reason you don't have a minAllowed set for memory?

kfdefs/overlays/moc/smaug/opf-trino-stage/vpa/vpa-custom-resource.yaml
```diff
@@ -0,0 +1,22 @@
...
+          cpu: 400m
```
HumairAK, line 16: Have you tested that worker's will run queries at `400m` ?

laxmisravyarudraraju: Yes, when we tested with long queries for around 10 min, the cpu utilization was around 200m. As we might run longer queries for longer duration, we set the min allowed as 400m to be on the safer side. 

As we don't have cpu usage most of the time, VPA automatically drops the cpu requests to 25m. And when there is a heavy load, it crashes because of low cpu. So, if we set the minAllowed, it tells VPA not to go below that value ever.

kfdefs/overlays/moc/smaug/opf-trino-stage/vpa/vpa-custom-resource.yaml
laxmisravyarudraraju: When we observed the historical data, the memory utilization hardly dropped below 500MB. So, we thought the minAllowed for Memory is not required as VPA by default sets the minAllowed as 250MB which is fine in our case. So we did not specify any value for Memory.

4n4nd:
looks good :+1: 
can you just [squash your commits to one](https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git)?

kfdefs/overlays/moc/smaug/opf-trino-stage/vpa/vpa-custom-resource.yaml
apoorvam: 250mb is not the minAllowed value https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L161 by default. However, from our observations, it never recommends less than 250mb.

HumairAK (approved):
/lgtm
/approve

PR #1395, opened 2021-12-01
Reformat smaug kustomization to fit the kustomize cli format.

HumairAK:
This shouldn't introduce any changes to the live environment, it's simply reformatting the file so that we don't see large diffs when people use `kustomize edit add`. It also fixes some of misplaced entries and places the alphabetically.


4n4nd (approved):
/lgtm

PR #1398, opened 2021-12-01
Configure VPA custom resource for opf-trino namespace's trino-worker

apoorvam:
@HumairAK / @4n4nd Current opf-trino (prod) request and limits are:
CPU: Request- 4 core, Limit- 8 core
Memory: Request: 9 GB, Limit: 16 GB

Based on this, min and max values that VPA can operate are configured in this PR. VPA UpdatePolicy will be set to Off for prod initially. Let me know if you think we should tune values differently.

```
maxAllowed:
          cpu: 8000m
          memory: 16Gi
        minAllowed:
          cpu: 1000m
          memory: 4Gi
```

These min/max values are per container of trino-worker. Ref: https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L165

### DIFF
--- a/kfdefs/overlays/moc/smaug/opf-trino/vpa/vpa-custom-resource.yaml
+++ b/kfdefs/overlays/moc/smaug/opf-trino/vpa/vpa-custom-resource.yaml
@@ -1,0 +1,24 @@
+---
+apiVersion: autoscaling.k8s.io/v1
+kind: VerticalPodAutoscaler
+metadata:
+  namespace: opf-trino
+  name: vpa-trino
+  labels:
+    app: vpa-trino
+spec:
+  resourcePolicy:
+    containerPolicies:
+      - containerName: trino-worker
+        maxAllowed:
+          cpu: 8000m
+          memory: 16Gi
+        minAllowed:
+          cpu: 1000m
+          memory: 4Gi
+  targetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: trino-worker
+  updatePolicy:
+    updateMode: "Off"
--- a/kfdefs/overlays/moc/smaug/opf-trino/kustomization.yaml
+++ b/kfdefs/overlays/moc/smaug/opf-trino/kustomization.yaml
@@ -1,3 +1,4 @@
+---
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 namespace: opf-trino
@@ -6,6 +7,7 @@ resources:
   - hive-metastores
   - obcs
   - secrets
+  - vpa
 patchesJson6902:
   - patch: |
       - op: add
--- a/kfdefs/overlays/moc/smaug/opf-trino/vpa/kustomization.yaml
+++ b/kfdefs/overlays/moc/smaug/opf-trino/vpa/kustomization.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: opf-trino
+resources:
+  - vpa-custom-resource.yaml


kfdefs/overlays/moc/smaug/opf-trino/vpa/vpa-custom-resource.yaml
```diff
@@ -0,0 +1,24 @@
...
+  namespace: opf-trino
+  name: vpa-trino
+  labels:
+    app: vpa-trino
+spec:
+  resourcePolicy:
+    containerPolicies:
+      - containerName: trino-worker
+        maxAllowed:
+          cpu: 8000m
```
4n4nd, line 14: Is this the max value allowed for both limits and requests?

apoorvam: @4n4nd Yes, this will be for both limits and requests. `containerPolicies` has another field called `controlledValues` and default of this is `RequestsAndLimits`. So these apply to both of it. 

ContainerResourcePolicy ref: https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L153

kfdefs/overlays/moc/smaug/opf-trino/vpa/vpa-custom-resource.yaml
```diff
@@ -0,0 +1,24 @@
...
+metadata:
+  namespace: opf-trino
+  name: vpa-trino
+  labels:
+    app: vpa-trino
+spec:
+  resourcePolicy:
+    containerPolicies:
+      - containerName: trino-worker
+        maxAllowed:
```
HumairAK, line 13: So to be clear, this is per container yes? 

I.e. each container can have minAllowed 1000m cpu each?

apoorvam: Yes, these min/max values are per container of trino-worker. Ref: https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L165

HumairAK: okay let's keep the maxAllowed to what's specified in the deployments because otherwise it might get capped by the resourcequota on the namespace (in other words vpa may get stuck trying to schedule pods if namespace is capped on resources as specified. 

We can't access smaug atm, but these values for the workers are defined here: https://github.com/operate-first/apps/blob/master/kfdefs/overlays/moc/smaug/opf-trino/kustomization.yaml#L15

apoorvam: Yes, updated maxMemory to 16Gi thus, max values match the limits.

HumairAK: So by default `controlledValues` is `RequestsAndLimits`, but it seems like `limits` are scaled [proportionally to the requests](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L205), does that mean the amounts specified in `maxAllowed` are for `requests`, and `limits` values are inferred by the original `deployment` ?

apoorvam: From our observations, it's true that proportionality of requests and limits is maintained, but then if either limits or requests go outside min/max values, they are capped. So min/max values apply to both requests and limits. 
When a new container is created, requests and limits are set to values in `deployment`. However, they both will be scaled by VPA as it sees fit.

There are scenarios where both requests and limits were set to same values by VPA, however deployment had a ratio of 1:2 between them. This is because of capping with min/max values.

HumairAK: > From our observations, it's true that proportionality of requests and limits is maintained, but then if either limits or requests go outside min/max values, they are capped.  

Based on this, it seems like `requests` cannot go beyond what is specified in the deployment if you're using `requestsandlimits` right?

apoorvam: No, values in deployment are used only when pod is created new and vpa is yet to recommend any values. Once vpa controls it, min/max are the only limits. Proportionality is not guaranteed when either go beyond min/max values.

VPA config: 
```
Spec:
  Resource Policy:
    Container Policies:
      Container Name:  trino-worker
      Max Allowed:
        Cpu:     3000m
        Memory:  3Gi
      Min Allowed:
        Cpu:  400m
```
Recommendation:
```
  Recommendation:
    Container Recommendations:
      Container Name:  trino-worker
      Lower Bound:
        Cpu:     400m
        Memory:  3Gi
      Target:
        Cpu:     400m
        Memory:  3Gi
      Uncapped Target:
        Cpu:     25m
        Memory:  4506574562
      Upper Bound:
        Cpu:     400m
        Memory:  3Gi
```

In this recent recommendation by VPA, actual recommended value is 25m CPU, but minAllowed CPU is 400, so target is set to 400m CPU. Similarly memory is outside maxAllowed, so it caps it to maxAllowed value of 3Gi. 

For this particular example, deployment values are:
```
```

apoorvam: This explains it better: 

https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler#examples

Your right with previous point that request is recommended by VPA. But limits are set based on proportionality in deployment and limit range.

HumairAK: Okay, so still not clear on how this policy enactment trickles down to the container, say you have: 

spec from original deployment, with 1:2 ratio: 
```
    Limits:
      cpu:     8
    Requests:
      cpu:     4
```
so now when deployed, pod has 4 requests, 8 limits on cpu. 

and let's say your policy is: 
```
      Max Allowed:
        Cpu:     8
```
and vpa reports: 
```
      Uncapped Target:
        Cpu:     9
```
based on your previous comment, `Target` will be reduced to `8`, yes?


So then, will the values for the container be: 
```
    Limits:
      cpu:  8
    Requests:
      cpu: 8
```
?

apoorvam: @HumairAK For this example, values for container will be:

```
limits: 
  cpu: 8
requests:
  cpu: 4
```
Recommendation of 9 cannot be set as that's beyond max. So it sets limit to 8 and request to 4, to keep the proportionality. 

If maxAllowed was 12 here, it will set limit to 12 and request to 6. (Uncapped target of 9)

HumairAK: Gotcha, so as long as `maxAllowed` matches the limit, _then_ `requests` will not surpass the `requests` in the deployment spec? 

I ask this because, we need to ensure that VPA limits/requests are not tuned in such a way that the `resourcequota` for the namespace is not respected. This seems to be the only way to ensure that as far as I can tell, unless VPA already takes quotas in to account.

apoorvam: Yes. VPA is not aware of the resource quotas. So setting boundaries with this should avoid problems. One of the limitations say: 

> VPA recommendation might exceed available resources (e.g. Node size, available size, available quota) and cause pods to go pending

apoorvam: @HumairAK / @4n4nd Please let me know if this looks good now. (Just following since its Friday, so we can monitor the VPA over weekend if this is merged)

HumairAK: yep thanks for the clarification, this lgtm

HumairAK:
/lgtm
/approve

merging, ci reporting failures due to smaug outage -- manually ran tests, and confirmed it's good.

PR #1318, opened 2021-11-15
feat: Add kubeval validation job to .prow.yaml

mimotej:
Adding job which will use script defined here https://github.com/operate-first/toolbox/pull/55.
Part of: https://github.com/operate-first/SRE/issues/107

PR #1400, opened 2021-12-02
fix: Fix issues in manifests in apps repo

mimotej:
This PR tries to fix various issues in our manifests. Changes were made, when CI which will be introduced in PR https://github.com/operate-first/apps/pull/1318, failed with error on some resource. 
Part of: https://github.com/operate-first/SRE/issues/107

### DIFF
--- a/cluster-scope/overlays/prod/moc/zero/nodenetworkconfigurationpolicies/moc-nfs-network.yaml
+++ b/cluster-scope/overlays/prod/moc/zero/nodenetworkconfigurationpolicies/moc-nfs-network.yaml
@@ -3,7 +3,6 @@ kind: NodeNetworkConfigurationPolicy
 metadata:
   name: moc-nfs-network-enp4s0f1np1
 spec:
-  parallel: true
   nodeSelector:
     moc/primary-interface: enp4s0f1np1
   desiredState:
@@ -24,7 +23,6 @@ kind: NodeNetworkConfigurationPolicy
 metadata:
   name: moc-nfs-network-enp4s0f1
 spec:
-  parallel: true
   nodeSelector:
     moc/primary-interface: enp4s0f1
   desiredState:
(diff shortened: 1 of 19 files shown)


HumairAK:
/lgtm

Let's also get @larsks  to take a look (in particular the NNCP fixes)

HumairAK:
Side note: I don't think we need 7 commits for this, one commit stating "Update resources to fit openapi specs" or something is fine I think.

cluster-scope/overlays/prod/moc/zero/nodenetworkconfigurationpolicies/moc-nfs-network.yaml
```diff
@@ -3,7 +3,6 @@ kind: NodeNetworkConfigurationPolicy
 metadata:
   name: moc-nfs-network-enp4s0f1np1
 spec:
-  parallel: true
```
larsks, line 6: We were setting `parallel: true` explicitly to work around [a problem][1] (see a37d254a2d). I suspect it's safe to remove, although I'd like to follow up with someone on the networking side to see if this is actually a bug in the openapi spec.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1967771

mimotej: Tried tests again just to be sure and it really gives an error (on picture below) which might suggest a bug in the openapi spec... Anyway these changes are done so that we don't have to delete whole zero (because rest of manifest seem fine), so it shouldn't matter too much anyway.
image

mimotej:
> Side note: I don't think we need 7 commits for this, one commit stating "Update resources to fit openapi specs" or something is fine I think.

Good point :+1: . Squashed and rebased commits. Which showed me that there were some changes in schemas, so I will create a PR with new schemas to be add to schema-store.

larsks (approved):

PR #1409, opened 2021-12-03
Add new gpg key and remove unneeded ones for emea rick cluster-scope.

HumairAK:
This will allow @rbo to encrypt/decrypt emea/rick kustomizations using sops/ksops. 

I also removed other gpg keys besides the op1st key, don't think we need them.


Gregory-Pereira:
Pre-commit succeeds and kustomize build works locally.

/Igtm

PR #1414, opened 2021-12-07
rules for row-based data vault

erikerlandson:


erikerlandson:
xref: https://github.com/os-climate/data-platform-demo/issues/36

Gregory-Pereira:
kustomize build fails, missing the `aws-secret.yaml` file in `apps/odh-manifests/osc-cl1/trino/base` .

erikerlandson:
@HumairAK   has something changed about the expected structure that would cause the errors @Gregory-Pereira is listing above?

Gregory-Pereira:
Can't speak for Humair, but I checked out the zero, smaug, and smaug-test environment directories for Trino deployment, and they all looked pretty much the same, but I am pretty sure that the odh-manifests used to be deployed either somewhere else or in some other way. Do you remember trino being deployed correctly on zero at some point?

File differences per environment: zero didn't have a kustomization file and was missing a route, and osc was only missing a route in comparison to smaug.

HumairAK:
@Gregory-Pereira `odh-manifests` is a niche case where it's actually deployed via ODH operator and not ArgoCD. 

ODH operator is combining these manifests with manifests found [here](https://github.com/opendatahub-io/odh-manifests). To get a complete picture you have to run a `kfctl build` command, which results in the `kustomization` build files to which you can then run `kustomize` build on. The process _is_ a bit convoluted. But the takeaway is that this directory is not meant to be built using `kustomize` directly, it's for ODH operator to manage.

HumairAK (approved):
/lgtm
/approve

PR #1416, opened 2021-12-07
Add cert manager and issuer & cert to infra cluster.

HumairAK:
Should fix: https://github.com/operate-first/SRE/issues/435

I left the IAM cred values I need for the issuer/secret empty until this is resolved: https://github.com/operate-first/SRE/issues/433

Because infra is not yet using a domain from the gcp instance, I didn't use the common files, though we should at some point switch over. But I'm guessing that would require us to update keycloak and other locations, let's for now stick with the moc domain, so we can unblock the issue above, then we can have a follow up issue to switch the operate-first domain and use the gcp creds/hostname, like for smaug.


HumairAK:
/hold 
for https://github.com/operate-first/SRE/issues/433

cert-manager/overlays/moc/infra/api-cert.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: api-certificate-letsencrypt
+  namespace: openshift-config
```
4n4nd, line 5: @HumairAK will we need a separate issuer in `openshift-config` namespace for this cert?

cert-manager/overlays/moc/infra/default-ingress-cert.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: default-ingress-certificate
+  namespace: openshift-ingress
+spec:
+  issuerRef:
+    name: ingress-letsencrypt
+  secretName: default-ingress-certificate-letsencrypt
```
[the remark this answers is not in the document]
larsks: That's correct, the secret name here needs to match what we configure in https://github.com/operate-first/apps/blob/master/cluster-scope/base/operator.openshift.io/ingresscontrollers/default/ingresscontroller.yaml

cert-manager/overlays/moc/infra/openshift-ingress-issuer.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: cert-manager.io/v1
```
larsks, line 1: Why not just create a `ClusterIssuer` rather than two identically configured namespaced-issuers?

cluster-scope/overlays/prod/moc/infra/kustomization.yaml
```diff
@@ -16,11 +16,11 @@ resources:
   - ../../../../base/core/namespaces/openshift-nmstate
   - ../../../../base/core/namespaces/opf-alertreceiver
   - ../../../../base/nmstate.io/nmstates/nmstate
-  - ../../../../base/operator.openshift.io/ingresscontrollers/default
```
larsks, line 19: I don't think this should be deleted.

cluster-scope/overlays/prod/moc/infra/kustomization.yaml
```diff
@@ -35,7 +35,7 @@ resources:
   - ../../../../base/storage.k8s.io/csidrivers/org.democratic-csi.nfs
   - ../../../../base/storage.k8s.io/storageclasses/moc-nfs-csi
   - ../../../../base/user.openshift.io/groups/cluster-admins
-
+  - apiserver/api_server_cert.yaml.yaml
```
larsks, line 38: I think you have too many .yamls here.

cluster-scope/overlays/prod/moc/infra/ingresscontrollers/default-ingress-certificate.enc.yaml
```diff
@@ -1,95 +0,0 @@
-apiVersion: v1
```
larsks, line 1: You've deleted this file, but it needs to be removed from `secret-generator.yaml`

cluster-scope/overlays/prod/moc/infra/kustomization.yaml
HumairAK: woops, I intended to delete the cert, not the controller. Thanks!

cluster-scope/overlays/prod/moc/infra/kustomization.yaml
HumairAK: yikes, thanks updated

cert-manager/overlays/moc/infra/openshift-ingress-issuer.yaml
HumairAK: Yeah, I suppose I am trying to be consistent with the way we deployed it on smaug, if I change it to `clusterissuer` then we''d have ClusterIssuer's in `cluster-scope` and then `issuers` in `cert-manager` app,

4n4nd: if you set up a `ClusterIssuer` in a cluster, is it available to be used in any namespace?
If yes, I think we should stick to using single issuers to prevent other users from being able to use this letsencrypt account

HumairAK: How about in order to maintain consistency we go with `issuer` for now, and split off a separate discussion in a different issue, we'll have to reorganize the cert-manager directory at some point anyway as it's a mess right now.

larsks (approved):

PR #1417, opened 2021-12-07
Adding user to Fybrik user group on Smaug cluster

eletonia:
This PR adds a user to the Fybrik user group on the Smaug cluster

@michaelfactor @ronenkat @HumairAK


HumairAK:
FYI due to https://github.com/operate-first/SRE/issues/430 and https://github.com/operate-first/support/issues/493 

Right now this change can't be brought to live, once those outages are resolved then this will be deployed.

PR #1423, opened 2021-12-09
Restructure moc infra cert manager resources.

HumairAK:
Follow up to: https://github.com/operate-first/apps/pull/1416

The only actual change to live is adding the iam creds to the `openshift-config` namespace (which I already added manually on live to get the api issuer to work). The rest is just a little big of juggling the files around so it's a little cleaner. A more robust restructuring should be followed after an issue.


4n4nd (approved):
/lgtm

PR #1443, opened 2021-12-14
Enable prune for ocp prod argocd app.

HumairAK:
This seems to already be enabled on live, the pr just sync's this with git.


larsks:
/lgtm

"Prune" means "delete things that are no longer part of the app", right?

HumairAK:
@larsks afaik yes. This is how I understand it works if you're curious: 

- When argocd deploys manifests from git, it will add this label to each of those resource manifests: `app.kubernetes.io/instance: <app-name>`
- Suppose you remove a resource from git. It will notice that there's a resource in this app (by checking for this label), but it's not in git, so it concludes that it was removed, and thus remove it. 

I think you could test this out by creating a resource with that label for a particular app, and see argocd tries to remove it, to verify this behavior.

PR #1446, opened 2021-12-14
Add containerfile for gitops-books builder image

### LINKED ISSUE #1445: Add notebook image for gitops book in smaug jh
We need to use the aicoe-ci app to build the image that contains the notebooks in `docs/content/notebooks`, then add this image to JH, and retire the HitchHiker's guide notebook as per: https://github.com/operate-first/SRE/issues/424

4n4nd:
Update aicoe-ci config to build gitops-books image

Resolves #1445

### DIFF
--- a/.aicoe-ci.yaml
+++ b/.aicoe-ci.yaml
@@ -1,2 +1,15 @@
 ---
-check: []
+check:
+  - overlays-build
+overlays_dir: docs/content
+overlays:
+  - name: notebook
+    build:
+      build-stratergy: Source
+      build-source-script: "image:///opt/app-root/builder"
+      base-image: quay.io/operate-first/s2i-sre-notebook:latest
+      custom-tag: latest
+      registry: quay.io
+      registry-org: operate-first
+      registry-project: op1st-gitops-books
+      registry-secret: operate-first-pusher-secret
--- a/docs/Containerfile
+++ b/docs/Containerfile
@@ -1,0 +1,9 @@
+FROM quay.io/operate-first/opf-toolbox:v0.4.2 AS toolbox
+FROM quay.io/thoth-station/s2i-custom-py38-notebook:v0.4.1
+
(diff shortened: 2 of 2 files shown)


.aicoe-ci.yaml
```diff
@@ -1,2 +1,15 @@
 ---
-check: []
+check:
+  - overlays-build
+overlays_dir: docs/content
```
HumairAK, line 4: @harshad16 just want to confirm, is this the only way to specify a path to the build path? 

Is there anyway we can have this _per overlay_, so that in the future we can have multiple builds if we need it, pointing the build working directory to different parts of the `apps` repo?

HumairAK: Also shouldn't it be: 
```
overlays_dir: docs
```
?

HumairAK: Based on our discussion in call. It doesn't seem like we need to use `overlays_dir` here. 

Also, we don't want s2i to copy the whole repo, but to force it to only copy `notebooks` (what we want) we need to be able to specify the `context_path` var, as per @harshad16. It is currently not possible to specify this in the `aicoe-ci.yaml`.

A workaround suggested by @4n4nd is to just update the `source` strategy to `Dockerfile` and write our own dockerfile that copies the contents from the notebooks folder into the image.

Gregory-Pereira: > @harshad16 just want to confirm, is this the only way to specify a path to the build path?
> 
> Is there anyway we can have this _per overlay_, so that in the future we can have multiple builds if we need it, pointing the build working directory to different parts of the `apps` repo?

Seconding that I don't believe we can do this without extending the AICoE app to support different paths from which to build certain overlays. Also, I believe we discussed that workaround briefly in Face In the Clouds so I know that Harshad is aware of it and said it could work.

HumairAK:
Don't need this anymore, we are going in a different direction from notebooks.  closing.

PR #1450, opened 2021-12-15
Increase osc trino memory req & limit.

HumairAK:
SSIA


HumairAK:
/hold

making changes

PR #1453, opened 2021-12-16
Add users to test virtual machine capabilities

cooktheryan:
Users need to be added to see if the machine learning vm will work before requesting NS and quota

### DIFF
--- a/cluster-scope/base/user.openshift.io/groups/octo-ushift-dev/group.yaml
+++ b/cluster-scope/base/user.openshift.io/groups/octo-ushift-dev/group.yaml
@@ -7,5 +7,5 @@ users:
     - cooktheryan
     - oglok
     - screeley44
-    - mcliffor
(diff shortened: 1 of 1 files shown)


cluster-scope/base/user.openshift.io/groups/octo-ushift-dev/group.yaml
```diff
@@ -7,3 +7,5 @@ users:
     - cooktheryan
     - oglok
     - screeley44
+    - mcliffor
```
HumairAK, line 10: doesn't seem like this user exists? was this supposed to be `mclifford` ?

cooktheryan: @HumairAK patched 

good catch

HumairAK (approved):
/lgtm
/approve

PR #1457, opened 2021-12-16
Add access to Grafana for the thoth team

codificat:
There are a few Grafana dashboards for the Thoth team, but the team is not explicitly listed as a group with access.

In particular, as I am not part of any of the current groups, this means I don't have access to Grafana :innocent: 

This PR is to add the `thoth` team to Grafana.


4n4nd (approved):
/lgtm

PR #1458, opened 2021-12-16
Add new pub bucket info in public access docs.

HumairAK:
Bucket had to be recreated due to: https://github.com/operate-first/SRE/issues/430#issuecomment-987040127

Resolve: https://github.com/operate-first/support/issues/499


Gregory-Pereira (approved):
/Igtm 👍

PR #1472, opened 2021-12-20
Update smaug monitoring configs to use default storageclass for PVCs

4n4nd:
This will make `openshift-monitoring` and `openshift-user-workload-monitoring` prometheuses on Smaug use the default storage class for their PVCs
Related https://github.com/operate-first/SRE/issues/416


HumairAK (approved):
/lgtm
/approve

PR #1475, opened 2021-12-21
Make trino configurations on smaug compatible with odh 1.1.1 updates.

HumairAK:
Related: #1257 and https://github.com/operate-first/apps/issues/1476


4n4nd (approved):
/lgtm

Issue #1489, opened 2022-01-06
Remove manifests and references to Zero cluster


HumairAK:
As the zero cluster does not exist any more, we should clean up the lingering manifests and references to it.

PR #1490, opened 2022-01-06
Remove zero cluster manifests and references.

HumairAK:
Resolve: https://github.com/operate-first/apps/issues/1489


kfdefs/base/monitoring/alertmanager.yaml
```diff
@@ -5,6 +5,6 @@ metadata:
   name: odh-monitoring
   namespace: opf-monitoring
 spec:
-  externalUrl: "http://alert-manager-opf-monitoring.apps.zero.massopen.cloud/"
+  externalUrl: ""
```
Gregory-Pereira, line 8: I know were waiting on figuring out what to do with kfdef/base but shouldn't we just change this value to where the alertmanager is currently deployed: `https://alertmanager-main-openshift-monitoring.apps.smaug.na.operate-first.cloud`

HumairAK:
Intentionally left alone the NNCP in the infra cluster https://github.com/operate-first/apps/tree/master/cluster-scope/overlays/prod/moc/infra/nodenetworkconfigurationpolicies that make mention of the zero cluster

And also left alone https://github.com/operate-first/apps/tree/master/kfdefs/overlays/emea/rick/opf-monitoring which seems to remote write to an old zero endpoint, but I think we need to remove the whole `opf-monitoring` from rick as it's not being used, will do this in a follow up pr.

Gregory-Pereira (approved):
/LGTM.

4n4nd (approved):
/lgtm

PR #1492, opened 2022-01-06
Remove arrow flight module argoCD application

eletonia:
This PR removes the arrow flight module argoCD application since it is already deployed by Fybrik's blueprint controller

@revit13 @roytman @HumairAK


HumairAK (approved):
/lgtm
/approve

PR #1504, opened 2022-01-07
Restructure smaug trino to move cfg mgmt to argocd.

HumairAK:
Restructure trino configs so that they are managed by argocd, and stored in a syntax/editor friendly format (as separate files), for the smaug cluster. 

As an aside I also noticed that the files for osc trino located [here](https://github.com/operate-first/apps/tree/master/kfdefs/overlays/osc/osc-cl1/trino/configs) were not being consumed, the updates to the folder `odh-manifests/osc-cl1/trino/base` should set the deployments to use these configmaps.


HumairAK:
@redmikhail @erikerlandson fyi, the changes for osc trino can be found in these directories: `odh-manifests/osc-cl1/trino/base`

4n4nd (approved):
/lgtm

PR #1520, opened 2022-01-15
Argocd notifications slack integration

Gregory-Pereira:
Waiting on rebasing until I get some feedback with this as I am sure I am missing some things here.

Related to [Investigating ArgoCD-Notifications issue](https://github.com/operate-first/SRE/issues/410). Meant to be the smallest slice towards a solution here.

Also waiting on removing some slack attachments from `argocd-notifications-cm` for all templates that are not `on-sync-running` because I could not access quick lab for most of today, so I couldn't test changes to the templates. We want to do this to reduce extra noise, so the initial notification would be the information on the sync running, and then each other condition as a small message with the resolved status.

### DIFF
--- a/argocd-notifications/base/kustomization.yaml
+++ b/argocd-notifications/base/kustomization.yaml
@@ -1,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: argocd
+
+resources:
+- 'https://raw.githubusercontent.com/argoproj-labs/argocd-notifications/v1.2.0/manifests/install.yaml'
+patchesStrategicMerge:
+- configmap.yaml
--- a/argocd-notifications/base/configmap.yaml
+++ b/argocd-notifications/base/configmap.yaml
@@ -1,0 +1,192 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: argocd-notifications-cm
+data:
+  context:
+    argocdUrl:
+  service.slack: |
+    token: $slack-token
+    icon: ":anchor:"
+  trigger.on-sync-failed: |
+    - when: app.status.operationState.phase in ['Error', 'Failed']
+      description: Application syncing has failed
+      send: [app-sync-failed]
+  trigger.on-sync-succeeded: |
+    - when: app.status.operationState.phase in ['Succeeded']
+      description: Application syncing has succeeded
+      send: [app-sync-succeeded]
+      oncePer: app.status.sync.revision
+  trigger.on-sync-status-unknown: |
+    - when: app.status.sync.status == 'Unknown'
+      description: Application status is 'Unknown'
+      send: [app-sync-status-unknown]
+  trigger.on-health-degraded: |
+    - when: app.status.health.status == 'Degraded'
+      description: Application has degraded
+      send: [app-health-degraded]
+  trigger.on-sync-running: |
+    - when: app.status.operationState.phase in ['Running']
+      description: Application is being synced
+      send: [app-sync-running]
+      oncePer: app.status.sync.revision
+  template.app-sync-failed: |
+    message: |
+      {{if eq .serviceType "slack"}}:x:{{end}}  The sync operation of application {{.app.metadata.name}} has failed at {{.app.status.operationState.finishedAt}} with the following error: {{.app.status.operationState.message}}
+      Sync operation details are available at: {{.context.argocdUrl}}/applications/{{.app.metadata.name}}?operation=true .
+    slack:
+      attachments: |
+        [{
+          "title": "Git information: {{ .app.metadata.name}}",
+          "color": "#ffffff",
+          "fields": [
+            {
+              "title": "Target Branch: ",
+              "value": "{{.app.spec.source.targetRevision}}",
+              "short": true
+            },
+            {
+              "title": "Repository: ",
+              "value": "{{.app.spec.source.repoURL}}",
+              "short": true
+            },
+            {
+              "title": "Overlay Path: ",
+              "value": "{{.app.spec.source.path}}"
+            },
+            {
+              "title": "Revision: ",
+              "value": "{{.app.status.sync.revision}}"
+            }
+          ]
+        }]
+      notifyBroadcast: true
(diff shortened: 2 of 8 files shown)


argocd-notifications/README.md
```diff
@@ -0,0 +1,81 @@
+# Argocd-Notifications
+
+Docs: https://argocd-notifications.readthedocs.io/en/latest/
+
+The Docs are pretty sparse so I will try to provide all the work arounds and knowledge I gained by learning to deploy this.
+
+Also note that Argocd-notifications are currently deployed seperately, although that will be changed. [Argocd-Notifications has been migrated to argo-cd](https://github.com/argoproj/argo-cd/pull/7744), however is not available in any releases at the moment.
```
harshad16, line 7: Maybe change the wording a little bit.
I feel we should just inform the content which would be useful for the reader. 
and sentence like `I will try ` can be removed . what do you feel ?

argocd-notifications/base/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: argocd
+
+resources:
+- 'https://raw.githubusercontent.com/argoproj-labs/argocd-notifications/v1.2.0/manifests/install.yaml'
+patchesStrategicMerge:
+- 'https://raw.githubusercontent.com/argoproj-labs/argocd-notifications/v1.2.0/catalog/install.yaml'
+- configmap.yaml
```
harshad16, line 9: add a new line to fix the pre-commit issue here.

harshad16 (changes requested):
lgtm
some minor changes, lets try to keep the docs in the 2nd person or 3rd person point of view.

argocd-notifications/README.md
Gregory-Pereira: I agree, I think keeping it in the third person is best. Making the changes.

argocd-notifications/base/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: argocd
+
+resources:
+- 'https://raw.githubusercontent.com/argoproj-labs/argocd-notifications/v1.2.0/manifests/install.yaml'
+patchesStrategicMerge:
+- 'https://raw.githubusercontent.com/argoproj-labs/argocd-notifications/v1.2.0/catalog/install.yaml'
```
HumairAK, line 8: do we need this catalog? It doesn't seem like we are using any of the triggers/templates added here.

argocd-notifications/README.md
```diff
@@ -0,0 +1,81 @@
+# Argocd-Notifications
```
HumairAK, line 1: Let's move this to: https://github.com/operate-first/apps/tree/master/docs/content/argocd-gitops 

And add it to the jupyterbook, instructions here: https://www.operate-first.cloud/apps/CONTRIBUTING.html

HumairAK: The purpose of this doc isn't clear to me starting out, why should I read this doc? What sort of information do we expect users to seek? In my mind the doc should answer this question: 

> Where and How do I add notifications to my ArgoCD apps?

argocd-notifications/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Future plans
+
+Deploying argocd notifications and utilizing the slack integration was meant to help give us visibility into argocd's deployment without actually having to open up the link, but truly was more of a POC than anything. The orrigional goal was to integrate it with a github application which would write the sync status back to the pr that merged the changes that started argocd sync in the first place. However Github apps proved to be more tricky than anticipated, and so the scope was refined to simply deploying argocd notifications with a service that allowed insight into the sync status and errors in a certrallized location.
+
+Feel free to add suggestions and ideas about what else should be done with this project:
+
+  - once argocd-notifications is integrated with our github application, for every pr in an application that would be deployed through argocd, have a check for it that would checkout that temp branch of the code, and build it from argocd to see if it would build correctly. If the build passes then the check passes but if not it would fail, not allowing it to be merged.
```
HumairAK, line 81: Sounds like this should be in an ADR, let's take it out of this doc.

argocd-notifications/README.md
HumairAK: I see that you want to add additional descriptions for Triggers. So maybe separate the doc into 2 parts: 

> How to add notifications
   * step by step instruction on how to trigger a notification for an argocd app in the `apps` repo, what files need to be updated? where? etc. 
   
> References
* describe further if you feel that it's needed argocd-notification concepts like triggers/templates/etc. and you can refer to these in the above section should the need arise.

argocd-notifications/README.md
```diff
@@ -0,0 +1,81 @@
+# Argocd-Notifications
+
+Docs: https://argocd-notifications.readthedocs.io/en/latest/
+
+The Docs are pretty sparse so additional information will be added here.
+
+Also note that Argocd-notifications are currently deployed seperately, although that will be changed. [Argocd-Notifications has been migrated to argo-cd](https://github.com/argoproj/argo-cd/pull/7744), however is not available in any releases at the moment.
+
+## TLDR (summary if lazy)
```
HumairAK, line 9: I would not add informal phrasing like this. Keep the doc formal.

argocd-notifications/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+  More documentation and examples to come, for now see [the docs](https://argocd-notifications.readthedocs.io/en/stable/templates/).
+
+## Notification services / integrations
+
+  [The Docs](https://argocd-notifications.readthedocs.io/en/stable/services/overview/) are the best resource on this topic. Refer to them.
+
+## Deployment
+
+  Currently argocd-notifications is deployed in the argocd namespace, alongside argocd itself.
```
HumairAK, line 73: Don't think we need a section for this, you can mention this in the intro with something like "we deploy argocd-notifications along side argocd etc. etc."

HumairAK (approved):
/lgtm
/approve

PR #1523, opened 2022-01-17
add reloader annotations to trino deployments

dystewart:


HumairAK:
@dystewart can you also add it to the secrets: https://github.com/operate-first/apps/tree/master/kfdefs/overlays/moc/smaug/opf-trino/secrets you'll need to use sops

HumairAK:
can you also do the same for trino stage https://github.com/operate-first/apps/tree/master/odh-manifests/smaug-stage/trino/base

dystewart:
@HumairAK what is the distinction between the trino manifests in the smaug vs smaug-stage dirs here?
[https://github.com/operate-first/apps/tree/master/odh-manifests](https://github.com/operate-first/apps/tree/master/odh-manifests)

HumairAK:
@dystewart ODH deploys these manifests into different namespaces. The ones in `smaug-stage` go to staging namespaces, in trino's case that's `opf-trino-stage`

dystewart:
@HumairAK One more question/sanity check. There is a secret (file-auth-secret.enc.yaml) at [secrets](https://github.com/operate-first/apps/tree/master/kfdefs/overlays/moc/smaug/opf-trino/secrets) that does not appear to be used by either the trino-worker or trino-coordinator deployments. The other secrets in this dir are all in use via env variables by the deployments. Is there something I'm missing or is this secret in fact no being used in these deployments?

HumairAK:
@dystewart it's not being used by a deployment, but I created that to store in the namespace so we were tracking the login credentials for trino/cloudbeaver _somewhere_ so that information was not lost.

HumairAK (approved):
/lgtm
/approve
Awessome

PR #1525, opened 2022-01-17
Add instructions on how to add an ocp group.

HumairAK:
While this portion is mentioned in the onboarding docs, and/or in the notebook. There are situations where teams /users may want to create an OCP group after onboarding, or without also requiring a namespace.


dystewart (commented):
/lgtm

PR #1546, opened 2022-01-21
Add osc-cl1 cluster to osc clusterset.

HumairAK:
This will create a new `manageclusterset` called `osc`, and add the `os-cl1` cluster to it. In the future we will be adding the prod OSC cluster to this cluster set as well. I have left out the provisioner role for the [osc admins group](). 

This will result in the following [issue](https://github.com/operate-first/support/issues/436) -- but based on the comments on the jira we should be able to just create a namespace via gitops to workaround it (we can test this when importing the osc prod cluster).

### DIFF
--- a/acm/overlays/moc/infra/managedclustersetbindings/osc.yaml
+++ b/acm/overlays/moc/infra/managedclustersetbindings/osc.yaml
@@ -1,0 +1,7 @@
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: ManagedClusterSetBinding
+metadata:
+  name: osc
+  namespace: argocd
+spec:
+  clusterSet: osc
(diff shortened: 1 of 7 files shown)


acm/overlays/moc/infra/managedclustersetbindings/osc.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: ManagedClusterSetBinding
+metadata:
+  name: osc
+  namespace: argocd
```
harshad16, line 5: i just noticed the namespace is only added in this manifest.

harshad16 (commented):
LGTM
just one small comment, the namespace is not added to all manifest.
is the one added was intentional?

acm/overlays/moc/infra/managedclustersetbindings/osc.yaml
HumairAK: yep that's intended, the other manifests are cluster-wide.

harshad16 (approved):
lgtm
thanks :100:

4n4nd (approved):
/lgtm

PR #1548, opened 2022-01-24
Grant thoth-devops access to tekton-chains

codificat:
I would like to add our team to access tekton-chains.

If it helps: this would match the existing access to the [opf ci pipelines access](https://github.com/operate-first/apps/blob/9d082291740f45154d58f62b601fd62f062ba172/cluster-scope/base/core/namespaces/opf-ci-pipelines/kustomization.yaml#L3).

Tekton chains are integrated with the pipelines and I believe this access makes sense.


harshad16 (approved):
/lgtm 
thanks :+1:

PR #1554, opened 2022-01-25
Add Tekton-backed Kubeflow Pipelines

anishasthana:
Closes https://github.com/operate-first/support/issues/435

Signed-off-by: Anish Asthana <anishasthana1@gmail.com>

### DIFF
--- a/kfp-tekton/base/kfp-tekton.yaml
+++ b/kfp-tekton/base/kfp-tekton.yaml
@@ -1,0 +1,1504 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    application-crd-id: kubeflow-pipelines
+  name: kubeflow-pipelines-container-builder
+  namespace: kubeflow
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    application-crd-id: kubeflow-pipelines
+  name: kubeflow-pipelines-metadata-writer
+  namespace: kubeflow
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    application-crd-id: kubeflow-pipelines
+  name: kubeflow-pipelines-viewer
+  namespace: kubeflow
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    application-crd-id: kubeflow-pipelines
+  name: metadata-grpc-server
+  namespace: kubeflow
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    application-crd-id: kubeflow-pipelines
+  name: ml-pipeline
+  namespace: kubeflow
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    application-crd-id: kubeflow-pipelines
+  name: ml-pipeline-persistenceagent
+  namespace: kubeflow
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    application-crd-id: kubeflow-pipelines
+  name: ml-pipeline-scheduledworkflow
+  namespace: kubeflow
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    application-crd-id: kubeflow-pipelines
+  name: ml-pipeline-ui
+  namespace: kubeflow
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    application-crd-id: kubeflow-pipelines
+  name: ml-pipeline-viewer-crd-service-account
+  namespace: kubeflow
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    application-crd-id: kubeflow-pipelines
+  name: ml-pipeline-visualizationserver
+  namespace: kubeflow
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    application-crd-id: kubeflow-pipelines
+  name: mysql
+  namespace: kubeflow
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    application-crd-id: kubeflow-pipelines
+  name: pipeline-runner
+  namespace: kubeflow
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  labels:
+    app: kubeflow-pipelines-metadata-writer-role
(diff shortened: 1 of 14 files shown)


anishasthana:
Need to figure out the implications of the required SCC, probably move some of the contents of this PR around to different folders too.

HumairAK:
Okay I'm going through the manifests. So off the bat, all the cluster-scope manifests should go to the `cluster-scope` directory here: https://github.com/operate-first/apps/tree/master/cluster-scope 

See the readme docs in that link on how to structure it. 

You might find this tool usefule for this https://github.com/larsks/halberd Remember you have to pull these into the `smaug` overlay. 

Having said that, there are _a lot_ of questionable manifests here ... in my opinion, at least for what it _seems_ to be asking. And I will try to go through them.

kfp-tekton/base/kfp-tekton.yaml
```diff
@@ -0,0 +1,3172 @@
...
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  labels:
+    app.kubernetes.io/component: webhook
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-pipeline-loops
+    pipeline.tekton.dev/release: devel
+  name: validation.webhook.pipelineloop.custom.tekton.dev
+webhooks:
```
HumairAK: There's no `rules` file defined here...so it seems like...it does nothing ([docs](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-rules))? Same with the `MutatingWebhookConfiguration`. I'm not super familiar with admission webhooks, I think @tumido or @harshad16 might know. 

If it does nothing...we could probably scrap the deployment/service associated with this as well..

HumairAK:
This PR is installing [csi-s3](https://github.com/ctrox/csi-s3/tree/master/deploy/kubernetes): 

> This is a Container Storage Interface (CSI) for S3 (or S3 compatible) storage. This can dynamically allocate buckets and mount them via a fuse mount into any container. 

I don't think we want to just install a new CSI that can mount s3 storage w/o a separate discussion on why we should add it...

Do we need this? Which component needs it? Can we use the s3 we already have?

kfp-tekton/base/kfp-tekton.yaml
```diff
@@ -0,0 +1,3172 @@
...
+data:
+  accessKeyID-base64: bWluaW8=
+  bucket-base64: bWxwaXBlbGluZQ==
+  endpoint-base64: aHR0cDovL21pbmlvLXNlcnZpY2Uua3ViZWZsb3c6OTAwMA==
+  secretAccessKey-bas64: bWluaW8xMjM=
+kind: ConfigMap
+metadata:
+  labels:
+    application-crd-id: kubeflow-pipelines
+  name: csi-s3-parameters
```
HumairAK: doesn't look like configmap `csi-s3-parameters` is used anywhere

kfp-tekton/base/kfp-tekton.yaml
```diff
@@ -0,0 +1,3172 @@
...
+          path: /var/data/kubelet/plugins_registry
+          type: Directory
+        name: registration-dir
+      - hostPath:
+          path: /dev
+          type: Directory
+        name: dev-dir
+---
+apiVersion: storage.k8s.io/v1
+kind: CSIDriver
```
HumairAK: Looks like this is also part of csi-s3 (from the earlier comment), but it's not explicitly listed [here](https://github.com/ctrox/csi-s3/tree/master/deploy), I think we should try making kfp-tekton work without this.

HumairAK:
@anishasthana I also noticed in your demo deployed on the dev cluster that the `cache-deployer` and `cache-server` are not running successfully either. Is this something we need? Maybe we can try taking these out to reduce some of this clutter.

I skimmed through the manifests/configfiles/secrets and I also don't see any reference to the cache-deployer service.

kfp-tekton/base/kfp-tekton.yaml
```diff
@@ -0,0 +1,3172 @@
...
+  namespace: kubeflow
+---
+apiVersion: v1
+data:
+  defaultPipelineRoot: ""
+kind: ConfigMap
+metadata:
+  labels:
+    application-crd-id: kubeflow-pipelines
+  name: kfp-launcher
```
HumairAK: don't see this being referenced anywhere

kfp-tekton/base/kfp-tekton.yaml
```diff
@@ -0,0 +1,3172 @@
...
+apiVersion: v1
+data:
+  METADATA_GRPC_SERVICE_HOST: metadata-grpc-service
+  METADATA_GRPC_SERVICE_PORT: "8080"
+kind: ConfigMap
+metadata:
+  labels:
+    application-crd-id: kubeflow-pipelines
+    component: metadata-grpc-server
+  name: metadata-grpc-configmap
```
HumairAK: don't see this being referenced anywhere

kfp-tekton/base/kfp-tekton.yaml
```diff
@@ -0,0 +1,3172 @@
...
+  region: us-south
+  secretKey: minio123
+  serviceEndpoint: http://minio-service.kubeflow:9000
+  serviceInstanceID: 'crn:v1:bluemix:public:cloud-object-storage:global:a/ID-dummy-1231231231-123abcdefgh:dummy-values::'
+kind: ConfigMap
+metadata:
+  labels:
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-pipelines-loops
+  name: object-store-config
```
HumairAK: don't see this being referenced anywhere

HumairAK:
So there's a `minio` deployment also included, we could probably trim this down and just use an OBC to get our bucket. The following deployments seem to require the s3 creds: 

```
apiVersion: apps/v1
kind: Deployment
metadata:
    name: ml-pipeline-ui
    namespace: kubeflow
---

apiVersion: apps/v1
kind: Deployment
metadata:
    name: ml-pipeline
    namespace: kubeflow
```

anishasthana:
We disabled creds-init by setting the  `disable-creds-init` field in the feature-flag configmap in the openshift-pipelines namespace to True. This fixed another error.
Context: https://github.com/shipwright-io/build/issues/679#issuecomment-829259173

anishasthana:
`track_artifacts: "false"` in the kfp-tekton config map

HumairAK (approved):
/lgtm
/approve

PR #1556, opened 2022-01-26
add DSL management for trino rules.json files

erikerlandson:
This PR adds a DSL management layer to the `rules.json` files for trino deployments.

The motivation is that `rules.json` rules are:
- difficult to read
- conducive to errors, since one has to scan down all rules to identify first-match in all relevant circumstances
- easy to break, because adding a rule in one place can shadow lower rules if one is not careful
- do not support human readable comments

I implemented a simplified DSL-like yaml format to try and make these configurations easier:
https://github.com/os-climate/osc-trino-acl-dsl

The basic concept is that one creates configurations in this DSL format, and it is translated into a `rules.json` file.

The DSL is declarative - one does NOT have to figure out which entries apply "first": they are order independent.
It can also be written in YAML - and so supports white space and comments, to increase readability and maintainability.

One trade-off with this approach is that proper consistency between this DSL file and the corresponding rules.json file must
be maintained. I have addressed this by implementing a pre-commit check that verifies the rules.json file being committed
is the result of running the translator on the corresponding DSL file.  If it detects a difference it will fail the commit.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,3 +26,16 @@ repos:
         files: \.(yaml|yml)$
         types: [file, yaml]
         entry: yamllint --strict -c yamllint-config.yaml
+
+  - repo: https://github.com/os-climate/osc-trino-acl-dsl
+    rev: v0.2.3
+    hooks:
+      # manage rules.json files using a DSL for trino ACL
+      # this check enforces that rules.json is consistent with dsl file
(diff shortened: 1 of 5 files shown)


erikerlandson:
cc @rimolive @caldeirav @redmikhail @MichaelTiemannOSC

HumairAK:
/lgtm
/approve

Awesome! Thanks @erikerlandson !

PR #1559, opened 2022-01-26
feat(onboarding): Add team varangian

KPostOffice:


cluster-scope/base/core/namespaces/varangian-test/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+    - namespace.yaml
+components:
+    - ../../../../components/project-admin-rolebindings/varangian
+    - ../../../../components/limitranges/default
```
HumairAK, line 7: This seems to be missing a resource quota, from the onboarding request it looks like you want `small` so please add the following line under `components`: 

```
- ../../../../components/resourcequotas/small 
```

This is due to the `CUSTOM_QUOTA=True` value in the notebook, which should probably be false by default, I'll create an issue for this.

HumairAK: I went ahead and added this for you

PR #1563, opened 2022-01-29
feat: Add temporary a devconf.cz 2022 namespace

tumido:
Creating a temporary namespace - will be cleaned up/reverted in a day.


pacospace (approved):
lgtm! 💯

PR #1570, opened 2022-01-31
Adding the documentation for resource bundling

mubariskhan96:
Please suggest any improvements for the documentation. Thanks!

### DIFF
--- a/cluster-scope/README.md
+++ b/cluster-scope/README.md
@@ -15,6 +15,9 @@ cluster-scope
 │           └── <resource_metadata_name>
 │               ├── kustomization.yaml
 │               └── <resource_kind_singular>.yaml
+├── bundles
+│   └── <operator>
+│       ├── kustomization.yaml
 ├── components
 │   ├── <descriptive_component_name>
 │   └── kustomization.yaml
@@ -33,6 +36,11 @@ The directories and files in this structure fulfill the following purpose:
 - `<resource_metadata_name>` : The name of the resource deployed at cluster scope. It must match the resource `.metadata.name` in `<resource_kind_singular>.yaml`. Since this resource is deployed at the cluster scope, its name has to be unique.
 - `kustomization.yaml` : Kustomization file including a single resource, which is a sibling file within the folder. Can pull in `components` for additional bootstrapping.
 - `<resource_kind_singular>.yaml` : **Single** resource manifest with a `.spec` that can be deployed to all the clusters.
+
+#### bundles/
+Bundled manifests are manifests that are dependent on each other and should be deployed together on the same cluster. You should create a bundle that references all these manifests from a single kustomization.yaml file following the structure above. This bundle can then be easily imported to the target cluster's by adding it to `overlays/<env>/<cluster_name>/kustomization.yaml`. Bundle related resources by including (or removing) a single line in the relevant overlay.
+  - `<bundle-name>` : The resource that is being bundled.
+  - `kustomization.yaml` : The list of the resources that belong to the operator.
 
 #### components/
 Bootstrapping components pulled into `base` (this is different to the standard usage in Kustomize!).


cluster-scope/README.md
```diff
@@ -34,6 +37,11 @@ The directories and files in this structure fulfill the following purpose:
 - `kustomization.yaml` : Kustomization file including a single resource, which is a sibling file within the folder. Can pull in `components` for additional bootstrapping.
 - `<resource_kind_singular>.yaml` : **Single** resource manifest with a `.spec` that can be deployed to all the clusters.
 
+#### bundles/
+Resource Bundling was introduced in this [PR](https://github.com/operate-first/apps/pull/1552). And is discussed more extensively [here](https://github.com/operate-first/SRE/issues/442)
```
HumairAK, line 41: I think we should just explain briefly when/how users should bundle resources instead of linking to threads (otherwise they would have to read entire discussions to understand).

mubariskhan96: done

cluster-scope/README.md
```diff
@@ -34,6 +37,11 @@ The directories and files in this structure fulfill the following purpose:
 - `kustomization.yaml` : Kustomization file including a single resource, which is a sibling file within the folder. Can pull in `components` for additional bootstrapping.
 - `<resource_kind_singular>.yaml` : **Single** resource manifest with a `.spec` that can be deployed to all the clusters.
 
+#### bundles/
+Bundle related resources by including ( or removing) a single line in the relevant overlay.
+  - `<operator>` : The resource that is being bundled.
+  - `kustomization.yaml` : The list of the resources that belong to the operator.
+
```
HumairAK, line 44: Let's change `<operator>` to `<bundle-name>` since it doesn't have to be an operator.

We should further clarify when/what to bundle. For instance maybe add something like the following as well: 

> Bundled manifests are manifests that are dependent on each other and should be deployed together on the same cluster. You should create a `bundle` that references all these manifests from a single `kustomization.yaml` file following the structure above. This bundle can then be easily imported to the target cluster's by adding it to `overlays/<env>/<cluster_name>/kustomization.yaml`.

cluster-scope/README.md
```diff
@@ -34,6 +37,11 @@ The directories and files in this structure fulfill the following purpose:
 - `kustomization.yaml` : Kustomization file including a single resource, which is a sibling file within the folder. Can pull in `components` for additional bootstrapping.
 - `<resource_kind_singular>.yaml` : **Single** resource manifest with a `.spec` that can be deployed to all the clusters.
 
+#### bundles/
+Bundled manifests are manifests that are dependent on each other and should be deployed together on the same cluster. You should create a bundle that references all these manifests from a single kustomization.yaml file following the structure above. This bundle can then be easily imported to the target cluster's by adding it to overlays/<env>/<cluster_name>/kustomization.yaml. Bundle related resources by including ( or removing) a single line in the relevant overlay.
```
HumairAK, line 41: can you add code formatting to: overlays/<env>/<cluster_name>/kustomization.yaml 
e.g. `overlays/<env>/<cluster_name>/kustomization.yaml` 

`( or removing)` change to `(or removing)`

mubariskhan96: ok! will do it right away

mubariskhan96: done!

larsks (changes requested):
@mubariskhan96 the changes look good, but referring back to the presentation on writing good commit messages (http://oddbit.com/code-and-communication/#18 and the follow slides), the commit included in this PR has the subject "code formatting for file path and removing extra space from inside the paranthesis". I'd like to see that updated to at least reflect the pr description.

mubariskhan96:
> @mubariskhan96 the changes look good, but referring back to the presentation on writing good commit messages (http://oddbit.com/code-and-communication/#18 and the follow slides), the commit included in this PR has the subject "code formatting for file path and removing extra space from inside the paranthesis". I'd like to see that updated to at least reflect the pr description.

@larsks yeah, I just realized that I had amended my commit and only the final amended messages stuck around in the PR. 

I have updated it now!

larsks:
I've made an edit to the commit message to bring it in line with the conventions we had previously discussed; I'd like to see messages more like that in fhe future. Otherwise, this PR looks fine!

PR #1571, opened 2022-02-01
Configure Smaug for valid API certificate

margarethaley:
By using a cert-manager managed certificate, tools will be able to connect to the OpenShift API without having to trust the OpenShift CA.

With the default API server certificate, clients outside of the cluster cannot verify the API server’s certificate by default. This results in certificate errors when external tools attempt to connect. This PR adds a cert-manager managed certificate that solves this problem.

Resolves https://github.com/operate-first/SRE/issues/441


cert-manager/overlays/moc/smaug/api/openshift-api-issuer.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: ingress-letsencrypt
```
larsks, line 4: Because this issuer isn't for the `openshift-ingress` namespace we should name it something other than `ingress-letsencrypt`. Perhaps `api-letsencrypt`? You'll need to update the corresponding information in the `Certificate` resource.

cert-manager/overlays/moc/smaug/api/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: openshift-config
+resources:
+  - api-cert.yaml
+  - openshift-api-issuer.yaml
+  - ../../common
```
larsks, line 7: We don't want to include `../../common` here, because that will create the `openshift-ingress` issuer, which is already created by the `smaug/ingress` directory. What problem needs to happen here is:

- Split up `common` into `secrets`, `ingress-issuer`, and `api-issuer`.
- Make the `balrog` overlay look like `smaug` and `infra` in terms of layout (e.g., have an `ingress` directory with the ingress configuration)
- Where we're creating an ingress issuer for `operate-first.cloud`, include `../../common/secrets` and `../../common/ingress-issuer`
- Where we're creating an api issuer for `operate-first.cloud` (currently just Smaug), include `../../common/api-issuer`.

Does that make sense? The goal is basically to (a) make everything look as similar as possible and (b) avoid creating objects in the wrong place.

larsks (changes requested):

larsks (changes requested):
This is getting closer! There are still a few things we need to deal with.

1. We want to aim for as much parallel construction as possible -- that is, the api certificate should be implemented in a manner that is as close as possible to the ingress certificate.

    Since the `cert-manager` directory was originally structured just for the ingress certificate, this means we need reorganize several parts of the repository, including the `base` directory. Specifically, we should include the base version of api certificate in `base` just like we're currently doing for the ingress certificate.

    Another implication of this goal is that we should use existing naming conventions, so if the existing ingress issuer is named `ingress-letsencrypt-production`, we should probably name the corresponding api issuer `api-letsencrypt-production` and follow the same order of elements in the manifest (and so forth for any other parallel resources).

2. The existing google cloud dns secret has an explicit namespace of `openshift-ingress`. We're going to need the same secret in the `openshift-config` namespace (which is where we're creating the api issuer and certificate). The easiest way to handle that is probably to remove the namespace from the secret (`overlays/moc/common/secrets/clouddns-dns01-na-svc-acct.enc.yaml`) and set it in the various kustomzation files (anything that includes the secret).

3. When you moved the contents of `overlays/balrog/` to `overlays/balrog/ingress`, we need to add `overlays/balrod/kustomization.yaml` for that overlay to build correctly.

Let me know if I can clarify any of these comments, or if you'd like to set up a call to walk through things in more detail!

margarethaley:
Okay @larsks, sounds great! I do have a question about the dns secret, referencing this comment:

> The existing google cloud dns secret has an explicit namespace of openshift-ingress. We're going to need the same secret in the openshift-config namespace ...

When working on the pr, I was looking into / basically modeling off of how it's done in `infra`. `overlays/moc/infra/api` and `overlays/moc/infra/ingress` both reference `overlays/moc/infra/secrets`, which contains a secret only in the `openshift-ingress` namespace. I am confused as to how / why this works in `infra`, when we need the secret in both `openshift-ingress` and `openshift-config` namespaces for the other clusters.

larsks:
> I am confused as to how / why this works in infra, when we need the secret in both openshift-ingress and openshift-config namespaces for the other clusters.

If you take a look at `overlays/moc/infra/api`, you'll note there's an explicit `namespace:` set there:

```
kind: Kustomization
namespace: openshift-config
resources:
  - api-cert.yaml
  - openshift-api-issuer.yaml
  - ../secrets
```

If you run `kustomize build` in that directory, the resulting `Secret` resource looks like:

```
apiVersion: v1
kind: Secret
metadata:
  name: aws-route53-credentials
  namespace: openshift-config
[...]
```

The namespace set in `kustomization.yaml` overrides the namespace in `../secrets/route53-credentials.enc.yaml`.

larsks:
W/r/t to my comments about changes to the `base` directory, I was thinking about a layout something like this:

```
.
├── base
│   ├── api-certificate
│   └── ingress-certificate
└── overlays
    ├── moc
    │   ├── balrog
    │   │   └── ingress
    │   ├── common
    │   │   ├── api-issuer
    │   │   ├── ingress-issuer
    │   │   └── secrets
    │   ├── infra
    │   │   ├── api
    │   │   ├── ingress
    │   │   └── secrets
    │   └── smaug
    │       ├── api
    │       └── ingress
    └── route53
        └── us-east-1
            └── osc-cl1
```

margarethaley:
@larsks I've implemented those changes (for the most part) -

The only issue I ran into was when I tried to remove the namespace from the secret - after doing this, `kustomize build` resulted 
in an error along the lines of 
> trouble decrypting file Failed to verify data integrity ...

But when I replaced it, and then just added the appropriate namespace to the relevant kustomization files, everything seemed to build correctly.

I also wanted to double check that the file structure is as it should be. Specifically, I wasn't sure if you wanted the `api certificate` and `ingress certificate` under `base` to be their own folders, or just the certificate files themselves. I made them folders, but can change to just the files if that's preferable.

larsks:
We're going to need to update `overlays/route53/us-east-1/osc-cl1/kustomization.yaml` probably to include `base/ingress-certificate` instead of just `base`, since we've moved things around (and we can probably remove `base/kustomization.yaml`). Otherwise, this looks good!

margarethaley:
@larsks I made those last two changes!

larsks (commented):
/lgtm

larsks (commented):
/approve

larsks:
@margarethaley Well, I missed something in the review. Our goal was to make the api issuer look as much like the ingress issuer as possible, but I didn't notice until after merging the change that you used the route53 issuer (e.g., `overlays/moc/infra/openshift-api-issuer.yaml`) as a model, rather than the google cloud issuer (e.g., `overlays/moc/common/ingress-issuer`).

That's my bad; I should have caught that. I've submitted #1597 with the fix. Feel free to take a look and let me know if I've missed anything!

margarethaley:
@larsks thank you for correcting that. I believe I initially thought I was modeling off of the implementation `infra`, so I replicated that issuer rather than making it similar to ingress. My apologies for the mistake!

PR #1580, opened 2022-02-03
feat(infra): Grant oglok cluster admin

tumido:
@oglok needs a way to manage what ACM addons are installed to his microshift clusters and experiment with it a bit. Since `KlusterletAddonConfig` is a cluster scoped resource kind, he can't touch them. Can we grant him cluster admin access for now?

### DIFF
--- a/cluster-scope/overlays/prod/moc/infra/groups/cluster-admins.yaml
+++ b/cluster-scope/overlays/prod/moc/infra/groups/cluster-admins.yaml
@@ -12,3 +12,4 @@ users:
   - naved001
   - rob-baron
   - tumido
+  - oglok


larsks:
Seems as if it would be pretty easy to create a clusterrole with appropriate access to klusterlet resources, rather than granting full admin rights.

larsks:
I think that would look like https://github.com/operate-first/apps/pull/1581

HumairAK:
/hold

yeah let's go with #1581 instead @tumido / @oglok  let us know if that works for you

larsks:
Closing in favor of #1581

tumido:
Thank you, #1581 is definitely better!

PR #1583, opened 2022-02-04
added carlosGimeno to the sre group

carlosGimeno:
Adding my GH ID to the SRE group

### DIFF
--- a/cluster-scope/base/user.openshift.io/groups/sre/group.yaml
+++ b/cluster-scope/base/user.openshift.io/groups/sre/group.yaml
@@ -17,3 +17,4 @@ users:
   - bryanmontalvan
   - suppathak
   - mubariskhan96
+  - carlosgimeno


cluster-scope/base/user.openshift.io/groups/sre/group.yaml
```diff
@@ -17,3 +17,4 @@ users:
   - bryanmontalvan
   - suppathak
   - mubariskhan96
+  - carlosGimeno
```
4n4nd, line 20: can you make this all lowercase?

4n4nd (changes requested):

cluster-scope/base/user.openshift.io/groups/sre/group.yaml
carlosGimeno: done !

4n4nd (approved):
/lgtm

Issue #1511, opened 2022-01-11
Trouble accessing the grafana dashboard


harshad16:
One of the developer of Thoth group is not able to login into grafana dashboard
user github id: Gkrumbach07

already added in the thoth group
https://github.com/operate-first/apps/commit/0b31c8b96f02cfeadfc4d3769b87000c30f9660e

however when the user tries to login always get the following issue:
![image](https://user-images.githubusercontent.com/14028058/148995152-26ca44b6-fb20-47f4-bd5e-d5ddcff98daa.png)

could be due to the capital letter in the id.

Gkrumbach07:
Has this issue been resolved? Is there something I need to do on my end. Thanks

Gregory-Pereira:
@Gkrumbach07 once #1593 gets merged try logging into [grafana](https://grafana.operate-first.cloud/login) and let me know if its fixed.

Gkrumbach07:
> @Gkrumbach07 once https://github.com/operate-first/apps/pull/1593 gets merged try logging into [grafana](https://grafana.operate-first.cloud/login) and let me know if its fixed.

That fixed it. Thanks!

PR #1593, opened 2022-02-08
removing capitals from thoth group

Gregory-Pereira:
In Response to [Trouble accessing the grafana dashboard](https://github.com/operate-first/apps/issues/1511). Once this gets merged we can test and see if this works. Another possible workaround that could be tried, if this is not an issue, is adding the [`fde` group](https://github.com/operate-first/apps/blob/master/cluster-scope/base/user.openshift.io/groups/fde/group.yaml) to the [`grafana-oauth.yaml` file](https://github.com/operate-first/apps/blob/master/grafana/overlays/moc/smaug/grafana-oauth.yaml#L38-39).

### DIFF
--- a/cluster-scope/base/user.openshift.io/groups/thoth/group.yaml
+++ b/cluster-scope/base/user.openshift.io/groups/thoth/group.yaml
@@ -12,4 +12,4 @@ users:
   - mayacostantini
   - xtuchyna
   - gregory-pereira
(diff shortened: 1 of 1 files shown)


HumairAK (approved):
/lgtm
/approve

Gregory-Pereira:
Improper URL for the issue. Linking to #1511

PR #1594, opened 2022-02-08
Switch strimzi management to argocd on smaug.

HumairAK:
Let's use the same method of strimzi deployment like we do on balrog, and use olm instead.


harshad16 (approved):
lgtm
thanks :100:

harshad16 (approved):
/lgtm
thanks :100:

PR #1597, opened 2022-02-08
Previous commit used the wrong issue for the api certificate

larsks:
PR #1571 was mistakenly using the route53 issuer for the smaug api
certificate. This corrects it to use the google cloud issuer.


HumairAK (approved):
/lgtm
/approve

PR #1604, opened 2022-02-10
Add parameter to argocd vault application.

revit13:
This PR adds parameter to argocd vault application.
By disabling vault.server.authDelegator the clusterRoleBinding `vault-server-binding` is not created by this helm deployment which currently causes the deployment to fail.
Moreover, `vault-server-binding` resource already [exists ](https://github.com/operate-first/apps/tree/master/cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/vault-server-binding)in the cluster.


HumairAK (approved):
/lgtm
/approve

PR #1605, opened 2022-02-10
Remove explicit version from the ODH operator subscription

codificat:
The current version of ODH operator is 1.1.2.

Trying to subscribe to ODH using the current `Subscription` from the repo results in a non-installed operator, due to a failing condition:

```
Message:               constraints not satisfiable:
- bundle community-operators/openshift-marketplace/stable/opendatahub-operator.v1.1.0 is deprecated
- subscription opendatahub-operator requires community-operators/openshift-marketplace/stable/opendatahub-operator.v1.1.0
```

This removes the explicit constraint on the CSV version to avoid that situation


codificat:
The change here has now been implemented already in https://github.com/operate-first/apps/pull/1658

/close

PR #1606, opened 2022-02-10
Make sparkoperator in jh ocp 4.9 compatible.

HumairAK:
While we wait for https://github.com/opendatahub-io/odh-manifests/pull/509 to merge. 

Could also help illustrate the PR is in working order. 

Note: this slightly violates the "crds in `cluster-scope/base` rule, but that's because the ODH operator is deploying these manifests, and they are effectively part of ODH platform (not something argocd manages/deploys).


Gregory-Pereira (approved):
/Igtm :+1:

PR #1609, opened 2022-02-10
Installing volsync using kustomize

larsks:
Now that our very exciting call is over, install volsync using our standard practice of exploding the helm chart into a set of kustomize manifests.


argocd/overlays/moc-infra/applications/envs/moc/smaug/cluster-management/volsync.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: volsync
+  namespace: argocd
+spec:
+  project: cluster-management
+  source:
+    path: volsync/overlays/moc/smaug
+    repoURL: https://github.com/operate-first/apps.git
+    targetRevision: HEAD
+  destination:
```
HumairAK, line 12: can we turn on auto sync? i.e.: 

```
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
    syncOptions:
      - Validate=false
```

HumairAK:
other than above lgtm

argocd/overlays/moc-infra/applications/envs/moc/smaug/cluster-management/volsync.yaml
larsks: Oops, yeah. All set.

argocd/overlays/moc-infra/applications/envs/moc/smaug/cluster-management/volsync.yaml
```diff
@@ -0,0 +1,20 @@
...
+    name: smaug
+    namespace: volsync-system
```
HumairAK, line 14: we should also add this namespace declaratively in cluster-scope/base

larsks:
(Your comment re: namespaces reminded me to add the namespace to the resource bundle.)

cluster-scope/bundles/volsync/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../../base/core/namespaces/volsync-system
+  - ../../base/apiextensions.k8s.io/customresourcedefinitions/replicationdestinations.volsync.backube
+  - ../../base/apiextensions.k8s.io/customresourcedefinitions/replicationsources.volsync.backube
+  - ../../base/rbac.authorization.k8s.io/clusterrolebindings/volsync-manager
+  - ../../base/rbac.authorization.k8s.io/clusterrolebindings/volsync-proxy
+  - ../../base/rbac.authorization.k8s.io/clusterroles/volsync-manager
+  - ../../base/rbac.authorization.k8s.io/clusterroles/volsync-metrics-reader
```
HumairAK, line 10: is the scc being imported else where? (I don't see it)

Also what about the namespace, is that something that would go in the bundle (I would think so)?

larsks: > is the scc being imported else where? (I don't see it)

Whoops, yeah, missed that.

> Also what about the namespace, is that something that would go in the bundle (I would think so)?

Yeah, added that while you were reviewing :)

larsks: Updated.

PR #1610, opened 2022-02-10
Refactor kfp-tekton

rimolive:
- Split kfp-tekton.yaml in multiple files for organization
- Extract the image tag information in kustomization.yaml


kfp-tekton/base/roles/kubeflow-pipelines-metadata-writer-role.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  labels:
+    app: kubeflow-pipelines-metadata-writer-role
+    application-crd-id: kubeflow-pipelines
+  name: kubeflow-pipelines-metadata-writer-role
+  namespace: kubeflow
```
4n4nd, line 8: let's not hard-code the namespace and set it in the kustomization?

4n4nd: same goes for any common labels in the manifests

4n4nd (changes requested):

4n4nd (approved):
/lgtm

PR #1611, opened 2022-02-10
Remove starting csv from odhseldon subscription in odh.

HumairAK:
Similar to: https://github.com/operate-first/apps/pull/1606 

This will make seldon compatible with ocp 4.9

We can remove this addition once either of these prs is merged upstream: 
https://github.com/opendatahub-io/odh-manifests/pull/522
https://github.com/opendatahub-io/odh-manifests/pull/521


4n4nd (approved):
/lgtm

PR #1612, opened 2022-02-10
Fix volsync smaug overlay path

larsks:
The smaug overlay for the volsync app didn't match the expectations of
the argocd application.


HumairAK (approved):
/lgtm
/approve

PR #1618, opened 2022-02-11
Add manifests to manage OCS odh-cl2 cluster infrastructure parameters

redmikhail:


cluster-scope/overlays/prod/osc/osc-cl2/machinesets/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
```
HumairAK, line 4: what about `machineset-notebook-gpu-g.yaml` ?

HumairAK: From @redmikhail 
> yeah , it is intentional . I am not ready activate it yet .

PR #1623, opened 2022-02-14
Creating Openshift group and adding a user to the group

mubariskhan96:


larsks:
What is the purpose of the "openshift" group? The name seems sufficiently generic that it's going to lead to confusion.

mubariskhan96:
It is going to be used for testing the alerts in Operate First. For my specific case, I will be testing alerts for Jupyter Hub

4n4nd (approved):
/lgtm

Issue #1465, opened 2021-12-17
Move Observatorium metrics/log storage to Ceph 


4n4nd:
Rn buckets are hosted on MOC NFS server which is not being maintained anymore.

- [x] Create 2 Object Bucket Claims in the `opf-observatorium` namespace (https://github.com/operate-first/apps/pull/1631)
- [ ] Migrate metrics and log data from current buckets to the new buckets
- [ ] Update Observatorium credentials to use the new buckets

larsks:
@bryanmontalvan note that instead of handling this migration externally, as we've discussed in chat, you could alternatively submit a pod to the cluster that would perform the migration internally. The pod would have access to all the necessary secrets (either exposed as environment variables or mounted as files on the filesystem).

I think that's probably the right way to handle this.

HumairAK:
stale, closing.

PR #1631, opened 2022-02-15
Add two OBCs for Observatorium storage move to ceph

bryanmontalvan:
## Description:
Currently our buckets are hosted on the MOC NFS server which is not being maintained anymore.

The goal is to create two Object Bucket Claims in the `opf-observatorium` namespace, and migrate the metrics and logs from our current buckets into the new ones

## Issue:
Related [#1465 ](https://github.com/operate-first/apps/issues/1465)

### Note:
Feel free to request changes!
- Name of manifest filename is subject to change along with bucketName

### DIFF
--- a/observatorium/overlays/moc/smaug/thanos/obcs/kustomization.yaml
+++ b/observatorium/overlays/moc/smaug/thanos/obcs/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- opf-observatorium-metrics.yaml
--- a/observatorium/overlays/moc/smaug/loki/kustomization.yaml
+++ b/observatorium/overlays/moc/smaug/loki/kustomization.yaml
@@ -8,6 +8,7 @@ commonLabels:
 resources:
   - ../../../../base/instance/loki
   - servicemonitors
+  - obcs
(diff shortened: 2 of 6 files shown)


observatorium/overlays/moc/smaug/thanos/obcs/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: opf-observatorium
```
4n4nd, line 3: don't need a namespace here if the main kustomization has a namespace set

observatorium/overlays/moc/smaug/thanos/obcs/observatorium-bucket.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: objectbucket.io/v1alpha1
+kind: ObjectBucketClaim
+metadata:
+  name: observatorium-obc
```
4n4nd, line 4: we need 2 different OBCs
1 for opf-observatorium-logs
1 for opf-observatorium-metrics

4n4nd (changes requested):

4n4nd (approved):
/lgtm

PR #1632, opened 2022-02-15
Remove manual spark deployment for jh prod in smaug.

HumairAK:
And only add it to stage until it's confirmed working.


4n4nd (approved):
/lgtm

PR #1634, opened 2022-02-16
Giving Grafana Admin permissions to opf-alerting group

mubariskhan96:
Giving admin permissions so the opf-alerting group can have access to the alerting metrics and the jupyterhub namespace to be able to test the alerts


4n4nd (approved):
/lgtm

PR #1644, opened 2022-02-18
Add subscription to snapscheduler operator

margarethaley:
Related to https://github.com/operate-first/operations/issues/431

In order to use snapscheduler to take snapshots of the Nooba db, we first need to install snapscheduler. This pr subscribes to the snapscheduler operator and pulls that resource into the smaug cluster.

cc: @larsks


Gregory-Pereira (approved):
/Igtm :+1:

PR #1650, opened 2022-02-22
Corrected namespace for snapscheduler

margarethaley:
The previous PR https://github.com/operate-first/apps/pull/1644 had the namespace set as snapscheduler-system, which doesn't exist. This PR corrects the namespace to openshift-operators in accordance with other operators installed in the cluster.

cc: @larsks


4n4nd (approved):
/lgtm

PR #1651, opened 2022-02-22
Add snapshotschedule to take daily snapshots of noobaa db

margarethaley:
This PR creates a snapshotschedule in smaug that will take daily snapshots of the noobaa database PVC within a 1-week retention time. 

Resolves https://github.com/operate-first/operations/issues/431

### DIFF
--- a/cluster-scope/overlays/prod/moc/smaug/snapshotschedules/schedule.yaml
+++ b/cluster-scope/overlays/prod/moc/smaug/snapshotschedules/schedule.yaml
@@ -1,0 +1,13 @@
+apiVersion: snapscheduler.backube/v1
+kind: SnapshotSchedule
+metadata:
+    name: daily
+    namespace: openshift-storage
+spec:
+    claimSelector:
+        matchLabels:
+            noobaa-db: postgres
+    retention:
+        maxCount: 7
+    # Note that schedules are in the UTC timezone
+    schedule: “0 0 * * *”
--- a/cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
+++ b/cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
@@ -132,3 +132,4 @@ resources:
 - ingresscontrollers/default.yaml
 - nodenetworkconfigurationpolicies/vlan-210-nfs.yaml
 - nodenetworkconfigurationpolicies/vlan-211-nese.yaml
+- snapshotschedules/schedule.yaml


larsks:
It would be great if you could update the commit message to include the same information as the pr description. This preserve the "what is this for?" information in the commit history.

margarethaley:
@larsks absolutely - just made the change!

cluster-scope/overlays/prod/moc/smaug/snapshotschedules/schedule.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+    name: daily
+    namespace: openshift-storage
+spec:
+    claimSelector:
+        matchLabels:
+            noobaa-db: postgres
+    retention:
+        expires: “168h”
+    schedule: “0 0 * * *”
```
larsks, line 12: Note that despite appearances, this will not run at midnight unless you happen to live in the UK. From [the documentation](https://backube.github.io/snapscheduler/usage.html):

> Dates and times specified in the cronspec are relative to the UTC timezone (i.e., a schedule of "0 5 * * *" will create a snapshot once per day at 5:00 AM UTC).

I don't think we need to change it, but it might worth a comment if you're changing something else (`# Note that schedules are in the UTC timezone`).

larsks (commented):
@margarethaley thanks for the commit message update! Although when I said "the same information as the pr description", I meant that literally: I thought the pr description provided a good summary, and it includes a link to the relevant issue. The updated message is fine as well, although it should ideally be wrapped at 75 columns or so.

I've made a few minor comments here, but I don't think any of them are critical. I'm going to hold off on approving this until tomorrow morning sometime in case you're feeling inspired, but I'll merge it tomorrow regardless of whether you update it or not.

cluster-scope/overlays/prod/moc/smaug/snapshotschedules/schedule.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: SnapshotSchedule
+metadata:
+    name: daily
+    namespace: openshift-storage
+spec:
+    claimSelector:
+        matchLabels:
+            noobaa-db: postgres
+    retention:
+        expires: “168h”
```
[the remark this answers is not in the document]
margarethaley: Yeah, I agree on this - count is more explicit

cluster-scope/overlays/prod/moc/smaug/snapshotschedules/schedule.yaml
margarethaley: Totally missed the timezone detail! I'll add in a comment.

margarethaley:
@larsks updated again - thanks for the improvements!

PR #1655, opened 2022-02-23
Add identity mgmt operator to infra cluster.

HumairAK:
Related: https://github.com/operate-first/operations/issues/432

This change will deploy the identity mgmt operator (identitatem) to the
Infra cluster. Identitatem allows us to manage authentication via ACM.
This is in contrast to our current implementation of using Keycloak.
This commit does not affect existing OAUTHs for any clusters yet, but
lays the groundwork so that enabling Identitatem to manage any clusters
managed by ACM is as simple as adding required label to a ManagedCluster.

With this commit you will find an Authrealm that is configured to
authenticate any enrolled manageclusters via the Operate First github
org.

For reference docs: 
- https://identitatem.github.io/idp-mgmt-docs-upstream/install_connected.html#installing-while-connected-online
- https://identitatem.github.io/idp-mgmt-docs-upstream/quick_start.html#getting-started

### DIFF
--- a/acm/overlays/moc/infra/identitatem/authrealms/kustomization.yaml
+++ b/acm/overlays/moc/infra/identitatem/authrealms/kustomization.yaml
@@ -1,7 +1,5 @@
 ---
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-
 resources:
-  - argocd-managed-clusters.yaml
-  - primary-auth.yaml
+  - primary.yaml
--- a/acm/overlays/moc/infra/identitatem/authrealms/primary.yaml
+++ b/acm/overlays/moc/infra/identitatem/authrealms/primary.yaml
@@ -1,0 +1,20 @@
+apiVersion: identityconfig.identitatem.io/v1alpha1
+kind: AuthRealm
+metadata:
+  name: primary
+  namespace: idp-mgmt-config
+spec:
+  identityProviders:
+    - github:
+        clientID: b222cb4ccc33667e28bd
+        clientSecret:
+          name: identitatem-github-client-secret
+        organizations:
+          - operate-first
+      mappingMethod: claim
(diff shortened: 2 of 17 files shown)


HumairAK:
> I would love to see our commit messages include more information about the reasons we're making changes. 

I figured the PR referencing the issue would be enough (or maybe mentioning said issue in the commit message), but I suppose I could also add a more detailed commit message.

larsks:
> I figured the PR referencing the issue would be enough

The problem there is that when I am looking at `git log`, I don't see the PR description; I only see the commit messages. I'd like to quote from one of my favorite articles on commit message:

>To echo my earlier post, how we commit code can be as important as what code we commit. The code we write will be changed over time, but the commit message we write will live forever as a snapshot of our intent, what we thought the system should do and why.
>
> Our commit messages are our love letters to the future maintainers of our code.

This is from https://www.simplethread.com/git-commit-message-101/. The key idea here is that commit messages are a persistent part of the repository, while pull requests, issues, etc, are not: they can disappear if a project changes hosting providers, or if the code is abandoned and forked somewhere else, etc.

HumairAK:
> The key idea here is that commit messages are a persistent part of the repository, while pull requests, issues, etc, are not: they can disappear if a project changes hosting providers, or if the code is abandoned and forked somewhere else, etc. 

I think in such a scenario we would probably be accepting a huge degree of loss regardless...since a lot of our discussion/decision making is taking place on Github. 

But I get the reasoning, I will update the commit with more details!

4n4nd (approved):
/lgtm

larsks (approved):
This seems fine. I'm unclear about the advantages of identitatem over keycloak, and I think it's weird that we're using argocd to deploy a placement rule to ACM to deploy a thing, but I'll try anything once.

PR #1658, opened 2022-02-23
Switch odh deployment to use olm for osc-cl2.

HumairAK:
Resolve: https://github.com/open-services-group/devsecops/issues/7


Gregory-Pereira (approved):
/lgtm

PR #1667, opened 2022-02-25
moving to lowercase

Gregory-Pereira:
Related to #1511 https://github.com/operate-first/apps/issues/1511, I think users need to not be capitalized. Trying this change before an issue is created.


HumairAK (approved):
/lgtm
/approve

PR #1668, opened 2022-02-25
updating Margaret's permissions

Gregory-Pereira:
Related to slack conversation in `sig-operations` channel

PR #1669, opened 2022-02-25
updating access to argocd docs

Gregory-Pereira:
related to #1667
related to #1668 

When using our current git-ops setup all GitHub handles must be added to groups in lowercase.


HumairAK (approved):
/lgtm
/approve

Thanks!!

PR #1673, opened 2022-03-01
new kfdef pointers and removed overrides

Gkrumbach07:
https://github.com/open-services-group/devsecops/issues/8

- point all kfdefs deployed in apps repo for the [osc-cl2](https://github.com/operate-first/apps/tree/master/odh-manifests/osc) cluster to leverage the new odh-manifests fork
- remove old odh-manifests overrides in the apps repo for osc-cl2


kfdefs/overlays/osc/osc-cl2/jupyterhub/kfdef.yaml
```diff
@@ -29,17 +29,7 @@ spec:
           name: manifests
           path: radanalyticsio/spark/cluster
       name: radanalyticsio-spark-cluster
```
4n4nd, line 31: @HumairAK do we want to deploy the spark operator? since this cluster is ocp v4.9

HumairAK: this cluster is 4.8.28, but we don't want to deploy it, however I think for this PR we should focus on just restructuring git assets and not live deployment.

HumairAK:
hrm I just realized that the old `osc-cl1` cluster is also using these `odh-manifests` that are being removed, so let's update those as well: https://github.com/operate-first/apps/tree/master/kfdefs/overlays/osc/osc-cl1

Gkrumbach07:
> This cluster will be eventually removed so this is just temporarily going to point to the `osc-cl2-*` branch in the `odh-manifests` fork.

@HumairAK  Do you the overrides removed as well, in the same fashion as `cl2`

HumairAK:
> > This cluster will be eventually removed so this is just temporarily going to point to the `osc-cl2-*` branch in the `odh-manifests` fork.
> 
> @HumairAK Do you the overrides removed as well, in the same fashion as `cl2`

yep yep

kfdefs/overlays/osc/osc-cl1/jupyterhub/kfdef.yaml
```diff
@@ -29,17 +29,7 @@ spec:
           name: manifests
           path: radanalyticsio/spark/cluster
       name: radanalyticsio-spark-cluster
-    - kustomizeConfig:
-        parameters:
```
HumairAK, line 33: any `kfdef` we're using `parameters` we should not remove them, but instead move them to the original `kustomizeConfig` for that component. So for instance add: 

```
    - kustomizeConfig:
        parameters:
          - name: s3_endpoint_url
            value: s3.odh.com
        repoRef:
          name: opf
          path: odh-manifests/osc/jupyterhub/jupyterhub
      name: jupyterhub
```

to the `kustomizeConfig` with `name: jupyterhub` and `path: jupyterhub/jupyterhub` (line 15)

same goes for the others

Gkrumbach07: think I got it, only jupyterhub and trino had params

harshad16 (commented):
lgtm
thanks :100:

PR #1681, opened 2022-03-01
adding gregory and gage to both osc-cl* odh-admin groups

Gregory-Pereira:
Same as gage's Pr #1679  but using lowercase ghandles so no issues are created.


HumairAK (approved):
/lgtm
/approve

PR #1682, opened 2022-03-01
Switch odh deployment to use olm for osc-cl1.

Gregory-Pereira:
related to  #1658.
resolves  [Devsecops issue 7](https://github.com/open-services-group/devsecops/issues/7) but for osc-cl1 instead of 2.


HumairAK (approved):
/lgtm
/approve

PR #1683, opened 2022-03-02
Fix the cosigned namespace to be default cosign-system

harshad16:
Fix the cosigned namespace to be default cosign-system

The `cosign-system` namespace is default and defined in operate-first apps [namepsaces](https://github.com/operate-first/apps/blob/0c4d42abdf446a0ebd60a0d2fb85f16648193492/cluster-scope/base/core/namespaces/cosigned/namespace.yaml#L6)


cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
```diff
@@ -24,7 +24,6 @@ resources:
 - ../../../../base/core/namespaces/aicoe-meteor
 - ../../../../base/core/namespaces/apicurio-apicurio-registry
 - ../../../../base/core/namespaces/b4mad-racing
-- ../../../../base/core/namespaces/cosigned
```
4n4nd, line 27: should be renamed instead of being removed

4n4nd (changes requested):

cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
harshad16: this is currently not needed in smaug cluster. 
removed for now, will add it when required.

4n4nd: oops looks like this needs to be removed from the smaug cluster.

4n4nd (approved):
/lgtm

PR #1687, opened 2022-03-03
Meta labels added to the project namespace manifests

bryanmontalvan:
### Issue:
Related #1661 
### Description:
These labels will store project information which will be used in the Workload Dashboard.

**Labels:**
- Owner of project (group)
- The issue which the project/workload was onboarded
- A link the projects documentation

### Note: 
- This is a draft PR, we are still figuring the link logistics
- Observatorium onboarding link is most likely not the right one

PR #1688, opened 2022-03-03
update smaug grafana instance to v8.4.3

4n4nd:
grafana v8.4 has capabilities to regex format values in a table panel, which can be very helpful in displaying data like [this](https://github.com/operate-first/apps/pull/1687).


anishasthana (commented):
/lgtm

PR #1691, opened 2022-03-03
Move idp auth resources to separate namespace.

HumairAK:
This update revises the structure for acm placements a bit

IDP authrealm and placement moved to separate namespace,
as the management of these resources is a separate concern.

ArgoCD and IDP placement setup is affected. ArgoCD integration
is simpliefied a little bit. Users would now only need to
create a ManagedClusterSetBinding in one location and it should
be deployed in both idp/argocd namespaces. ArgoCD placement
will auto pick up any ManagedCluster bounded to its namespace,
while IDP will be "enabled" but the Managed cluster will still
need a specific label set to configure the associated cluster
with the configured Authrealm (in this case opf github org).

Docs were updated to reflect this change.


acm/overlays/moc/infra/placements/argocd-managed-clusters.yaml
```diff
@@ -4,7 +4,9 @@ metadata:
...
-  clusterSets:
-    - argocd-managed
-    - octo-ushift-dev
-    - osc
+  predicates:
+    - requiredClusterSelector:
+        labelSelector:
+          matchExpressions:
+            - key: cloud
+              opreator: Exists
```
4n4nd, line 12: `operator` maybe?

4n4nd (changes requested):

acm/overlays/moc/infra/placements/argocd-managed-clusters.yaml
itdove: Did you try with an empty labelSelector as Scott suggested ?

HumairAK: @itdove so something like: 

`labelSelector: {}` ?

itdove: yes, let me quickly try in my env.

itdove: yes it works, I set the label selector as above and my current cluster kept being selected. I also import an new cluster and set the clusterset with no additional label and it get selected.

HumairAK: okay thanks @itdove !!

HumairAK: Thanks for taking the time to test it out @itdove I've updated the above.

4n4nd (approved):
/lgtm

PR #1692, opened 2022-03-04
Add lowercase constraint for Superset users

Shreyanand:
Change "Shreyanand" to "shreyanand" and update docs to make sure usernames are always lowercase.
For issue https://github.com/operate-first/support/issues/531


HumairAK (approved):
/lgtm
/approve

TY!

PR #1694, opened 2022-03-04
Added Namespace Storage Overview dashboard

bryanmontalvan:
## Issues:
Related [#445](https://github.com/operate-first/support/issues/682)
- A portion of my co-op project (expand storage monitoring, specifically through dashboards)

Related [Workload Overview Dashboard](https://github.com/orgs/operate-first/projects/50)
- This dashboard will be linked in the Workload Overview dashboard, for a more detailed of a namespace's storage information
## Description:
This namespace-specific storage dashboard is created to show users/groups a high-level overview of their namespaces storage. 

When drafting this dashboard I've found myself overcomplicating what should be on there, thinking through an engineer/operations perspective I thought of visualizing latency, IOPS, disk utilization, etc. However, from a user perspective, the main concern is how much space you have at your disposal and how close you are to getting capped out.

This dashboard does just that by showing the percentage of free space by individual namespace and PVCs. Along with other panels with similar queries though visualized differently. If you notice there are 3 panels that show available, total, and used space by cluster and this is used to reference the namespace's space compared to the cluster your project/workload is sitting on. 

#### Dashboard Contents:
- Percentage of free space by namespace and by individual PVC
- Available, Capacity, and used space panels by cluster and namespace
- Gauges of used space by cluster and namespace
- Panel which shows top 10 pods, ordered by memory usage

#### Manifest Changes;
- Created new dashboard yaml file
- Added file to dashboard/kustomization.yaml

**Link to dashboard**: [Storage Overview by Namespace Dashboard](https://grafana.operate-first.cloud/d/zkg-tWLnk/storage-overview-by-namespace?orgId=1)

image

### DIFF
--- a/grafana/overlays/moc/smaug/dashboards/cluster-management/kustomization.yaml
+++ b/grafana/overlays/moc/smaug/dashboards/cluster-management/kustomization.yaml
@@ -10,3 +10,4 @@ resources:
   - namespace_cpu_and_memory_overview.yaml
   - node-feature-discovery.yaml
   - vpa-rc-cpu-and-memory-overview.yaml
+  - namespace_storage_overview.yaml
--- a/grafana/overlays/moc/smaug/dashboards/cluster-management/namespace_storage_overview.yaml
+++ b/grafana/overlays/moc/smaug/dashboards/cluster-management/namespace_storage_overview.yaml
@@ -1,0 +1,924 @@
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata:
+  labels:
+    app: grafana
+  name: namespace-storage-overview
+spec:
+  customFolderName: Cluster Management
+  plugins:
+    - name: grafana-piechart-panel
+      version: 1.6.2
+  json: |
+    {
+      "annotations": {
+        "list": [
+          {
+            "builtIn": 1,
+            "datasource": "-- Grafana --",
+            "enable": true,
+            "hide": true,
+            "iconColor": "rgba(0, 211, 255, 1)",
+            "name": "Annotations & Alerts",
+            "target": {
+              "limit": 100,
+              "matchAny": false,
+              "tags": [],
+              "type": "dashboard"
+            },
+            "type": "dashboard"
+          }
+        ]
+      },
+      "editable": true,
+      "gnetId": null,
+      "graphTooltip": 0,
+      "id": 53,
+      "iteration": 1646404468638,
+      "links": [],
+      "panels": [
+        {
+          "datasource": null,
+          "description": "Shows total free space left in the amespace",
+          "fieldConfig": {
+            "defaults": {
+              "color": {
+                "mode": "thresholds"
+              },
+              "mappings": [],
+              "min": 0,
+              "thresholds": {
+                "mode": "absolute",
+                "steps": [
+                  {
+                    "color": "dark-red",
+                    "value": null
+                  },
+                  {
+                    "color": "red",
+                    "value": 0.25
+                  },
+                  {
+                    "color": "orange",
+                    "value": 0.4
+                  },
+                  {
+                    "color": "#EAB839",
+                    "value": 0.5
+                  },
+                  {
+                    "color": "green",
+                    "value": 0.7
+                  },
+                  {
+                    "color": "dark-green",
+                    "value": 0.85
+                  }
+                ]
+              },
+              "unit": "percentunit"
+            },
+            "overrides": []
+          },
+          "gridPos": {
+            "h": 3,
+            "w": 7,
+            "x": 0,
+            "y": 0
+          },
+          "id": 18,
+          "options": {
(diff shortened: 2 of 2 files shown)


HumairAK:
I think "Total Space on Namespace" could be more specific. It looks like here you're specifically querying persistentvolumes, so I would mention that explicitly (at least in the title/info).

Also what you are currently including is "Total space free on existing mounted PVC's", there's also "Total space available for requesting on quota", which can be obtained by querying resourcequotas on that particular namespace. I'd imagine this would be a useful visual as well.

The "% of free space by pvc" visual does not seem like it would scale well, I would convert this to something like a tabular chart.

bryanmontalvan:
@HumairAK Made changes via the feedback you gave, I'm currently working on the ResourceQuota idea you gave however it seems we only have one query for that (`kube_resourcequota`). Currently looking into how we store it (`requests.storage` so it seems). Let me know if you have any ideas or any feedback since the new changes, thanks!

Here is a screenshot of the dashboard
image

Link: https://grafana.operate-first.cloud/d/zkg-tWLnk/storage-overview-by-namespace?orgId=1&from=1646857872964&to=1646861472964

bryanmontalvan:
Hey all wanted to share the new link and to share a little preview of another version of the dashboard I'm working on. This one focuses more on performance (IOPS and latency) although I'm on the wall with showing these metrics. (Also these are just mock ups if you all think that visualizing storage performance metrics is useful I will create a refined IOP and latency panels.)

This version removes the cluster panels which gets rid a good amount of bloat, which I personally prefer. But let me know what you think and whether I should show performance metrics. Thanks!

Demo Link: https://grafana.operate-first.cloud/goto/ojhTwpsnz?orgId=1
image

Gregory-Pereira:
> Hey all wanted to share the new link and to share a little preview of another version of the dashboard I'm working on. This one focuses more on performance (IOPS and latency) although I'm on the wall with showing these metrics.

I would consider showing the IOPs of every pod to be white noise, but I think it would be useful for specific services / pods (ex: Jupyterhub, Thoth graph-backup-job, etc.). Maybe a suggestion would be to make a new metric for only some pods based on how you check the IOPs, and then display that new metric. I think this way it would allow you to display it, but only for the services where it really matters. Additionally I wouldn't expect end users to always know what IOPs is, so maybe it would be good to put the un-abbreviated version in parentheses for that graph. What do others think?

HumairAK:
> This one focuses more on performance (IOPS and latency) although I'm on the wall with showing these metrics.

I think performance metrics are great. If we can get them at the pod level that would be nice as well, since the focus for this dashboard seems to be for the namespace (not cluster, so the user may not care about IO for all nodes), but if we can't do that this is sufficient. 

Once again I think we should distinguish between "avilable space" that is available for currently allocated PVC's vs what's available for the namespace quota. Because the former can be misleading, a team can have 20gb available in their currently mounted PVC's, but as soon as they allocate another 20GB PVC (quota permitting) then suddenly their 20GB free is now 40GB. This can be relevant for namespaces where PVC's are dynamically generated like for Juptyerhub. Here I would want to know how much space quota is left for users to create their Notebooks.

bryanmontalvan:
(In response to Gregory)
> I think this way it would allow you to display it, but only for the services where it really matters. 

I agree that it would be useful for specific services although I think the act of picking and choosing isn't the route to go. Since I think that this dashboard should be give the same experience to each service. Though we all agree that storage performance metrics is useful and should be added

(In response to Humair)
> If we can get them at the pod level that would be nice as well

Good thinking, I'll work on that!

> Once again I think we should distinguish between "available space" that is available for currently allocated PVC's vs what's available for the namespace quota.

You reminded me about the ResourceQuota vs PVC comment you left last time. I was a bit curious as to what you recommend we show for each? Do you think we should have both "ResourceQuota" and "PVC" panels or should ResourceQuota take priority?

To elaborate on that I mean in the context of the "% of Free space of Namespace", "Total Space on Namespace" and "Used Space by Namespace" panels. Should I swap one of these(or all) for ResourceQuota versions or should I make separate panels which specify ResourceQuota?
E.g.,
image

HumairAK:
> Should I swap one of these(or all) for ResourceQuota versions or should I make separate panels which specify ResourceQuota?

I think you should add additional panel(s) for resourcequota, but it should be clear at first glance what the visual is showing based on its title. Instead of "Used Space by namespace" maybe say "PVC space used by Namespace" and clarify in info that this is PVC storage used for currently allocated pvc storage (for example). Then for visuals w.r.t quota, title them as "PVC Storage quota used in namespace" "PVC Storage quota available".

bryanmontalvan:
Belated update on the addition of ResoureQuota panels, only performance characteristics are left :+1: I've been spending a bit too much time on researching a predictable consumption model

image

SamoKopecky:
/remove-lifecycle stale

@bryanmontalvan Are you planning to work on this more?

PR #1704, opened 2022-03-08
Redirect oscl1 kfdefs to oscl1 odh-manifests branch.

HumairAK:
Redirect to: https://github.com/operate-first/odh-manifests/tree/osc-cl1-v1.1.0 

We'll be keeping this cluster around for a bit longer so it's only appropriate that it points to it's own branch.


Gregory-Pereira (approved):
/lgtm

PR #1705, opened 2022-03-08
Onboarding Redhat-Impact team

Gregory-Pereira:
Addresses: https://github.com/operate-first/support/issues/527.

### DIFF
--- a/cluster-scope/base/core/namespaces/chrisproject/kustomization.yaml
+++ b/cluster-scope/base/core/namespaces/chrisproject/kustomization.yaml
@@ -1,0 +1,9 @@
+kind: Kustomization
+apiVersion: kustomize.config.k8s.io/v1beta1
+namespace: chrisproject
+resources:
+    - namespace.yaml
+components:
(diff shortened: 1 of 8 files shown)


Gregory-Pereira:
@metaversed

4n4nd (approved):
/lgtm

HumairAK:
FYI I think for the future for onboarding a short commit message like "Onboarding x project" is probably sufficient, don't need to list each resource that was added :)

Gregory-Pereira:
> FYI I think for the future for onboarding a short commit message like "Onboarding x project" is probably sufficient, don't need to list each resource that was added :)

Understood. Lars was rightfully telling me that in general my commits are too vague, but I think well known and documented processes like onboarding are an exception.

Gregory-Pereira:
> you can always add more descriptions to any commit messages, I don't think anyone will fault you for that even if it's for onboarding, but it's good to keep the _subject line_ short and minimal, then leave a space, and expand in the body.

Got it :+1:

PR #1706, opened 2022-03-08
Onboarding Smartoffic

Gregory-Pereira:
Related to https://github.com/operate-first/support/issues/477.

@rbo is this what you had in mind for onboarding `Smartoffice`? Or should I have created an `rh-dus` group and used that?


rbo:
Please create a extra group for `rh-dus`.

Gregory-Pereira:
Still not sure if you wanted me to remove them from the `sa-dach` group, and use the `rh-dus` group in the namespace file, or if I was supposed to just create an extra group like how it is setup now.

rbo:
Remove from sa-Dach

cluster-scope/components/project-admin-rolebindings/rh-dus/rbac.yaml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: namespace-admin-rh-dus
```
rbo, line 5: I guess here we need a namespace right?

rbo (changes requested):
I guess the RoleBinding is missing the Namespace smartoffice.

cluster-scope/components/project-admin-rolebindings/rh-dus/rbac.yaml
Gregory-Pereira: Don't think so but I am not sure, as that would make it different from how we do role-bindings in most onboard requests, [see example](https://github.com/operate-first/apps/pull/1559/files#diff-7a6d28a664cfc05e9fa830e17bb41cf9225bcf743fd0754ee9636015989d7a2b).

rbo: Perfect, we are both right. The namespace inherited from kustomize: https://github.com/operate-first/apps/pull/1706/files#diff-921a80e29e84b44a6dc981a45278ad2ef0dd8c521bf0ae478fd8ea1f62a0af72R3

rbo (approved):
/lgtm

HumairAK (approved):
/lgtm
/approve

PR #1707, opened 2022-03-08
Onboarding Opso group and Project sandiego

Gregory-Pereira:
Related to https://github.com/operate-first/support/issues/337.

Saw this hanging so I decided to make a PR for this. Also not sure about the status of storage on `Smaug` and if we have enough storage to satisfy this request (holding).

/hold


HumairAK:
/hold

that issue is is fairly old, can we get a confirmation from the creator this is still needed?

Gregory-Pereira:
> that issue is is fairly old, can we get a confirmation from the creator this is still needed?

Yeah I dropped a comment in the issue thread, we can deal with this when she gets back to us.

HumairAK:
based on the requester's comments, let's put this on hold for now.

PR #1718, opened 2022-03-10
Update superset kfdef in osc-cl2.

HumairAK:
This change will point superset kfdef deployment to the superset
manifests supplied by odh version 1.1.2. This change allows us to manage
certain secrets ourselves instead of relying on the odh operator. As
such these seecrets have been also added to the kfdef superset app.

ODH upgrading to ODH v1.1.2 will also update Superset version to 1.3.2.


Gregory-Pereira:
We are removing the `SUPERSET_IMAGE` parameter because the it will default to latest correct? If that is the case shouldn't we lock its version at `v1.3.2`?

HumairAK:
@Gregory-Pereira Good question, ODH actually hardcoded their superset image since v1.1.1. So we can't update it via this parameter going forward.

Gregory-Pereira:
> @Gregory-Pereira Good question, ODH actually hardcoded their superset image since v1.1.1. So we can't update it via this parameter going forward.
> 
> If you're curious: Our fork: https://github.com/operate-first/odh-manifests/blob/master/superset/base/kustomization.yaml#L119 Upstream: https://github.com/opendatahub-io/odh-manifests/blob/master/superset/base/kustomization.yaml#L119

Thanks for explaining.
/lgtm

kfdefs/overlays/osc/osc-cl2/superset/kfdef.yaml
```diff
@@ -11,13 +11,15 @@ spec:
       name: odh-common
     - kustomizeConfig:
         parameters:
```
4n4nd, line 13: is this still a valid yaml?

4n4nd (approved):
/lgtm

PR #1721, opened 2022-03-10
Update trino kfdef in osc-cl2.

HumairAK:
This update will use the odh manifests for trino from version 1.1.2.

Functionally nothing should change for trino, except that worker and
coordinator jvm configs are not separated.


Gregory-Pereira (approved):
/lgtm

4n4nd (approved):
/lgtm

PR #1722, opened 2022-03-10
Deprecate prescritption refresh job application from argocd

harshad16:
Deprecate prescritption refresh job application from argocd

This doesn't require to be in smaug cluster


HumairAK (approved):
/lgtm
/approve

@harshad16 can you sync this change in  the app-of-apps-smaug (don't sync _everything_ only this application)

PR #1725, opened 2022-03-10
Introduce sudoers on morty

rbo:
- Add group sudoers and clusterrolebinding to clusterrole sudoer
- Remove  rbo from cluster-admins
- Add rbo to sudoers

I don't want to be a cluster admin all the time. Next week in sig-ops meeting I will present and show how it works.


HumairAK:
Related: https://github.com/operate-first/operations/issues/429 

@larsks fyi ^ 

We should do this for all clusters :) 

/lgtm
/approve

larsks:
@HumairAK gosh this sounds familiar... :)

Issue #1677, opened 2022-03-01
List of Services / apps to which alerts are to be added


mubariskhan96:
I'm trying to set up alerts for apps within Operate First. In this issue I will go over the apps for which I will be setting up the Alerts. I will work on configuring the Alerts in the same order as I list them here. 

So, the tentative list of apps is: 

- [ ] Jupyter Hub 
- [ ] Kafka 
- [ ] seldon
- [ ] argocd
- [ ] Trino

If you want me to do them in a different order, then please let me know. Also, what are some other apps that you want to monitor or set the alerts for? The ones I listed above are from the moc/smaug cluster.

HumairAK:
I'd remove `argo` from that list. I don't think we're even deploying it, and if we are, I don't think anyone is using it. 

argocd is a good one, but I'm not sure if we're exposing metrics for that

mubariskhan96:
Thanks!  I will tentatively replace argo with argocd there and will investigate later if the metrics for it are being exposed

ravitejamanepalli:
[[](https://github.com/SeldonIO/seldon-core/blob/master/doc/source/analytics/analytics.md
](url)

Requests from the service orchestrator to a component, e.g., a model
seldon_api_executor_client_requests_seconds_count    (bucket,count,sum) - histogram type metric
seldon_api_executor_client_requests_seconds_summary_(count,sum) - summary type metric

durandom:
@mubariskhan96 @ravitejamanepalli can we close this issue? Looks like you've added a good set of alerts already.

/assign @mubariskhan96

PR #1727, opened 2022-03-10
Adding recording rules to blackbox exporter and adding more services …

mubariskhan96:
…to service monitor in blackbox

Adding operate first, thanos, trino, cloudbeaver to the service monitor to check the liveness probe with the help of blackbox exporter. Also, adding the recording rules for different periods of time (1h, 2h, and so on). Related to this issue: https://github.com/operate-first/operations/issues/444 and https://github.com/operate-first/apps/issues/1677

/cc @4n4nd

### DIFF
--- a/kfdefs/overlays/moc/smaug/opf-jupyterhub/alerts.yaml
+++ b/kfdefs/overlays/moc/smaug/opf-jupyterhub/alerts.yaml
@@ -26,3 +26,53 @@ spec:
             < 0.1
           labels:
             severity: critical
+    - name: SLOs-probe_success
+      rules:
+        - alert: RHODS Probe Success Burn Rate
+          annotations:
+            message: 'High error budget burn for {{ $labels.instance }} (current value: {{ $value }}).'
+            triage: "https://gitlab.cee.redhat.com/service/managed-tenants-sops/-/blob/main/RHODS/JupyterHub/rhods-probe-success-burn-rate.md"
+            summary: RHODS Route Error Burn Rate
+          expr: |
+            sum(probe_success:burnrate5m{instance=~"jupyterhub"}) by (instance) > (14.40 * (1-0.98000))
+            and
+            sum(probe_success:burnrate1h{instance=~"jupyterhub"}) by (instance) > (14.40 * (1-0.98000))
+          for: 2m
+          labels:
+            severity: critical
+        - alert: RHODS Probe Success Burn Rate
+          annotations:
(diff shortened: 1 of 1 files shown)


mubariskhan96:
I'm not sure if the target ports that I have specified are correct for thanos, cloudbeaver and trino. Please advice if it is right, thanks!

grafana/overlays/moc/smaug/blackbox-exporter/recording-rules.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
```
4n4nd, line 2: applying this rule manually and checking

4n4nd: rule looks good :+1:

grafana/overlays/moc/smaug/blackbox-exporter/service-monitor.yaml
```diff
@@ -21,6 +21,72 @@ spec:
...
+          targetLabel: instance
+        - replacement: 'https://thanos-query-frontend-opf-observatorium.apps.smaug.na.operate-first.cloud/'
+          targetLabel: target
+      params:
+        module:
+          - http_2xx
+        target:
+          - 'https://thanos-query-frontend-opf-observatorium.apps.smaug.na.operate-first.cloud/'
+      path: /probe
+      targetPort: 10902
```
4n4nd, line 56: The path and targetPort here should be those of the Blackbox Exporter. So they should be same for all endpoints.

grafana/overlays/moc/smaug/blackbox-exporter/service-monitor.yaml
```diff
@@ -21,6 +21,72 @@ spec:
...
+        - replacement: 'http://cloudbeaver-opf-trino.apps.smaug.na.operate-first.cloud/'
+          targetLabel: target
+      params:
+        module:
+          - http_2xx
+        target:
+          - 'http://cloudbeaver-opf-trino.apps.smaug.na.operate-first.cloud/'
+      path: /probe
+      targetPort: 8978
   namespaceSelector:
```
4n4nd, line 90: Also need [`spec.selector`](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#servicemonitorspec). It can be empty:
```yaml
spec:
    selector: {}
```

4n4nd (changes requested):

PR #1726, opened 2022-03-10
Removal of share-icon after image not showing

bryanmontalvan:
### Issue:
The PR #1596 adds a section which instructs a user on how to backup their dashboard. For some reason the image of the *share icon* is not showing up. (Tested on Google Chrome and FireFox)
Link: https://www.operate-first.cloud/apps/content/grafana/add_grafana_dashboard.html
image

### Changes:
This PR removes the `` tag which references the **share-icon.png** image. Along with removing the image itself.
Currently working on alternatives to show the share icon.

PR #1728, opened 2022-03-11
Fix grafana share image

larsks:
It looks like `jupyter-book` only parses Markdown image references (e.g.,
``),  so it wasn't copying `img/share-icon.png` into the
published docs.

This commit replaces the HTML `` tag with a Markdown image
reference, and resizes the image file to an appropriate size.

Replaces #1726


bryanmontalvan (approved):
/lgtm

PR #1734, opened 2022-03-11
Adding Probe Success burn rate based Alerts for JupyterHub in Smaug

mubariskhan96:
Introducing the burn rate based alerts for Jupyter hub in the Smaug Cluster. Related: https://github.com/operate-first/operations/issues/444

P.S This will work once the service monitor code is deployed. Thanks!

/cc @4n4nd


mubariskhan96:
Undesired code also went into this commit. Need to change and create a fresh PR

PR #1736, opened 2022-03-11
add sandbox schema to cl2 trino

erikerlandson:


MichaelTiemannOSC (commented):
I see your sandbox, and I raise...a question.  While I see the value of there being a fat, communal sandbox, it seems to me that individual developers and/or teams need their own safe spaces in which to work and prototype.

Previously, I'd create schemaname_mdt to protect others from my work before I was ready to declare schemaname ready for use.  This sandbox doesn't seem to really have those properties.  I see the following immediate use cases (of *something*, not necessarily this concept of a sandbox):

1. A way for individual users and teams to create their own schemas where they can work without breaking things that other depend upon.
2. A way to create canonical schemas that actually represent the group of tables that define a pipeline. It could be users or teams who do this on their own say-so, or it could be they create a PR to be accepted by some Data Governance body of the Data Commons.

It could be that #2 is done by individuals and teams in DEV and done by a Data Governance body for PROD.

The good news is that I can see from this change how we need to define users (with various privs) to be able to demonstrate the Data Vault.  But first things first: we need to be able to create appropriately bounded schema that appropriate users can populate and use.

erikerlandson:
We could do that.  It would require an update to the DSL, which currently only has a concept of groups, not individual users.

In terms of providing a per-user-sandbox, I'm a bit concerned about how that scales as a policy.  Does every user get one?  We already have over a hundred osc members that are potential trino users.

MichaelTiemannOSC:
As we have shown with JupyterHub and SuperSet, it is certainly possible to create a local instance of Trino that gives individuals tons of freedom within their own environments...but in those cases they'd still need a way to connect to the canonical Data Commons if that's their aim.  So that might be a way for individual users to prototype in their own worlds.

For groups of users, the scalability might be a somewhat easily automated way to create a class of development user with a set of users associated.  In this case, osc_interns with a membership of three developers would meet the need.  But I think there are basically three categories of potential hierarchy here: workstreams and sub-workstreams (which bring together multiple pipelines and contributing companies), developers and teams within organizations (which participate in multiple workstreams and work on multiple pipelines), and ingestion pipelines and collections of pipelines (which may start off serving a single workstream, but ultimately serve multiple workstreams, with developers participating from multiple pipelines).  The key question is how to provide isolation to any given participant so they can do work without breaking others', and how to provide integration so that progress doesn't become too fragmented.

Over in the world of "code as code", GitHub has a well-defined theory of repos, teams, branches, etc., that provide useful answers to these sorts of questions.  We need to bootstrap a similar theory for "data as code" that encompasses our range of data domains, participants, and projects.

caldeirav:
I would propose we have personal sandbox schemas actually and have a sample pull request for new developers to follow. What I am not clear is whether we need to have the whole developers group as admin or we could actually specify only a given user as admin while leaving everything open for read at catalog level.

I think in Trino it would look like this somehow:

{
  "schemas": [
    {
      "role": "admin",
      "schema": ".*",
      "owner": true
    },
    {
      "user": “caldeirav”,
      "catalog": "osc_datacommons_dev",
      "schema": "sandbox_caldeirav",
      "owner": true
    }
  ]
}

This is for ownership of the sandbox schema dedicated to a given GitHub profile

"tables": [
    {
      "schema": “sandbox_caldeirav",
      "table": ".*",
      "user": “caldeirav",
      "privileges": ["SELECT", "INSERT", "DELETE", "OWNERSHIP"]
    },
    {
      "schema": “sandbox_caldeirav”,
      "table": ".*",
      "privileges": ["SELECT"],
    }
],

This would let me do everything in my schema (including table creation) and let everyone only run SELECT so I can share my data.

MichaelTiemannOSC:
Vincent, I presume the tables declarations have table stanzas with ".*" as the match, not ".".  In any case, I'm eager to give some of these ideas a try!

caldeirav:
Absolutely, I don't understand why the characters do not appear in my comment but it should be ".*" for definition at table level.

HumairAK (approved):
/lgtm
/approve

MichaelTiemannOSC:
When logged in thusly: `using connect string: trino://MichaelTiemannOSC@trino-secure-odh-trino.apps.odh-cl2.apps.os-climate.org:443/osc_datacommons_dev` I can see no schemas.  The code `engine.execute("show schemas").fetchall()` return an empty list.

 I also cannot create new schemas in osc_datacommons_dev: `TrinoUserError: TrinoUserError(type=USER_ERROR, name=PERMISSION_DENIED, message="Access Denied: Cannot create schema osc_datacommons_dev.test_mdt", query_id=20220317_094320_00002_ibmth)`

erikerlandson:
If you create `osc_datacommons_dev.sandbox` does that work?

MichaelTiemannOSC:
Nope.  Using CLI with access token:

```
trino> create schema osc_datacommons_dev.sandbox;
Query 20220317_125218_00010_ibmth failed: Access Denied: Cannot create schema osc_datacommons_dev.sandbox
```

caldeirav:
Should we maybe think about managing DEV in an open manner except for the
schemas that are restricted (i.e. we let people create them freely and THEN
restrict with access rights)?

MichaelTiemannOSC:
I successfully created a sandbox schema.  Thanks!

PR #1737, opened 2022-03-13
Update superset kfdef in osc-cl1 to ODH v1.1.2

Gregory-Pereira:
This change will point the kfdef deployment to the superset manifests in 'operate-first/odh-manifests' branch 'osc-cl1-v1.1.2'. This change allows us to manage certain secrets ourselves instead of relying on the odh operator. As such these seecrets have been also added to the kfdef superset app. ODH upgrading to ODH v1.1.2 will also update Superset version to 1.3.2.

Related to: https://github.com/open-services-group/devsecops/issues/9
Related to: https://github.com/operate-first/apps/pull/1718

My own branch has been created at https://github.com/Gregory-Pereira/odh-manifests/tree/osc-cl1-v1.1.2, and is simple to push this to `operate-first/odh-manifests`. However, because I was the one committing instead of @HumairAK the `SHA`s of the commits have changed at my forked branch compared to [branch `osc-cl2-v1.1.2`](https://github.com/operate-first/odh-manifests/tree/osc-cl2-v1.1.2) in `operate-first/odh-manifests`. Seeing as this would defeat the whole point of `cherry-pick`ing and would create an entirely new git history, for `operate-first/odh-manifests` branch `osc-cl1-v1.1.2`, @HumairAK should be the one to create the branch. In any case holding because the branch does not yet exist. To resolve this problem in the future, I will look into [`open-services-group/devsecops` issue 11](https://github.com/open-services-group/devsecops/issues/11).

/hold

PR #1738, opened 2022-03-13
Update trino kfdef in osc-cl1 to ODH v1.1.2

Gregory-Pereira:
Related to: https://github.com/operate-first/apps/pull/1721 but for `osc-cl1` instead of `osc-cl2`.


Gregory-Pereira:
Same issue about the branch not existing as mentioned in #1737. -->
/hold

PR #1742, opened 2022-03-14
[grafana] Add the pulp-metrics dashboard

codificat:
This is to add a `GrafanaDashboard` CR for the pulp metrics dashboard.

I was looking at https://github.com/thoth-station/pulp-metrics-exporter/issues/17#issuecomment-1031244059 and could not find the dashboard that is mentioned in that comment.

Looking at the instructions on https://github.com/thoth-station/thoth-application/tree/master/grafana-dashboard#create-new-dashboards I believe that steps 1 and 2 of dashboard creation (i.e. create a JSON file for the dashboard) were implemented in https://github.com/thoth-station/thoth-application/pull/2326, but step 3 (the grafana CR) is still missing. This PR is to add that.

I am not fully sure of the proper location for the dashboard, or if I missed something else.

### DIFF
--- a/grafana/overlays/moc/smaug/dashboards/kustomization.yaml
+++ b/grafana/overlays/moc/smaug/dashboards/kustomization.yaml
@@ -8,6 +8,7 @@ resources:
   - operate-first
   - prometheus
   - prometheus-anomaly-detector
+  - pulp
   - thanos
   - thoth-station
   - kubeflow
--- a/grafana/overlays/moc/smaug/dashboards/pulp/kustomization.yaml
+++ b/grafana/overlays/moc/smaug/dashboards/pulp/kustomization.yaml
@@ -1,0 +1,5 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - pulp-metrics.yaml
--- a/grafana/overlays/moc/smaug/dashboards/pulp/pulp-metrics.yaml
+++ b/grafana/overlays/moc/smaug/dashboards/pulp/pulp-metrics.yaml
@@ -1,0 +1,10 @@
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata:
+  labels:
+    version: v1
+    app: grafana
+  name: pulp-metrics
+spec:
+  customFolderName: Pulp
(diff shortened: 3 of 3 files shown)


codificat:
cc @pacospace

grafana/overlays/moc/smaug/dashboards/operate-first/pulp-metrics.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata:
+  labels:
+    app: grafana
```
pacospace, line 5: what about `version` and `component` labels? to be consistent with the other CRDs?

pacospace:
Please also add the folder name as in https://github.com/operate-first/apps/pull/1744

grafana/overlays/moc/smaug/dashboards/operate-first/pulp-metrics.yaml
codificat: well, I copied mine from the `opf-availability.yaml` in the same dir... so, it is consistent with that one :innocent: 

I see that the dashboards related to thoth have a `version` and a `component` (= thoth-station), so I am adding a `version` now... but it's unclear to me what `component` I should use.

There are no guidelines around labels in https://www.operate-first.cloud/apps/content/grafana/add_grafana_dashboard.html, and it seems that the thoth dashboards are the only ones that currently specify a `component`... but pulp is not specific to thoth.

Any suggestions for `component`?

codificat:
> @codificat Please also add the folder name as in #1744

It was already there, but set to `Operate-First` instead of `thoth-station`. Should we have that dashboard under `thoth-station`?

grafana/overlays/moc/smaug/dashboards/operate-first/pulp-metrics.yaml
pacospace: no problem @codificat, when I talk about consistency, I think about Thoth usually :) 

I think `component` label is also a usual label we have across all thoth components. Maybe @harshad16 has more insights here :)

pacospace:
> > @codificat Please also add the folder name as in #1744
> 
> It was already there, but set to `Operate-First` instead of `thoth-station`. Should we have that dashboard under `thoth-station`?

I'm sorry I didn't see it. Yes let's have it like for the other `thoth-station` dashboards :)

PR #1754, opened 2022-03-17
Add cuda buildchain to osc-cl2 jupyterhub.

HumairAK:
This change will instruct ODH on osc-cl2 to intiate cuda image builds
which will produce GPU enabled Jupyterhub images, and display them in
the JH spawner.


4n4nd (approved):
/lgtm

PR #1763, opened 2022-03-18
Remove invalid argocd cluster secret.

HumairAK:
The secret is breaking the build, this is a quick fix. We should update the secret if we still need it and re-encrypt.


Gregory-Pereira (approved):
Tested and agreed.
/lgtm
/approve

harshad16 (approved):
/lgtm
/approve

PR #1771, opened 2022-03-21
Add myself to the microshift group

codificat:
This PR is to add myself to the `octo-ushift-dev` group.

The main goal is to get access to the resources managed through ArgoCD for the `octo-ushift` project.

Context: I was trying to troubleshoot current sync problems between Argo and the microshift clusters involved in the octo summit demo, and I am missing part of the picture without that access.


Gregory-Pereira (approved):
/lgtm
/approve

PR #1776, opened 2022-03-22
Adding burn rate based alerts for JupyterHub in Smaug cluster

mubariskhan96:
Introducing burn rate based alerts for JupyterHub in the Smaug Cluster. Related: https://github.com/operate-first/operations/issues/444

/cc @4n4nd

### DIFF
--- a/kfdefs/overlays/moc/smaug/opf-jupyterhub/alerts.yaml
+++ b/kfdefs/overlays/moc/smaug/opf-jupyterhub/alerts.yaml
@@ -26,3 +26,41 @@ spec:
             < 0.1
           labels:
             severity: critical
+    - name: SLOs-probe_success
+      rules:
+        - alert: Probe Success Burn Rate
+          annotations:
+          expr: |
+            sum(probe_success:burnrate5m{opf-instance=~"jupyterhub"}) by (opf-instance) > (14.40 * (1-0.98000))
+            and
+            sum(probe_success:burnrate1h{opf-instance=~"jupyterhub"}) by (opf-instance) > (14.40 * (1-0.98000))
+          for: 2m
+          labels:
+            severity: critical
+        - alert: Probe Success Burn Rate
+          annotations:
+          expr: |
+            sum(probe_success:burnrate30m{opf-instance=~"jupyterhub"}) by (opf-instance) > (6.00 * (1-0.98000))
+            and
+            sum(probe_success:burnrate6h{opf-instance=~"jupyterhub"}) by (opf-instance) > (6.00 * (1-0.98000))
+          for: 15m
+          labels:
+            severity: critical
+        - alert: Probe Success Burn Rate
+          annotations:
+          expr: |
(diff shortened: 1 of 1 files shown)


kfdefs/overlays/moc/smaug/opf-jupyterhub/alerts.yaml
```diff
@@ -26,3 +26,53 @@ spec:
             < 0.1
           labels:
             severity: critical
+    - name: SLOs-probe_success
+      rules:
+        - alert: RHODS Probe Success Burn Rate
```
4n4nd, line 31: @mubariskhan96 RHODS  is an internal only service, so this alert name does not make sense.

kfdefs/overlays/moc/smaug/opf-jupyterhub/alerts.yaml
```diff
@@ -26,3 +26,53 @@ spec:
...
+          annotations:
+            message: 'High error budget burn for {{ $labels.instance }} (current value: {{ $value }}).'
```
4n4nd, line 33: I don't think we have a label `instance` in our metrics

kfdefs/overlays/moc/smaug/opf-jupyterhub/alerts.yaml
```diff
@@ -26,3 +26,53 @@ spec:
...
+            triage: "https://gitlab.cee.redhat.com/service/managed-tenants-sops/-/blob/main/RHODS/JupyterHub/rhods-probe-success-burn-rate.md"
```
4n4nd, line 34: This is an internal only link, we would need to see if we can add it to the apps repo

kfdefs/overlays/moc/smaug/opf-jupyterhub/alerts.yaml
```diff
@@ -26,3 +26,53 @@ spec:
...
+            summary: RHODS Route Error Burn Rate
+          expr: |
+            sum(probe_success:burnrate5m{instance=~"jupyterhub"}) by (instance) > (14.40 * (1-0.98000))
```
4n4nd, line 37: no `instance` label

4n4nd (changes requested):

kfdefs/overlays/moc/smaug/opf-jupyterhub/alerts.yaml
mubariskhan96: fixed it. I hope the fix is correct! I verified it is syntactically correct

kfdefs/overlays/moc/smaug/opf-jupyterhub/alerts.yaml
mubariskhan96: removed that link

kfdefs/overlays/moc/smaug/opf-jupyterhub/alerts.yaml
mubariskhan96: fixed it. I hope the fix is correct! I verified it is syntactically correct

PR #1777, opened 2022-03-22
Add oc pipelines as a bundle to os-cl2.

HumairAK:
Add oc pipelines to osc-cl2. This commit also bundles the manifests that
currently exist in the base dir and switches to bundle reference in the
smaug cluster as well. No change to live deployment to the smaug cluster
is introduced here. The patch for the subscription was removed since the
channel remains the same for both clusters, it just adds an unecessary
step otherwise.

### DIFF
--- a/cluster-scope/base/operators.coreos.com/subscriptions/openshift-pipelines-operator-rh/subscription.yaml
+++ b/cluster-scope/base/operators.coreos.com/subscriptions/openshift-pipelines-operator-rh/subscription.yaml
@@ -4,7 +4,7 @@ kind: Subscription
 metadata:
   name: openshift-pipelines-operator-rh
 spec:
-  channel: DEFINED_IN_OVERLAY
+  channel: stable
   installPlanApproval: Automatic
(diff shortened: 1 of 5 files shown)


cluster-scope/base/operators.coreos.com/subscriptions/openshift-pipelines-operator-rh/subscription.yaml
```diff
@@ -4,7 +4,7 @@ kind: Subscription
 metadata:
   name: openshift-pipelines-operator-rh
 spec:
-  channel: DEFINED_IN_OVERLAY
```
Gregory-Pereira, line 7: Any reason we are not just setting this in the OSC overlays? Do no other overlays use different channels?

HumairAK: I just set it in the base since it's the same for both clusters. I don't see a reason to patch it separately. Less work imo.

PR #1779, opened 2022-03-22
Add kfp-tekton app to osc-cl2.

HumairAK:
This emulates the kfp-tekton deployment from smaug onto osc-cl2.
The bucket being used is one created in the smaug cluster via odf. Once confirmed working we can switch to a more permanent solution.

### DIFF
--- a/argocd/overlays/moc-infra/applications/envs/osc/osc-cl2/cluster-management/kfp-tekton.yaml
+++ b/argocd/overlays/moc-infra/applications/envs/osc/osc-cl2/cluster-management/kfp-tekton.yaml
@@ -1,0 +1,20 @@
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: kfp-tekton
+spec:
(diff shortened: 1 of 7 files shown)


argocd/overlays/moc-infra/applications/envs/osc/osc-cl2/cluster-management/kfp-tekton.yaml
```diff
@@ -0,0 +1,20 @@
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: kfp-tekton
+spec:
+  destination:
+    name: osc-cl2
+    namespace: kubeflow
+  project: operate-first
```
Gregory-Pereira, line 10: Are we sure the project should be `operate-first`? all the other argocd cluster resource applications in this overlay use `os-climate`.

Gregory-Pereira (commented):
Aside from making sure the project is right,
/lgtm

PR #1787, opened 2022-03-23
Add minio artifact secret to osc-cl2 kfptekton.

HumairAK:
The copy artifact archival step in kfp-tekton requires the existence of
this secret. For more details read the description annotation included
in this commit. Description is added as an annotation since sops will
purge yaml comments upon encryption/decryption.


harshad16 (approved):
/lgtm
/approve
:+1:

PR #1800, opened 2022-03-25
modifying kustomization overlays in each cluster for `opfcli onboard`

Gregory-Pereira:
Kustomization overlays have been restructured to be compatible with the `opfcli onboard` command. This will produce, small and isolated diffs when on-boarding with the `opfcli` rather than large ones that compete over formatting rules. Resources will now also be automatically sorted.

Related to: merging of [opfcli pr 44](https://github.com/operate-first/opfcli/pull/44)


SamoKopecky:
/remove-lifecycle stale

@Gregory-Pereira Are you planning to work on this more?

Gregory-Pereira:
were moving away from using the opfcli for onboarding.
/close

PR #1807, opened 2022-03-28
Add pachyderm operator and deployment to osc-cl2.

HumairAK:
This PR creates a new bundle in base directory comprising of the
pachyderm operator subscription and the namespace to which pachyderm
will be deployed. The bundle is included in the osc-cl2 cluster
kustomization. There is also included a pachyderm custom resource
deployment that will instruct the operator to deploy pachyderm into
the namespace. An argocd application that will deploy the newly
introduced pachyderm custom resources to the osc-cl2 is also included.


HumairAK:
/hold
For aws s3 bucket to be created for this pachyderm instance as per https://github.com/pachyderm/openshift-operator#usage
I'll update the PR once that's done.

PR #1808, opened 2022-03-28
[dashboard] Fix the node feature discovery grafana dashboard

4n4nd:
This will make the metrics datasource configurable so in future when we add more clusters, it should dynamically update the list


HumairAK (approved):
/lgtm
/approve

PR #1814, opened 2022-03-29
update cl2 size options

### LINKED ISSUE #1748: Elyra notebooks for OS-Climate do have GPU selector but not memory size selector
In a previous incarnation, it was possible to select RAM size for Elyra notebooks (from 4GB to 32GB I believe).  And while it is possible to select whether the notebook uses a single GPU or no GPU, it is not possible to select RAM size, which is problematic for loading PUDL data (which momentarily needs lots of ram to process a 3.9GB tarfile that sits within a 6.5GB tgz file).  When the tarfile is skipped, the notebook can complete.  But when it is not skipped, the notebook is killed due to OOM.

Can we please have either 4GB, 8GB, 16GB, 32GB (S, M, L, XL) or 4GB, 16GB, 64GB (S, M, L) as selections?

erikerlandson:
Fixes #1748

### DIFF
--- a/kfdefs/overlays/osc/osc-cl2/jupyterhub/profiles/singleuser-profiles-container-sizes.yaml
+++ b/kfdefs/overlays/osc/osc-cl2/jupyterhub/profiles/singleuser-profiles-container-sizes.yaml
@@ -31,22 +31,6 @@ data:
         limits:
           memory: 32Gi
           cpu: 4
-    - name: '1 cpu, 2GB mem'
-      resources:
-        requests:
-          memory: '1Gi'
-          cpu: '1'
-        limits:
-          memory: '2Gi'
-          cpu: '1'
-    - name: '1 cpu, 4GB mem'
-      resources:
-        requests:
-          memory: '2Gi'
-          cpu: '1'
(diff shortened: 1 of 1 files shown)


erikerlandson:
I think kustomize-build may be broken
```
Checking changes in range 75aff0e1dc65c5fe10081cec7bc30a03ae827187..2ff35c916058929ee84675146b955b4533b16e3a only.
checking kfdefs/overlays/dev
Error: accumulating resources: accumulation err='accumulating resources from 'opf-kafka': evalsymlink failure on '/home/prow/go/src/github.com/operate-first/apps/kfdefs/overlays/dev/opf-kafka' : lstat /home/prow/go/src/github.com/operate-first/apps/kfdefs/overlays/dev/opf-kafka: no such file or directory': evalsymlink failure on '/home/prow/go/src/github.com/operate-first/apps/kfdefs/overlays/dev/opf-kafka' : lstat /home/prow/go/src/github.com/operate-first/apps/kfdefs/overlays/dev/opf-kafka: no such file or directory
Aborting due to errexit on line 43. Exit code: 1
```

PR #1821, opened 2022-03-29
Add pachyderm osc admin group.

HumairAK:
Set up a group that has admin access to the pachyderm namespace within the osc-cl2 cluster.

Also added a commit to update the pachyderm resource as the operator does not accept the metrics field. 

Also added owners file with osc contributors.


Gregory-Pereira (commented):
/lgtm
Only question is can we / should we have multiple `OWNERS` files in our repository?

HumairAK:
@Gregory-Pereira yes

caldeirav:
Ultimately we want a single osc-admin group that has admin rights for pachyderm, trino, superset and all future components of the data platform. Is this the path we are taking?

HumairAK:
@Gregory-Pereira can you re-review

HumairAK:
@caldeirav I'll create the issue and clarify terms/etc so that we can hash out what sort of groups we should have.

dystewart (approved):
/lgtm

PR #1822, opened 2022-03-29
Fixing vault statefulset.yaml to include missing volume claims

dystewart:
Some of the volume claim templates got cut off at some point causing the vault deployment to fail in smaug. I've added the necessary volume claim templates so that the deployment should complete successfully.


HumairAK (approved):
/lgtm
/approve

Awesome, thanks @dystewart !

PR #1824, opened 2022-03-29
Add minio to osc ml pipelines deployment.

HumairAK:
This commit switches back to using minio deployment as provided by
upstream kfp-tekton for artifact passing. There seemed to be some
hardcoded mentions of minio within ml pipelines code and this was
causing issues for the running deployment.


4n4nd (approved):
/lgtm

PR #1830, opened 2022-03-30
adjust vault pod volume claims and namespace resource quota

dystewart:
I didn't account for the deployment being HA so I did not designate enough storage in the resource quota initially. I've also scaled down the volume claims a bit.


HumairAK (approved):
/lgtm
/approve

PR #1835, opened 2022-03-31
Use identitatem to manage morty cluster auth.

HumairAK:
This change will remove morty keycloak client
and instead use Identitatem to provision the
openshift auth configuration for morty.

Some docs and comments are also introduced
to aid with future cluster incorporations.


Gregory-Pereira:
> Since you remove the keycloak infra overlay secret generator, remove it from the `kustomization.yaml`:
> https://github.com/operate-first/apps/blob/7f1986b5dc5893d2a1627978f89326b0463bd00c/keycloak/overlays/moc/infra/kustomization.yaml#L6

My bad I thought you deleted the file, but you only removed morty from it. Ignore that comment.
/approve

PR #1859, opened 2022-04-05
Remove oauth base resource from morty cluster.

HumairAK:
This is a leftover from a previous pr where we switched over to using
identitatem to manage auth for morty cluster. #1835


Gregory-Pereira (commented):
/lgtm

PR #1860, opened 2022-04-05
Update mapping method for idp authrealm.

HumairAK:
We found that using the "claim" mapping method results in user identity
conflicts. Instead we'll use the "add" method which allows for the
mapping of a username to any existing user on the cluster.

Further reading: 
https://docs.openshift.com/container-platform/4.7/authentication/understanding-identity-provider.html#identity-provider-parameters_understanding-identity-provider


dystewart (commented):
/lgtm

Issue #1662, opened 2022-02-24
[workload-overview] How can we make the dashboard public?


4n4nd:
If we make a single stop grafana dashboard for opf workloads, how do we make it public since the current grafana instance is behind a proxy.

bryanmontalvan:
After some digging, I think that going the snapshot route would be a good since if we share a snapshot of a dashboard or panel the following happens to both dashboard and panel snapshots.
> Grafana strips sensitive data like queries (metric, template and annotation) and panel links, leaving only the visible metric data and series names embedded into your dashboard. Dashboard snapshots can be accessed by anyone with the link.

The only difference between **dashboard snapshots** and **panel snapshots** is that panel snapshots can be embedded to a web-page using `<iframe src="https://play.grafana....">` 

The only issue I'm seeing is that **"Are these snapshot dashboards being updated live?"** Because after some testing it seems to only show the metrics from the time that the **snapshot share link** was created to 6 hours before that. And whenever another time-frame is selected an `Unauthorized` error is shown.

**Links:**
Share a panel: https://grafana.com/docs/grafana/latest/sharing/share-panel/
Snapshot API: https://grafana.com/docs/grafana/latest/http_api/snapshot/

4n4nd:
@bryanmontalvan Yeah we want the data to not be static. So if a new service/app is onboarded, the dashboard should update automatically to include it.

I think @HumairAK's suggestion to go with a separate public grafana instance makes more sense for this.

HumairAK:
we have a public grafana instance now for public dashboards https://github.com/operate-first/apps/tree/master/grafana/overlays/moc/smaug-public

PR #1867, opened 2022-04-06
Deploy apublic instance of Grafana on Smaug

4n4nd:
* Add `opf-grafana-public` namespace
* Deploy grafana-operator in `opf-grafana-public` namespace
* Add grafana-public argocd app
* Add smaug-public overlay in grafana operate-first app

Related: https://github.com/operate-first/apps/issues/1662


cluster-scope/base/operators.coreos.com/subscriptions/opf-grafana-public/subscription.yaml
```diff
@@ -0,0 +1,11 @@
+---
```
HumairAK, line 1: The subscription here is the same as this one: 

https://github.com/operate-first/apps/tree/master/cluster-scope/base/operators.coreos.com/subscriptions/opf-grafana

I think we should rename `opf-grafana` to `grafana-operator` 

Then create a bundle for opf-grafana-public. In it, include: 

`base/core/namespaces/opf-grafana-public`
`base/operators.coreos.com/operatorgroups/opf-grafana-public`
`base/operators.coreos.com/subscriptions/grafana-operator`

grafana/overlays/moc/smaug-public/grafana-public.yaml
```diff
@@ -0,0 +1,25 @@
...
+      org_role: Viewer
+    log:
+      level: warn
+      mode: console
+  dashboardLabelSelector:
+    - matchLabels:
+        app: grafana
+        public: "true"
+  ingress:
+    enabled: true
```
HumairAK, line 25: since we're creating our own route, can we disable this?

cluster-scope/base/operators.coreos.com/subscriptions/opf-grafana-public/subscription.yaml
4n4nd: added the bundle :+1:

cluster-scope/bundles/opf-grafana-public/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: opf-grafana-public
+resources:
+  - ../../base/core/namespaces/opf-grafana-public
+  - ../../base/operators.coreos.com/operatorgroups/opf-grafana-public
+  - ../../base/operators.coreos.com/subscriptions/grafana-operator
```
HumairAK, line 7: I think grafana resources are getting a bit disorganized, can we clean them up as part of this pr? 

Right now the subscription is the same as [this](https://github.com/operate-first/apps/blob/master/cluster-scope/base/operators.coreos.com/subscriptions/opf-grafana/subscription.yaml) one. 

And we also have a bundle for grafana [here](https://github.com/operate-first/apps/blob/master/cluster-scope/bundles/opf-monitoring/kustomization.yaml)

Can we have these 2 bundles in their stead: 

opf-grafana
  - ../../base/core/namespaces/opf-monitoring
  - ../../base/rbac.authorization.k8s.io/clusterrolebindings/opf-monitoring-grafana-cluster-view
  - grafana-serviceaccount.yaml
  - ../../base/operators.coreos.com/subscriptions/grafana-operator

opf-grafana-public 
  - ../../base/core/namespaces/opf-grafana-public
  - ../../base/operators.coreos.com/operatorgroups/opf-grafana-public
  - ../../base/operators.coreos.com/subscriptions/grafana-operator

what do you think? would that be better?

HumairAK: hrmm I'm realizing now that would not work, we don't use `opf-monitoring` bundling to only install the operator

HumairAK: okay for now just use the same `subscription` [here](https://github.com/operate-first/apps/blob/master/cluster-scope/base/operators.coreos.com/subscriptions/opf-grafana/subscription.yaml) in both places, since the bundle kustomization will update the namespace. That should at least reduce duplication.

4n4nd:
```
../../base/core/namespaces/opf-monitoring
../../base/rbac.authorization.k8s.io/clusterrolebindings/opf-monitoring-grafana-cluster-view
```
These are independent of the grafana operator so I have made it as a separate bundle. Plus it doesn't make sense to restructure unrelated stuff in this PR, we can make a separate PR for that.

4n4nd: > okay for now just use the same subscription here in both places, since the bundle kustomization will update the namespace. That should at least reduce duplication.

yep that is how it is rn.

PR #1868, opened 2022-04-06
Introducing servicemonitors for argocd

mubariskhan96:
The argocd metrics are currently not exposed to prometheus. And for that we would need the service monitors for argocd-metrics, argocd-server-metrics and argocd-repo-server. More details on why that is so we can found here: https://argo-cd.readthedocs.io/en/stable/operator-manual/metrics/ .

 Related issue: https://github.com/operate-first/apps/issues/1677


argocd/base/servicemonitors/argocd-repo-server
```diff
@@ -0,0 +1,13 @@
+apiVersion: monitoring.coreos.com/v1
```
Gregory-Pereira, line 1: This file should be a `.yaml` file.

Gregory-Pereira (commented):
Aside from `argocd-repo-server` not being a `.yaml` file.

/lgtm

argocd/base/servicemonitors/argocd-repo-server
mubariskhan96: ok will fix and make a push, thanks!

mubariskhan96: Yeah! Good catch! I just added it along with the prev fix

argocd/base/servicemonitors/argocd-metrics.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: argocd-metrics
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: argocd-metrics
+  endpoints:
+  - port: metrics
+  namespaceSelector:
+    matchNames:
+      - opf-monitoring
```
4n4nd, line 13: argocd is deployed in namespace: `argocd`

mubariskhan96: fixed, thanks!

4n4nd (approved):
/lgtm

PR #1872, opened 2022-04-07
fix `opf_instance` label in Jupyterhub alerts and blackbox servicemonitor

4n4nd:
`opf-instance` is not a valid label name and is being replaced  by `opf-instance`
This is because `-` is not an accepted character, it needs to be replaced with `_`


HumairAK (approved):
/lgtm
/approve

Issue #1843, opened 2022-04-01
upgrade OSC clusters (cl1, cl2) to Trino 373


erikerlandson:
We'll need a `trino:373` image to be built first.
373 includes some bug fixes that relate to problems @MichaelTiemannOSC and I have been seeing around dropping tables in iceberg catalogs, and also includes a delta-lake connector.
cc @caldeirav @rimolive

erikerlandson:
https://github.com/os-climate/os_c_data_commons/issues/154

erikerlandson:
@HumairAK   when I get a 373 image, we would like to figure out how to re-initialize the CL2 trino deployment from a totally clean state on the underlying s3 bucket (so any existing tables would be lost, but we are OK with that).  This particular deployment began with trino:362 and has known bugs relating to s3 file management under the hood, so we can't trust the current state of the files.
cc @MichaelTiemannOSC

HumairAK:
> when I get a 373 image, we would like to figure out how to re-initialize the CL2 trino deployment from a totally clean state on the underlying s3 bucket 

Is that another way of saying, just wipe the bucket clean?

MichaelTiemannOSC:
Yes, recursively from osc-datacommons-s3-bucket-dev02/data on down, leaving osc-datacommons-s3-bucket-dev02 pristine.


Michael

________________________________

erikerlandson:
> Is that another way of saying, just wipe the bucket clean?
yes

erikerlandson:
there was a suggestion that maximum open files could be an issue for partition writers, but the current ulimits don't seem very restricted:
```
sh-4.4$ ulimit -a
core file size          (blocks, -c) unlimited
data seg size           (kbytes, -d) unlimited
scheduling priority             (-e) 0
file size               (blocks, -f) unlimited
pending signals                 (-i) 254062
max locked memory       (kbytes, -l) 64
max memory size         (kbytes, -m) unlimited
open files                      (-n) 1048576
pipe size            (512 bytes, -p) 8
POSIX message queues     (bytes, -q) 819200
real-time priority              (-r) 0
stack size              (kbytes, -s) 8192
cpu time               (seconds, -t) unlimited
max user processes              (-u) 4194304
virtual memory          (kbytes, -v) unlimited
file locks                      (-x) unlimited
```

erikerlandson:
`quay.io/os-climate/trino:373` is now available

PR #1877, opened 2022-04-07
upgrade OSC trino images to quay.io/os-climate/trino:373

erikerlandson:
Fixes https://github.com/operate-first/apps/issues/1843

xref: https://github.com/os-climate/os_c_data_commons/issues/154


MichaelTiemannOSC:
Will we also get a clean slate with the upgrade to 373?  The out-of-sync metadata and the amount of garbage left over from tables dropped but not deleted makes for a messy starting point.  The only tables in the sandbox schema were created by me and I can easily recreate them.

PR #1880, opened 2022-04-08
Add training operator.

HumairAK:
Adds various crds/rbac to cluster-scope base folder, and also adds a new
argocd app to deploy the training operator onto the osc-cl1 cluster.

One of the crds was over default "large file" size limit set by the
"check-added-large-files" hook, so I increased this maximum to 600kb.

Related: https://github.com/operate-first/support/issues/537


4n4nd (approved):
/lgtm

PR #1883, opened 2022-04-09
Set query.max-length=10M (overriding 1M default)

MichaelTiemannOSC:
The default query.max-length=1000000.  For datasets such as SEC_DERA, the 1M limit allows for approx 1500 rows per insertion.  When there are millions of rows, this not only creates thousands of snapshots, but each snapshot file also references the chain of snapshots, growing these files from a few K to several MB per snapshot.

By increasing the max-length parameter to 10M, we not only cut down the number of objects in the metastore, but we also decrease the number of snapshots each of these objects needs to reference, reducing storage requirements by a factor of 100 (10*10).


HumairAK:
FYI @MichaelTiemannOSC , if you join the operate-first org your prs will automatically run ci checks (without needing another member type /ok-to-test). It's easy to do (just adding one line to a file). Instructions https://github.com/operate-first/common/blob/main/docs/add_gh_member_and_access.md#become-a-github-member.

PR #1890, opened 2022-04-11
Set query.max-length=10M in correct files

MichaelTiemannOSC:
Per guidance from Trino slack channel:

"Everything unless mentioned needs to be set in both coordinators and workers. Only things like security configs and global system level things (like overall cluster memory) are required only on coordinator."


HumairAK:
here goes round 2
/lgtm
/approve

PR #1891, opened 2022-04-11
Switch to manual odh deploy on osc clusters.

HumairAK:
In this cluster we have handed out project admin permissions to some
users so we are forced to remove odh from olm management due to security
reasons. Otherwise these users will be able to deploy kfdefs.


Gregory-Pereira (commented):
/lgtm

Gregory-Pereira (commented):
/approve

PR #1892, opened 2022-04-12
feat(osc-cl1): Fix BYON deployment

tumido:
Related to: https://github.com/open-services-group/byon/issues/20
Requires: https://github.com/operate-first/odh-manifests/pull/19

- Deploying ODH Dashboard to `jupyterhub-stage`: Upstream of ODH Dashboard has no notion of namespaces and treats everything like it's in the same namespace, that means BYON pipelines needs to live in the same namespace as JupyterHub and Dashboard, that's the only way how Dashboard is able to trigger the pipelines and JH to see the resulting imagestreams.
- Reverting the `odh-dashboard` namespace to `v1.1.2` tag.
- Deploying BYON from JH overlay

/cc @harshad16

### DIFF
--- a/kfdefs/overlays/osc/osc-cl1/dashboard/kfdef.yaml
+++ b/kfdefs/overlays/osc/osc-cl1/dashboard/kfdef.yaml
@@ -14,5 +14,5 @@ spec:
       name: odh-dashboard
   repos:
     - name: manifests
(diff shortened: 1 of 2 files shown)


harshad16 (approved):
sounds good
/lgtm

PR #1893, opened 2022-04-12
feat(osc-cl1): Add openshift-pipelines

tumido:
Required for https://github.com/open-services-group/byon/issues/20#issuecomment-1096620589


harshad16 (changes requested):
If our goal is to make BYON deployment same a prod like in the stage deployment, we should do this via Kfdef as ODH has openshift-pipeline in their [manifests](https://github.com/operate-first/odh-manifests/tree/master/openshift-pipelines), that would the way it would install in prod system as well.
How about following the same procedure?

HumairAK:
I don't really see a point in using odh to deploy operators, we've been moving away from odh components in operate-first that are simply "deploy this subscription" (see strimzi/grafana/etc). 

/lgtm
/approve
/unhold

tumido:
please don't merge yet. @harshad16 chatted with me and told me that this may introduce conflict for current Kubeflow installation on that cluster since that one brings bundled Tekton.

PR #1897, opened 2022-04-13
Remove unsupported crd fields in kfdef crd.

HumairAK:
Kfdef pulled from here:  https://github.com/opendatahub-io/opendatahub-operator/blob/master/deploy/crds/kfdef.apps.kubeflow.org_kfdefs_crd.yaml

Has fields that don't seem supported by the [current api spec](https://docs.openshift.com/container-platform/4.9/rest_api/extension_apis/customresourcedefinition-apiextensions-k8s-io-v1.html#apisapiextensions-k8s-iov1customresourcedefinitionsname), namely `subresources` , `validation`, `version`. 

We should follow up once this is remedied upstream.


HumairAK:
FYI this has no affect on live, as this information is stripped away as soon as we deploy a kfdef (since these fields are not suported). But because argocd keeps seeing these fields not present, it will keep trying to sync: 

image

PR #1910, opened 2022-04-19
Add argocd project for peribolos service

Shreyanand:
For issue https://github.com/open-services-group/peribolos-as-a-service/issues/32

@harshad16 @tumido

### DIFF
--- a/argocd/overlays/moc-infra/configs/argo_rbac_cm/policy.csv
+++ b/argocd/overlays/moc-infra/configs/argo_rbac_cm/policy.csv
@@ -30,3 +30,4 @@ g, sre, role:standard-user
 g, fybrik, role:standard-user
 g, tremor-demo, role:standard-user
 g, osc-admins, role:standard-user
+g, peribolos-as-a-service, role:standard-user
--- a/argocd/overlays/moc-infra/projects/peribolos-service.yaml
+++ b/argocd/overlays/moc-infra/projects/peribolos-service.yaml
@@ -1,0 +1,26 @@
+apiVersion: argoproj.io/v1alpha1
+kind: AppProject
+metadata:
+  name: peribolos-service
+  labels:
+    project-template: global
+spec:
+  destinations:
+    - namespace: peribolos-as-a-service    # if there is no generic team name prefix, then specify namespaces individually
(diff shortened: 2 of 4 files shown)


argocd/overlays/moc-infra/projects/peribolos-service.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: argoproj.io/v1alpha1
+kind: AppProject
+metadata:
+  name: peribolos-as-a-service
+  labels:
+    project-template: global
+spec:
+  destinations:
+    - namespace: '*'
+      name: 'smaug'
```
tumido, line 10: This grants the project access to any namespace on smaug, we don't want that.

argocd/overlays/moc-infra/configs/argo_rbac_cm/policy.csv
```diff
@@ -30,3 +30,5 @@ g, sre, role:standard-user
 g, fybrik, role:standard-user
 g, tremor-demo, role:standard-user
 g, osc-admins, role:standard-user
+g, peribolos-as-a-service, role:standard-user
```
tumido, line 33: I think there was some reason to *not* include the EOL at EOF here. cc @HumairAK for confirmation.

argocd/overlays/moc-infra/projects/peribolos-service.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: argoproj.io/v1alpha1
+kind: AppProject
+metadata:
+  name: peribolos-as-a-service
```
tumido, line 4: Filename should match this name here, I think. `peribolos-service.yaml` file vs `peribolos-as-a-service` project.

tumido (changes requested):

argocd/overlays/moc-infra/configs/argo_rbac_cm/policy.csv
HumairAK: yeah there was...but the reason is escaping me, so just be safe and keep it without eol

tumido (approved):
/lgtm
/approve

PR #1914, opened 2022-04-20
row/col access controls on sandbox.gppd

erikerlandson:
Signed-off-by: Erik Erlandson <eerlands@redhat.com>


harshad16 (approved):
lgtm when compare the similar format from existing code.
/lgtm

PR #1927, opened 2022-04-27
add grafana operator to b4mad-racing namespace

durandom:
@4n4nd @HumairAK I'm trying to install a grafana instance in my namespace and would like to use the operator for that.

Are there any reasons for making it not cluster wide?


4n4nd (approved):
/lgtm

Issue #1885, opened 2022-04-11
Add both uppercase and lowercase characters for github usernames in OCP groups


HumairAK:
Adding morty to identitatem management has resulted in an issue where github usernames are now converted to OCP usernames with character casings intact. For instance when logging into Morty cluster now, my username is "HumairAK" instead of "humairak". This is an issue because keycloak converted all names to lowercase and OCP groups are case sensitive, therefore rbac gets affected. 

Current suggestion is to parse all github names (there shouldn't be _that_ many, and for those that have both casings, introduce both usernames for a short interim, while we switch auth methods for each cluster, once done, the lowercase versions should be removed.

PR #1928, opened 2022-04-27
Adding github usernames with uppercase letters to OPF groups (excludi…

dystewart:
In response to: #1885 

Adding github handles for those names which have upper and lowercase letters (excluding groups in the OSC cluster)

### DIFF
--- a/cluster-scope/base/user.openshift.io/groups/argocd-admins/group.yaml
+++ b/cluster-scope/base/user.openshift.io/groups/argocd-admins/group.yaml
@@ -7,8 +7,10 @@ users:
   - anishasthana
   - dystewart
   - gregory-pereira
+  - Gregory-Pereira
   - harshad16
   - humairak
+  - HumairAK
   - larsks
   - naved001
   - tumido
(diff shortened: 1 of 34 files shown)


cluster-scope/base/user.openshift.io/groups/argocd-admins/group.yaml
```diff
@@ -7,8 +7,10 @@ users:
   - anishasthana
   - dystewart
   - gregory-pereira
```
Gregory-Pereira, line 9: If we are adding this should we not remove the lower-case versions?

Gregory-Pereira (approved):
/lgtm

cluster-scope/base/user.openshift.io/groups/argocd-admins/group.yaml
HumairAK: not yet because we're still using keycloak for these clusters so when these users log in their usernames will be with lowercase characters. The proper casing doesn't take affect until after we switch to identitatem.

Issue #1844, opened 2022-04-01
Enable PVCs in infra cluster


dystewart:
In order to complete the vault deployment in the infra cluster, we need to set up PVCs on the infra cluster. Perhaps enable local storage on the cluster? 
@larsks @HumairAK

HumairAK:
@larsks made request to nese folks, have not heard back yet 

options: 
1. move to smaug as short term solution 
2. There is another storage driver for k8s called longhorn, gives volume storage using local volumes

SamoKopecky:
@larsks @HumairAK Have we decided on the next step here?

> 



> @larsks made request to nese folks, have not heard back yet
> 
> options:
> 
>     1. move to smaug as short term solution
> 
>     2. There is another storage driver for k8s called longhorn, gives volume storage using local volumes

tumido:
Infra has PVCs now via ODF (data stored at NESE), closing as resolved.

PR #1954, opened 2022-05-02
Adding opf vault instance to smaug cluster

dystewart:
In response to [https://github.com/operate-first/support/issues/298](https://github.com/operate-first/support/issues/298)

Vault was previously added to the infra cluster but due to a [storage issue](https://github.com/operate-first/apps/issues/1844) we're adding vault to smaug in order to move forward with moving secrets away from ksops.


argocd/overlays/moc-infra/applications/envs/moc/smaug/operate-first/opf-vault.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: opf-vault
+spec:
+  project: operate-first
+  source:
+    repoURL: https://github.com/operate-first/apps.git
+    targetRevision: HEAD
+    path: vault/overlays/moc/vault
```
HumairAK, line 10: no such path exists, should be `vault/overlays/moc/smaug` (you'll need to create this)

dystewart: @HumairAK Good catch, updated path

cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
```diff
@@ -110,6 +111,7 @@ resources:
 - ../../../../base/rbac.authorization.k8s.io/clusterrolebindings/scc-privileged-opf-ci-pipelines-aicoe-ci
 - ../../../../base/rbac.authorization.k8s.io/clusterrolebindings/tekton-dashboard
 - ../../../../base/rbac.authorization.k8s.io/clusterrolebindings/vault-server-binding
+- ../../../../base/rbac.authorization.k8s.io/clusterrolebindings/vault
```
HumairAK, line 114: can we create a vault bundle here: https://github.com/operate-first/apps/tree/master/cluster-scope/bundles

And add the namespace/rolebindings to it, then import the bundle instead of the 2 resources separately?

HumairAK:
@dystewart since you're removing: https://github.com/operate-first/apps/blob/master/vault/overlays/moc/infra/kustomization.yaml

This will render this ArgoCD app in an error state: https://github.com/operate-first/apps/blob/master/argocd/overlays/moc-infra/applications/envs/moc/infra/operate-first/opf-vault.yaml

Let's remove this too

dystewart:
@HumairAK removed [https://github.com/operate-first/apps/blob/master/argocd/overlays/moc-infra/applications/envs/moc/infra/operate-first/opf-vault.yaml](https://github.com/operate-first/apps/blob/master/argocd/overlays/moc-infra/applications/envs/moc/infra/operate-first/opf-vault.yaml)
And added vault resource bundle

PR #1959, opened 2022-05-03
Update container image registry in Smaug to use OBC instead of PVC

4n4nd:
Resolves https://github.com/operate-first/support/issues/554


Gregory-Pereira (approved):
/lgtm

PR #1960, opened 2022-05-03
feat(argocd): Enable status badges

tumido:
Hi all, we would benefit from status badges in ArgoCD, can we enable them?

Reference: https://argo-cd.readthedocs.io/en/stable/user-guide/status-badge/


HumairAK (approved):
/lgtm
/approve

PR #1922, opened 2022-04-26
set peribolos-as-a-service argo app for deployment

harshad16:
set peribolos-as-a-service argo app for deployment
Signed-off-by: Harshad Reddy Nalla <hnalla@redhat.com>

Fixes: https://github.com/open-services-group/peribolos-as-a-service/issues/53

PR #1961, opened 2022-05-05
fix: Peribolos service app project server settings

tumido:
Fixes deployment of #1922

image

/cc @harshad16 @Shreyanand


harshad16 (approved):
/lgtm
/approve
thanks

PR #1977, opened 2022-05-12
Switch back to default alertmanager config.

HumairAK:
This commit re-instates the default alertmanager configs on the
emea and moc clusters (balrog, morty, smaug, infra). The other clusters
were never using this config to begin with. This allows use to start
with a clean slate and selectively filter on service via labels and
reduce the influx of alerts in the alerts repo. The commit also moves
the alerts into their cluster's respective overlays for more granular
rules catered specifically for that cluster's needs.

Related: https://github.com/operate-first/operations/issues/468


Gregory-Pereira (approved):
/lgtm

Issue #1958, opened 2022-05-03
Openshift Pipeline Feature flags being overwritten by operator


HumairAK:
To enable KFP with Tekton Backend on Smaug/OSC clusters, we've had to patch the feature flags from openshift-pipelines as per the instructions here: https://github.com/kubeflow/kfp-tekton/blob/master/guides/kfp_tekton_install.md#standalone-kubeflow-pipelines-with-tekton-backend-deployment

We do this in the apps repo [here](https://github.com/operate-first/apps/blob/master/cluster-scope/bundles/openshift-pipelines/kustomization.yaml#L7). Follow this path to find the configmap [here](https://github.com/operate-first/apps/blob/master/cluster-scope/base/core/configmaps/openshift-pipelines-feature-flags/configmap.yaml). 

This bundle is deployed in Smaug and OSC-CL2 (and eventually osc-cl1). 

Notice the cluster-resources argocd app for these 2 clusters keeps reporting this configmap as out of sync: 
https://argocd.operate-first.cloud/applications/cluster-resources-osc-cl2
https://argocd.operate-first.cloud/applications/cluster-resources-smaug

If it doesn't show as out of sync you may need to wait a while as someone may have ran sync action recently. Eventually you'll see it get over written, presumably by the ocp pipelines operator.

HumairAK:
this is resolved, we used tekton config resource to apply changes to the configmap instead.

PR #1989, opened 2022-05-16
Fix the labels of cm feature flags

harshad16:
Fix the labels of cm feature flags and patch strategy

Related-to: https://github.com/operate-first/apps/issues/1958

- the cm feature flag is already installed via openshift pipeline operator https://github.com/tektoncd/pipeline/blob/main/config/config-feature-flags.yaml, it need to be patched
- the label instance is to be patched by the argocd

### DIFF
--- a/cluster-scope/base/core/configmaps/openshift-pipelines-feature-flags/configmap.yaml
+++ b/cluster-scope/base/core/configmaps/openshift-pipelines-feature-flags/configmap.yaml
@@ -4,7 +4,6 @@ metadata:
   name: feature-flags
   namespace: openshift-pipelines
   labels:
-    app.kubernetes.io/instance: default
     app.kubernetes.io/part-of: tekton-pipelines
 data:
   running-in-environment-with-injected-sidecars: 'true'


cluster-scope/bundles/openshift-pipelines/kustomization.yaml
```diff
@@ -4,4 +4,5 @@ kind: Kustomization
 resources:
   - ../../base/core/namespaces/openshift-pipelines
   - ../../base/operators.coreos.com/subscriptions/openshift-pipelines-operator-rh
+patchesStrategicMerge:
```
HumairAK, line 7: why do we need a patch for this?

harshad16: mentioned in the desciption,
it is installed and created by openshift-pipeline operator

harshad16: oh i see, the patch would work on the base and overlays.
we might have to ignore the label somehow tough then.

harshad16:
Ready for review again

Issue #1993, opened 2022-05-17
Add OCP group access to vault restricted by cluster environment


HumairAK:
Add the following groups: 

- vault admin 
- vault emea admin 
- vault moc admin 
- vault osc admin 

Vault admin should have something similar to root access in vault. The remaining groups should have admin level privileges to their respective kv secrets engine. Populate these groups accordingly.

Add vault groups + group aliases that correspond to these ocp groups. 

End result should be that adding anyone to these OCP groups should give them access to the respective secrets engine.

PR #1995, opened 2022-05-17
Change path for slack api url in smaug.

HumairAK:
This is to reflect some changes to how secrets are organized. Instead of
using multiple engines for each clusters, it makes more sense to use one
kv secrets engine and have subdirectories for each env/cluster.

Related: https://github.com/operate-first/apps/issues/1993


dystewart (approved):
I like this secret engine setup better. It's a little more streamlined than separate secrets engines per cluster. 
/lgtm

PR #2010, opened 2022-05-23
Add resources needed to deploy rapidast-operator to smaug cluster

jpweiser:
Adds a catalog source, operator group, and subscription to deploy the RapiDAST operator to the smaug cluster.

### DIFF
--- a/cluster-scope/base/operators.coreos.com/catalogsources/rapidast/catalogsource.yaml
+++ b/cluster-scope/base/operators.coreos.com/catalogsources/rapidast/catalogsource.yaml
@@ -1,0 +1,9 @@
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: CatalogSource
+metadata:
+  name: rapidast-catalog
+  namespace: rapidast
+spec:
+  sourceType: grpc
+  image: quay.io/redhatproductsecurity/rapidast-operator-catalog:v0.0.1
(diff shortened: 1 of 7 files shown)


cluster-scope/base/operators.coreos.com/catalogsources/rapidast/catalogsource.yaml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: CatalogSource
+metadata:
+  name: rapidast-catalog
+  namespace: rapidast
```
HumairAK, line 6: @jpweiser this namespace will also need to be added, you will need to add it to `base` much like the other manifests and then pulled in to the smaug overlay: 

namespace manifests are found here: 
https://github.com/operate-first/apps/tree/master/cluster-scope/base/core/namespaces

you will need [to set a quota for the namespace](https://www.operate-first.cloud/apps/content/cluster-scope/add_resource_quotas.html).

jpweiser: This namespace was added previously https://github.com/operate-first/apps/tree/master/cluster-scope/base/core/namespaces/rapidast

I believe the quota is set [here](https://github.com/operate-first/apps/blob/master/cluster-scope/base/core/namespaces/rapidast/kustomization.yaml#L8)

jpweiser:
/test kustomize-build

I think things are back up, so trying a retest

cluster-scope/base/operators.coreos.com/catalogsources/rapidast/catalogsource.yaml
jpweiser: @HumairAK Does the above comment address your concerns, or did I miss something?

PR #2011, opened 2022-05-23
Added avrajath to cluster admin for curator cluster

AvRajath:
Currently a summer intern in Heidi's Team. Hence making a request.

### DIFF
--- a/cluster-scope/base/user.openshift.io/groups/curator/group.yaml
+++ b/cluster-scope/base/user.openshift.io/groups/curator/group.yaml
@@ -8,3 +8,4 @@ users:
   - matankb
   - leihchen
   - ajinkyababar
+  - avrajath


larsks:
@AvRajath can you provide details on why you need cluster admin access? Most changes to this cluster are made through this apps repository. Would a read-only admin role be sufficient for what you need to do?

larsks:
@AvRajath generally we don't hand out cluster-admin access (for the most part, any actions that require admin access should be performed by pull requests against the appropriate configuration repository). We do have a read-only admin role that might be appropriate. Would that work out for you?

skanthed:
@larsks he don’t need cluster admin access on curator. He already has it. I think this is the wrong PR. This PR was for having access on smaug cluster for specific namespace. @AvRajath right?

gagansk:
@larsks Yes. You can provide @AvRajath cluster-admin access to the `curator` cluster.

larsks:
@skanthed  this PR is for curator cluster admin access (see the modified files).

skanthed:
Yes. I saw it. That’s the reason I am saying. It’s the wrong modification. He is already able to access everything on curator. He needed it on Smaug. Maybe he can close this one and create another one for smaug @AvRajath

larsks:
@AvRajath if you do create a new PR for smaug, then what I said in [an earlier comment](https://github.com/operate-first/apps/pull/2011#issuecomment-1147483654) applies. There is an "sre"  group that has read-only admin access that may be appropriate.

skanthed:
He need curator access in curator group in smaug. No general access.

cluster-scope/base/user.openshift.io/groups/curator/group.yaml
```diff
@@ -8,3 +8,4 @@ users:
   - matankb
   - leihchen
   - ajinkyababar
+  - avrajath
```
dystewart, line 11: @AvRajath Adding github handle here will be sufficient for curator access in smaug

cluster-scope/overlays/prod/moc/curator/groups/cluster-admins.yaml
```diff
@@ -12,3 +12,4 @@ users:
   - hpdempsey
   - leihchen
   - ajinkyababar
+  - avrajath
```
dystewart, line 15: @AvRajath Remove your github handle from this this group

dystewart (changes requested):

dystewart:
@AvRajath can you squash your commits into 1?

AvRajath:
> @AvRajath can you squash your commits into 1?

Squashed into a single commit.

PR #2022, opened 2022-06-01
Fetchdata crd for curator-operator

skanthed:
As we are planning to deploy curator operator MVP in Smaug. This is the CRD for fetching data from koku and pushing it to the Postgres database.


dystewart (approved):
/lgtm

PR #2030, opened 2022-06-03
feat(service-catalog): Add initial catalog definition files

SamoKopecky:
Part of [#578](https://github.com/operate-first/support/issues/578)

- Add initial definition files for the [backstage](https://backstage.io/) static service catalog deployment. 
- These files will be consumed by the [app-config.yaml](https://backstage.io/docs/conf/writing) file, in `catalog.locations`  via a github file link, more info [here](https://backstage.io/docs/features/software-catalog/configuration).
- More information about the format of these files can be found [here](https://backstage.io/docs/features/software-catalog/descriptor-format).

## Content of the service catalog
- For now, the catalog only contains 6 services and looks like this.

image
- Along with the services it also defines a user, group, and system.
- This is just the initial draft of the definition files and will most definitely change in the future.

## Location of these files
- I was also wondering if this is the right location for these files, maybe [this](https://github.com/operate-first/operations) repository would be a better fit, or maybe some other repository would be a better fit, feel free to suggest one.


tumido:
cc @HumairAK @durandom 

Just to clarify, despite the looks of it, these are **NOT** kubernetes manifets or Custom Resources. It's config ingested by backstage on it's own. However the structure favors our kubernetes-like mindset :slightly_smiling_face:

Issue #1996, opened 2022-05-18
Replace all secrets in apps repo with external secrets


HumairAK:
Follow up to: https://github.com/operate-first/support/issues/298

Now that we have a configuration setup for smaug cluster between eso and hashicorp vault. Let's deploy ESO on remaining clusters and replace all the secrets in the apps repo with external secrets. 

For non `kind=secret` resources that have confidential data, we will have to wait for patch operator to see if that can serve as a workaround. 

## Acceptance criteria: 

- [x] ESO deployed on all clusters
- [x] All ESO k8s auth should be able to authenticate with smaug's vault instance 
- [x] Add secret stores to all cluster namespaces that require secret mgmt (https://github.com/operate-first/apps/pull/2124)
- [ ] External secrets for all namespaces using secrets stored in git.

PR #2063, opened 2022-06-16
Add secret stores balrog cluster

HumairAK:
This furthers: #1996 

Trying it out for balrog first. This just adds secret stores, does not change any secrets yet. 

Keeping it in cluster-scope reduces a lot of repetition, and since it only requires that this be done once, it will make it easier to incorporate as part of onboarding in the future(when adding a new namespace) without having to create a separate project/app.


dystewart (approved):
/lgtm

PR #2078, opened 2022-06-17
curator operator cluster level roles and role bindings

skanthed:


cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/curator-operator-manager-rb/clusterrolebinding.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: curator-manager-rolebinding
```
dystewart, line 4: For each role/rolebinding can you include operator in the name?
ie: 
name: curator-operator-manager-rolebinding

dystewart (changes requested):

cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/curator-operator-manager-rb/clusterrolebinding.yaml
skanthed: okay.

skanthed:
@dystewart Updated all the names.

dystewart (approved):

cluster-scope/base/rbac.authorization.k8s.io/clusterroles/curator-operator-manager-role/clusterrole.yaml
```diff
@@ -0,0 +1,148 @@
...
+  - jobs/status
+  verbs:
+  - get
+- apiGroups:
+  - ""
+  resources:
+  - persistentvolumeclaims
+  verbs:
+  - create
+  - delete
```
HumairAK, line 58: @skanthed why does this operator need write permissions to all pvc's on the cluster?

skanthed: @HumairAK Most of the RBAC are auto-generated when we are building an operator using operator-sdk. We have not written any logic that requires access to other PVCs. Even if we don't have access to other PVCs on the cluster, it should be good.

I can try changing it and test it on the curator cluster. I will update the PR accordingly.

PR #2080, opened 2022-06-20
fix(service-catalog): Increase resource quota

SamoKopecky:
A bigger resource quota is needed for the `service-catalog` postgres operator.


tumido (approved):
/lgtm
/approve

PR #2087, opened 2022-06-22
Correct inception grp & add cluster reader to osc.

HumairAK:
This change will fix a typo for a user in the inception admins group on
the osc-cl1 cluster. It will also correct linting disrepencies between
both osc-cl1 and osc-cl2 clusters. Some additional changes are also made
to make the inception setup consistent between the two clusters. The
cluster readers role is also added and patched within the cluster
overlays so as to keep this group's modification on these clusters
isolated from the rest.

Furthers: https://github.com/os-climate/corporate_data_pipeline/issues/55


HeatherAck:
any update on this?

redmikhail:
@HeatherAck It seems that this request has been merged already

PR #2098, opened 2022-06-27
Add bu-cs-book-dev namespace and group to smaug

dystewart:
This namespace is being added for intern @abirami-dhayalan 
CC @hpdempsey


HumairAK (approved):
/lgtm
/approve

PR #2106, opened 2022-06-30
Add Pep to the operate-first group

codificat:
This is to add myself to the `operate-first` group.

The trigger was that I meant to take a closer look at the images(treams) available in Jupyterhub, but I believe this membership will help me in future tasks as well.


tumido (approved):
/lgtm
/approve

PR #2108, opened 2022-06-30
Fix minio operator source

larsks:
The catalog source in the original PR was mis-spelled; the catalog name for
OperatorHub is `operatorhubio-catalog`, not `operathub.io-operators`.
However, it turns out that the `certified-operators` catalog has a more
recent version.


larsks:
`minio-operator` is available from three repositories:

```
$ kubectl get packagemanifest |grep minio
minio-operator-rhmp                                Red Hat Marketplace   315d
minio-operator                                     Certified Operators   315d
minio-operator                                     OperatorHub.io        132d
```

The one in the "Certified Operators" catalog is newer than the one from OperatorHub, so I've updated the subscription to use that one.

HumairAK:
@dystewart fyi

PR #2120, opened 2022-07-06
Added Hakan (hsap) to the admin list for argocd

hakasapl:


hakasapl:
@larsks

larsks (changes requested):
It's worth noting that we almost never interact directly with argocd, other than when diagnosing problems: most  of our interaction is through the git repositories.

hakasapl:
I found it useful to disable argocd on staging temporarily so I could test live changes to the yaml manifest on openshift, and reenable after I was done. That's the only use case I'm looking for, I was usually bothering @naved001 to do this.

I'll change it to my github username. Let me know if you think there's a better way to do the above.

larsks:
/hold

pending response from @HumairAK

HumairAK:
am I correct to assume @hakasapl only requires admin level access to argocd applications under the [mocops project](https://argocd.operate-first.cloud/applications?proj=mocops&sync=&health=&namespace=&cluster=&labels=) ? if so we can set up a new group with such access

PR #2124, opened 2022-07-06
Add secretstores to all namespaces using eso.

HumairAK:
This change accomplishes 2 things:

1) Move all existing secret stores to cluster overlays, and makes their
   usage pattern consistent across clusters.
2) Add secretstores to all namespaces that currently require encrypted
   secrets to be stored in the apps repo.

Majority of the changes consist of (2). The only real affect to the live
environment is renaming some of the secret stores so that they are
consistent with the others (this requires updating the external secrets
themselves). The remainder of the secret stores only add secret stores,
and thus have no real affect to the existing environment. Once secret
stores are in place, we will be able to replace all sops encrypted
secrets with external secrets for all these namespaces.


cluster-scope/overlays/prod/emea/morty/secret-mgmt/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - openshift-config
+  - openshift-ingress
+  - openshift-monitoring
```
harshad16, line 6: is `opf-alertreceiver`  missing from this list intentional ?

cluster-scope/overlays/prod/moc/infra/secret-mgmt/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
...
+resources:
+- argocd
+- opf-alertreceiver
+- idp-auth-primary
+- acm
+- openshift-monitoring
+- openshift-logging
+- openshift-config
+- openshift-ingress
+- democratic-csi
```
harshad16, line 12: can we have these in alphabetical order? might help in future

harshad16 (changes requested):
lgtm
just few questions

cluster-scope/overlays/prod/emea/morty/secret-mgmt/kustomization.yaml
HumairAK: fixed

PR #2128, opened 2022-07-07
Add bundle for strimzi operator deployment. Add strimzi to osc-cl2

redmikhail:
Strimzi subscription as part of https://github.com/os-climate/os_c_data_commons/issues/160

### DIFF
--- a/cluster-scope/overlays/prod/osc/osc-cl2/kustomization.yaml
+++ b/cluster-scope/overlays/prod/osc/osc-cl2/kustomization.yaml
@@ -56,6 +56,7 @@ resources:
 - ../../../../bundles/openshift-pipelines
 - ../../../../bundles/pachyderm-operator
 - ../../../../bundles/reloader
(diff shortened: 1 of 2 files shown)


cluster-scope/overlays/prod/osc/osc-cl2/kustomization.yaml
```diff
@@ -56,6 +56,7 @@ resources:
 - ../../../../bundles/openshift-pipelines
 - ../../../../bundles/pachyderm-operator
 - ../../../../bundles/reloader
+- ../../../../bundles/kafka-operator
```
Gregory-Pereira, line 59: Not certain but I believe the subscription to the  Strimzi operator should also  be added to the overlay for osc-cl2, [as it is  done for smaug](https://github.com/operate-first/apps/blob/master/cluster-scope/overlays/prod/moc/smaug/kustomization.yaml#L113)

PR #2139, opened 2022-07-14
Use tektonconfig instead of configmap.

HumairAK:
Pipelines Operator will overwrite the feature flags configmap with the
values of tektonconfig resource. It is advised that when adding new
flags we use the tektonconfig resources instead.

Related: https://github.com/operate-first/community/issues/206


HumairAK:
Furthers: https://github.com/open-services-group/scrum/issues/3

PR #2147, opened 2022-07-16
WIP: Adding Stackrox to the apps repo

traytorous:
As of right now the stackrox central deployment seems to be completed. However I am having some trouble adding the collector to each node for them to collect data from each node.


traytorous:
Found some errors. Still working on it.

PR #25, opened 2020-11-17
Add additional JH notebook imagestreams

4n4nd:
Addresses https://github.com/operate-first/odh-moc-support/issues/2

PR #2149, opened 2022-07-18
feat(service-catalog): Add peribolos-as-a-service and probot-extensions to service catalog

SamoKopecky:
Part of [#25](https://github.com/operate-first/service-catalog/issues/25)
Add two components to the service catalog:
- `Peribolos-as-a-service`
- `Probot-extensions`
 
A dependency is added:
- `Peribolos-as-a-service` is dependent on `Probot-extensions`

Add tags for `Peribolos-as-a-service` to show which probot extensions it is using.


tumido (approved):
/lgtm
/approve

PR #2162, opened 2022-07-20
Add project admin and view options for strimzi.

HumairAK:
The PR allows for a role in osc-cl2 that will grant all project admins
the ability to create kafka resources in any namespace they have project
admin in. It also allows for a view/admin group, wherein if a user is
added, they have view/admin permissions to openshift-operators
namespace respectively.

These roles are added to base so they may be used elsewhere going
forward.

The commit also sets up kafka structure for osc-cl2 for future
manifests.

Furthers: https://github.com/os-climate/os_c_data_commons/issues/191 
closes: https://github.com/operate-first/support/issues/613

### DIFF
--- a/cluster-scope/base/core/namespaces/kafka/kustomization.yaml
+++ b/cluster-scope/base/core/namespaces/kafka/kustomization.yaml
@@ -1,0 +1,7 @@
+kind: Kustomization
+apiVersion: kustomize.config.k8s.io/v1beta1
+namespace: kafka
+resources:
+    - namespace.yaml
+components:
+    - ../../../../components/project-admin-rolebindings/kafka-admins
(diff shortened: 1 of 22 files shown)


cluster-scope/base/core/namespaces/kafka/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Kustomization
+apiVersion: kustomize.config.k8s.io/v1beta1
+namespace: kafka
```
harshad16, line 3: any particular reason for not using the existing namespace definition of opf-kafka ?
https://github.com/operate-first/apps/tree/master/cluster-scope/base/core/namespaces/opf-kafka

harshad16 (commented):
lgtm in first iteration
one question though

cluster-scope/base/core/namespaces/kafka/kustomization.yaml
harshad16: ack, :) 
so we are not going to maintain it , right :)

PR #2166, opened 2022-07-21
feat(service-catalog): Add k8s plugin annotation to peribolos

SamoKopecky:
Part of https://github.com/operate-first/service-catalog/issues/12

SSIA

### DIFF
--- a/service-catalog/components/peribolos-as-a-service.yaml
+++ b/service-catalog/components/peribolos-as-a-service.yaml
@@ -7,6 +7,7 @@ metadata:
     operate-first.cloud/logo-url: https://raw.githubusercontent.com/operate-first/peribolos-as-a-service/main/static/robot.png
     argocd/app-name: peribolos-as-a-service-smaug
     github.com/project-slug: operate-first/peribolos-as-a-service
+    backstage.io/kubernetes-label-selector: 'app.kubernetes.io/name=peribolos-as-service'
   links:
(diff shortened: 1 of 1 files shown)


service-catalog/components/peribolos-as-a-service.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
     operate-first.cloud/logo-url: https://raw.githubusercontent.com/operate-first/peribolos-as-a-service/main/static/robot.png
     argocd/app-name: peribolos-as-a-service-smaug
     github.com/project-slug: operate-first/peribolos-as-a-service
+    backstage.io/kubernetes-label-selector: 'app.kubernetes.io/name: peribolos-as-service'
```
tumido, line 10: This is not a valid label selector. I think it should be:

```suggestion
    backstage.io/kubernetes-label-selector: 'app.kubernetes.io/name=peribolos-as-service'
```

tumido (changes requested):

service-catalog/components/peribolos-as-a-service.yaml
SamoKopecky: Yea thats right, I just copied it without checking. I will pay more attention next time.

SamoKopecky:
Unblocked, ready for review again.

tumido (approved):
/lgtm
/approve

PR #2168, opened 2022-07-21
Resolving service account naming issue from curator-operator

skanthed:
Dylan trying to resolve the issue that we discussed regarding the leader election role trying to access the service account that didn't exist.


dystewart (approved):
/lgtm

PR #2170, opened 2022-07-22
Add banzaicloud-vault ns to smaug and banzaicloud vault operator subscription

dystewart:
Sandbox namespace for testing new operators etc


HumairAK:
@dystewart is the operator generally installed in its own namespace (as opposed to the global namespace)? 

If so,  how about we create a non-sandbox namespace for this to be deployed in, we can consider it playground until it's deemed otherwise. 

Also -- let's add the operator group and subscription to a bundle.

dystewart:
@HumairAK The operator is deployed in its own namespace and I've renamed the namespace to reflect the use case.
Also bundled the operator resources

Gregory-Pereira (commented):
/lgtm

PR #2179, opened 2022-07-26
feat(schema-store):  Add service account for schema store

mimotej:
This PR creates schema-store service account and adds it to `common` so that it is accessible in all clusters. This service account is needed for `schema-store` cronjob to be able to access `/openapi/v2` endpoints and update schemas of CRDs.
Part of https://github.com/open-services-group/scrum/issues/19 and https://github.com/open-services-group/scrum/issues/6


tumido (changes requested):
please move `base/serviceaccounts` to `base/core/serviceaccounts`

mimotej:
> please move `base/serviceaccounts` to `base/core/serviceaccounts`

Resolved in latest commit

tumido (approved):
/lgtm
/approve

PR #2190, opened 2022-07-27
Increase the ResourceQuota from medium to large for the smart-village-view project

computate:
Hello Operate First team! Could I please increase the ResourceQuota of the smart-village-view project from medium to large? I have large data imports and it seems to be limited by the database, search engine, cluster manager, and application having enough CPU limit to process the data. 2 cpu cores does not seem to be enough to handle the load, so I would like to request the large ResourceQuota to obtain 4 cpu cores please.


computate:
Thank you so much!

PR #2200, opened 2022-07-29
Add admin group to openmetadata namespace on osc-cl2 clusters

redmikhail:
Grant access to  openmetadata namespace to openmetadata-admins group on OSC cluster 2 . Increase PVC to 5Gi from original 1Gi as part of https://github.com/operate-first/operations/issues/489


redmikhail:
@HumairAK

PR #2208, opened 2022-08-01
added stackrox to secret-mgmnt

traytorous:


cluster-scope/overlays/prod/moc/smaug/secret-mgmt/stackrox/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: stackrox  # replace with your namespace
```
HumairAK, line 3: can you remove the comment
> # replace with your namespace

traytorous:
will make the necessary changes to fix the commit history

Issue #2206, opened 2022-08-01
add cooktheryan to access ACM


durandom:
@cooktheryan should be able to access https://multicloud-console.apps.moc-infra.massopen.cloud/ and deploy clusters to AWS

PR #2217, opened 2022-08-03
adding cooktheryan to ACM users and docs

durandom:
fixes https://github.com/operate-first/apps/issues/2206

@Gregory-Pereira @HumairAK can you help me with pointing me in the right direction?


Gregory-Pereira:
Yes, I am pretty sure that `SRE` group gives ACM access.
/lgtm
/approve

PR #2219, opened 2022-08-03
Add various OKD infrastructure

LorbusChris:
This PR adds various resources required for building OKD.

The `okd-centos` namespace is created to host imported centos stream images in an imagestream, and a buildconfig for the `okd-golang-builder` image.
The `okd-golang-builder` image is then used as a base for building the `kvm-device-plugin` image and referencing it in an imagestream.
The `kvm-device-plugin` is then also deployed with a service account. Note that the daemonset is privileged.

All of this is added to the smaug overlay.

### DIFF
--- a/cluster-scope/base/core/namespaces/kvm-device-plugin/kustomization.yaml
+++ b/cluster-scope/base/core/namespaces/kvm-device-plugin/kustomization.yaml
@@ -2,6 +2,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 components:
     - ../../../../components/limitranges/default
     - ../../../../components/resourcequotas/medium
+    - ../../../../components/project-admin-rolebindings/okd-team
 kind: Kustomization
 namespace: kvm-device-plugin
 resources:
--- a/cluster-scope/base/core/namespaces/kvm-device-plugin/kustomization.yaml
+++ b/cluster-scope/base/core/namespaces/kvm-device-plugin/kustomization.yaml
@@ -1,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+components:
+    - ../../../../components/limitranges/default
+    - ../../../../components/resourcequotas/medium
+kind: Kustomization
+namespace: kvm-device-plugin
+resources:
+    - namespace.yaml
--- a/cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
+++ b/cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
@@ -43,6 +43,8 @@ resources:
 - ../../../../base/core/namespaces/manageiq
 - ../../../../base/core/namespaces/neu-cs6620
 - ../../../../base/core/namespaces/od-demo
+- ../../../../base/core/namespaces/okd-team
+- ../../../../base/core/namespaces/okd-wg
 - ../../../../base/core/namespaces/openshift-cnv
 - ../../../../base/core/namespaces/openshift-logging
 - ../../../../base/core/namespaces/openshift-operators
--- a/cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
+++ b/cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
@@ -17,7 +17,6 @@ resources:
 - ../../../../base/apiextensions.k8s.io/customresourcedefinitions/reportapis.curator.operatefirst.io
 - ../../../../base/apiextensions.k8s.io/customresourcedefinitions/reports.batch.curator.openshift.io
 - ../../../../base/apiextensions.k8s.io/customresourcedefinitions/reports.curator.operatefirst.io
-- ../../../../base/apps/daemonsets/kvm-device-plugin
 - ../../../../base/core/namespaces/ai-services-thanos-grpc
 - ../../../../base/core/namespaces/aicoe-meteor
 - ../../../../base/core/namespaces/api-designer
@@ -41,7 +40,6 @@ resources:
 - ../../../../base/core/namespaces/fedora
 - ../../../../base/core/namespaces/koku-metrics-operator
 - ../../../../base/core/namespaces/kubeflow
-- ../../../../base/core/namespaces/kvm-device-plugin
 - ../../../../base/core/namespaces/manageiq
 - ../../../../base/core/namespaces/neu-cs6620
 - ../../../../base/core/namespaces/od-demo
@@ -139,8 +137,6 @@ resources:
 - ../../../../base/rbac.authorization.k8s.io/clusterroles/oauth-token-access-review
 - ../../../../base/rbac.authorization.k8s.io/clusterroles/service-catalog-k8s-plugin
 - ../../../../base/rbac.authorization.k8s.io/clusterroles/tekton-dashboard
-- ../../../../base/rbac.authorization.k8s.io/rolebindings/kvm-device-plugin
-- ../../../../base/rbac.authorization.k8s.io/roles/kvm-device-plugin
 - ../../../../base/storage.k8s.io/csidrivers/org.democratic-csi.nfs
 - ../../../../base/storage.k8s.io/storageclasses/moc-nfs-csi
 - ../../../../base/storage.k8s.io/storageclasses/ocs-external-storagecluster-ceph-rbd
@@ -152,6 +148,7 @@ resources:
 - ../../../../bundles/fybrik
 - ../../../../bundles/jaeger-operator
 - ../../../../bundles/kfp-tekton
+- ../../../../bundles/kvm-device-plugin
 - ../../../../bundles/meteor-operator
 - ../../../../bundles/nfd
 - ../../../../bundles/openshift-pipelines
--- a/cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
+++ b/cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
@@ -17,6 +17,7 @@ resources:
 - ../../../../base/apiextensions.k8s.io/customresourcedefinitions/reportapis.curator.operatefirst.io
 - ../../../../base/apiextensions.k8s.io/customresourcedefinitions/reports.batch.curator.openshift.io
 - ../../../../base/apiextensions.k8s.io/customresourcedefinitions/reports.curator.operatefirst.io
+- ../../../../base/apps/daemonsets/kvm-device-plugin
 - ../../../../base/core/namespaces/ai-services-thanos-grpc
 - ../../../../base/core/namespaces/aicoe-meteor
 - ../../../../base/core/namespaces/api-designer
@@ -40,6 +41,7 @@ resources:
 - ../../../../base/core/namespaces/fedora
 - ../../../../base/core/namespaces/koku-metrics-operator
 - ../../../../base/core/namespaces/kubeflow
+- ../../../../base/core/namespaces/kvm-device-plugin
 - ../../../../base/core/namespaces/manageiq
 - ../../../../base/core/namespaces/neu-cs6620
 - ../../../../base/core/namespaces/od-demo
@@ -137,6 +139,8 @@ resources:
 - ../../../../base/rbac.authorization.k8s.io/clusterroles/oauth-token-access-review
 - ../../../../base/rbac.authorization.k8s.io/clusterroles/service-catalog-k8s-plugin
 - ../../../../base/rbac.authorization.k8s.io/clusterroles/tekton-dashboard
+- ../../../../base/rbac.authorization.k8s.io/rolebindings/kvm-device-plugin
+- ../../../../base/rbac.authorization.k8s.io/roles/kvm-device-plugin
 - ../../../../base/storage.k8s.io/csidrivers/org.democratic-csi.nfs
 - ../../../../base/storage.k8s.io/storageclasses/moc-nfs-csi
 - ../../../../base/storage.k8s.io/storageclasses/ocs-external-storagecluster-ceph-rbd
--- a/cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
+++ b/cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
@@ -43,6 +43,7 @@ resources:
 - ../../../../base/core/namespaces/manageiq
 - ../../../../base/core/namespaces/neu-cs6620
 - ../../../../base/core/namespaces/od-demo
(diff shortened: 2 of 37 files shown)


LorbusChris:
@durandom @HumairAK thanks for the help on the first round of PRs. This PR here now adds some additional resources.

The centos namespace is created to host imported centos stream images in an imagestream, and a buildconfig for the okd-golang-builder.
The okd-golang-builder is then used as a base for building the kvm-device-plugin image and referencing it in an imagestream.
The kvm-device-plugin is then also deployed with a service account. Note that the daemonset is privileged.
All of this is added to the smaug overlay.

I'm not exactly sure the dir structure (in `components/`) is correct, please let me know if I should better move this elsewhere.

lmzuccarelli (commented):
Tested locally on OKD 4.10 cluster all builds passing and daemonset deployed

lmzuccarelli:
@LorbusChris - great work

cluster-scope/components/centos/rolebinding.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: system:image-puller
+  namespace: centos
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:image-puller
+subjects:
+- kind: ServiceAccount
+  name: builder
+  namespace: kvm-device-plugin
```
LorbusChris, line 15: @aleskandro could we put `'*'` in here to allow builders from all namespaces to pull the image?

aleskandro: yes, it should work

aleskandro: No, we can't do it this way. We could assign this role to all ServiceAccounts in the cluster, but not all the ServiceAccounts with name "builder". I'm looking for some other solution to achieve what you say.

cluster-scope/components/centos/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+---
```
HumairAK, line 1: The `cluster-scope` folder is for resources that project admins are not allowed to deploy.

The folders `centos` and `kvm-device-plugin` largely contain `namespace` scoped resources that a project admin can deploy, so they should be moved to the root of the `apps` repo.

`Namespace` resources are cluster scoped, so they should be included in the `cluster-scope/base/core/namespaces` folder. 

The `daemonset` requires privileged access, so that too should be in `cluster-scope` (admittedly cluster-scope is misleading here, as `daemonset` is namespace scoped). 

The rest can be moved to the root of the apps repo, please follow the following structure: 


```
.
└── apps
    ├── centos
    │   ├── base
    │   │   └── kustomization.yaml
    │   └── overlays
    │       └── smaug
    │           └── kustomization.yaml
    ├── cluster-scope
    │   └── base
    │       ├── apps
    │       │   └── daemonsets
    │       │       └── kvm-device-plugin
    │       │           └── kustomization.yaml
    │       └── core
    │           └── namespaces
    │               ├── centos
    │               └── kvm-device-plugin
    └── kvm-device-plugin
        ├── base
        │   └── kustomization.yaml
        └── overlays
            └── smaug
                └── kustomization.yaml
```

Include any `cluster-scope/base/*` resources that need to be added to the `smaug` cluster [here](https://github.com/operate-first/apps/blob/master/cluster-scope/overlays/prod/moc/smaug/kustomization.yaml).

cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
```diff
@@ -138,6 +140,8 @@ resources:
 - ../../../../base/storage.k8s.io/storageclasses/moc-nfs-csi
 - ../../../../base/storage.k8s.io/storageclasses/ocs-external-storagecluster-ceph-rbd
 - ../../../../base/user.openshift.io/groups/cluster-admins
+- ../../../../base/user.openshift.io/groups/okd-team
+- ../../../../base/user.openshift.io/groups/okd-wg
```
HumairAK, line 144: Can you add the groups here instead [1](https://github.com/operate-first/apps/blob/master/cluster-scope/overlays/prod/common/kustomization.yaml)

cluster-scope/overlays/prod/moc/smaug/kustomization.yaml
```diff
@@ -156,6 +160,8 @@ resources:
 - ../../../../bundles/service-catalog
 - ../../../../bundles/tekton-chains
 - ../../../../bundles/volsync
+- ../../../../components/centos
+- ../../../../components/kvm-device-plugin
```
HumairAK, line 164: remove these as per above comments about restructuring

cluster-scope/components/kvm-device-plugin/role.yaml
```diff
@@ -0,0 +1,15 @@
...
+  namespace: kvm-device-plugin
+rules:
+- resources:
+  - securitycontextconstraints
+  apiGroups:
+  - security.openshift.io
+  verbs:
+  - use
+  resourceNames:
+  - privileged
```
HumairAK, line 15: this looks like it's not going to work since you're using `role` and `rolebindings` but `securitycontextconstraints` is a clusterwide resource

LorbusChris: Are you sure about this? This is how we're doing it on OpenShift's Prow too (https://github.com/openshift/release/blob/master/clusters/build-clusters/02_cluster/kvm-device-plugin/03-sa-and-role.yaml). I think a Role and RoleBinding does suffice since it only applies to the kvm-device-plugin namespaced here.

cluster-scope/components/centos/rolebinding.yaml
LorbusChris: deferring this as not blocking

cluster-scope/components/kvm-device-plugin/role.yaml
HumairAK: @LorbusChris [you're right](https://docs.openshift.com/container-platform/4.9/authentication/managing-security-context-constraints.html#role-based-access-to-ssc_configuring-internal-oauth), I guess `scc` are a special case here due to the verb `use` ?  :man_shrugging: 

still...

> Because RBAC is designed to prevent escalation, even project administrators are unable to grant access to an SCC. By default, they are not allowed to use the verb use on SCC resources, including the restricted SCC.

As such can we move the scc roles/rolebindings to `cluster-scope` folder as well? You can create: 

*  `apps/cluster-scope/base/rbac.authorization.k8s.io/roles/kvm-device-plugin/`
* `apps/cluster-scope/base/rbac.authorization.k8s.io/rolebinding/kvm-device-plugin/` 

specify the `kvm-device-plugin` within the kustomization.yaml in those directories and then pull them into the smaug overlay along with the other resources.

cluster-scope/components/centos/kustomization.yaml
LorbusChris: I've renamed the ns to `okd-centos` to clarify that it is maintained by the OKD team

LorbusChris:
Thanks for the review @HumairAK! I think I've addressed everything, but please take another look when you get a chance.

HumairAK:
Converted kvm-device-plugin cluster resources into a bundle and added argocd project `okd-wg` visible to the `okd-wg` ocp team (can be changed to another team by updating the AppProject resource). Also add the project apps `okd-centos` and `kvm-device-plugin` for smaug cluster.

cluster-scope/base/core/namespaces/kvm-device-plugin/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+components:
+    - ../../../../components/limitranges/default
+    - ../../../../components/resourcequotas/medium
```
HumairAK, line 4: @LorbusChris would it make sense to give the `okd-wg` or `okd-team` ocp group access to this namespace, to verify the state of this daemonset, and ensure it's working as intended?

argocd/overlays/moc-infra/projects/okd-wg.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: argoproj.io/v1alpha1
+kind: AppProject
+metadata:
+  name: okd-wg
+  labels:
+    project-template: global
+spec:
+  destinations:
+    - namespace: 'okd-*'
```
LorbusChris, line 9: For now, okd-wg members should only get access to the okd-wg ns, while okd-team should get access to both okd-wg and okd-wg.
Project-admin of centos and kvm-device-plugin ns should also be with okd-team for now, not okd-wg.

cluster-scope/base/core/namespaces/kvm-device-plugin/kustomization.yaml
LorbusChris: yes, okd-team would make sense!

LorbusChris:
@HumairAK thank you so much for looking into this!

/hold
per unresolved conversations

argocd/overlays/moc-infra/projects/okd-wg.yaml
LorbusChris: so essentially s/okd-wg/okd-team/g here

HumairAK: updated

cluster-scope/base/core/namespaces/kvm-device-plugin/kustomization.yaml
HumairAK: updated

LorbusChris:
LGTM!

/hold cancel

HumairAK:
@LorbusChris the argocd apps were deployed, and have deployed the committed resources. 

Please see if you can access the following argocd apps: 
- [kvm-device-plugin](https://argocd.operate-first.cloud/applications/kvm-device-plugin-smaug?resource=)
- [okd-centos](https://argocd.operate-first.cloud/applications/okd-centos-smaug?resource=sync%3AOutOfSync)


Also see if the relevant parties can access these namespaces: 
- https://console-openshift-console.apps.smaug.na.operate-first.cloud/k8s/ns/kvm-device-plugin/pods
- https://console-openshift-console.apps.smaug.na.operate-first.cloud/k8s/ns/okd-centos/pods
- https://console-openshift-console.apps.smaug.na.operate-first.cloud/k8s/ns/okd-wg/pods
- https://console-openshift-console.apps.smaug.na.operate-first.cloud/k8s/ns/okd-team/pods

I'm seeing some errors in `kvm-device-plugin`, the build seems to be failing though I didn't take a very close look. Anyone in the `okd-team` group should be able to take a look and diagnose the issue, if elevated privileges are required, let us know !

LorbusChris:
I do not have access to the argocd apps (`An authentication error occurred.`)

While I do have access to the `kvm-device-plugin`, `okd-team` and `okd-wg` ns,
I am unable to access `okd-centos`:
`pods is forbidden: User "LorbusChris" cannot list resource "pods" in API group "" in the namespace "okd-centos"`

The `kvm-device-plugin` build (and therefore also all DaemonSet deployment attempts) failed due to the reference to the wrong namespace in the build's FROM directive (`centos` instead of `okd-centos`). After fixing that manually and triggering a new build, I got:
```
Error creating build pod: pods "kvm-device-plugin-5-build" is forbidden: exceeded quota: medium, requested: limits.cpu=500m, used: limits.cpu=2, limited: limits.cpu=2
```
It looks like the DS tries to create 4 pods, using up all the quota. It wasn't obvious to me how to configure the number of those pods, so in order to unblock a new build, I've deleted the DS manually for now.

The new build was then able to run, but failed with:
```
error: build error: failed to pull image: After retrying 2 times, Pull image still failed due to error: initializing source docker://image-registry.openshift-image-registry.svc:5000/okd-centos/centos:okd-golang-builder: reading manifest okd-golang-builder in image-registry.openshift-image-registry.svc:5000/okd-centos/centos: manifest unknown: manifest unknown
```
This points to another issue with the `okd-golang-builder` build in `okd-centos`, which I don't have access to, as mentioned.

LorbusChris:
I've filed https://github.com/operate-first/apps/pull/2262 to address this.
Will still require some help with the ArgoCD access issue that is not addressed by that.

lmzuccarelli:
I also don't have access to the  argocd apps (An authentication error occurred.)

HumairAK:
Ah sorry - can you also add all the members here: 
https://github.com/operate-first/common/blob/main/github-config.yaml#L18

This will make everyone a github member of this org, and also enable authentication against argocd, and also some other [features](https://github.com/operate-first/common/blob/main/docs/add_gh_member_and_access.md#become-a-github-member)

HumairAK:
@LorbusChris 

> I've deleted the DS manually for now.

Note that this resource is being managed [by a separate app](https://github.com/operate-first/apps/blob/master/argocd/overlays/moc-infra/applications/envs/moc/smaug/cluster-management/cluster-resources.yaml). We use this app to manage all resources (yesterday we had this app off auto sync as we were testing some stuff, but it's back on now). So it's likely the DS was redeployed. If you'd like to be able to see [this app in argocd](https://argocd.operate-first.cloud/applications/cluster-resources-smaug?resource=sync%3AOutOfSync&conditions=false), you can request admin read only access by adding your self [here](https://github.com/operate-first/apps/blob/master/cluster-scope/base/user.openshift.io/groups/argocd-readonly/group.yaml). 

If you're expecting to update the DS a lot on live, I can keep this app off auto sync for now, but ideally these changes should be delegated to git so that we can keep them in sync with live, and turn on auto sync.

LorbusChris:
@HumairAK unfortunately even the large resourcequota for `kvm-device-plugin` is not enough - since the DS will deploy to each and all nodes, exhausting the quota, there is nothing left for the build to run :/

Is see two options:
- increase quota even more (depending on the number of nodes in the cluster, there could be a lot more quota required)
- limit the deployment of the DS to a subset of nodes with an annotation

WDYT?

HumairAK:
@LorbusChris do you have an idea of what the resource requirement will be? We can [allocate a custom quota](https://www.operate-first.cloud/apps/content/cluster-scope/add_resource_quotas.html#adding-a-custom-resource-quota-to-a-namespace).

HumairAK:
@LorbusChris also I noticed the DS doesn't specify any resources, i.e. the current resources being set are from the default limitrange, if each DS pod requires much less resources, we can also specify a lower amount.

PR #2226, opened 2022-08-03
Adjust osc-cl2 argocd project structure.

HumairAK:
Currently we put all argocd apps under the osc argocd project. This was
done to allow some separate between osc argocd apps, and the other
argocd apps deployed within the operate-first cloud. This was useful for
provisioning access to only osc argocd project. But we now have a
situation where we have an existing argocd project (fybrik), that we
would like to leverage for fybrik apps deployed on osc-cl2. So for now
we will make an exception and have fybrik apps within argocd be under
the fybrik argocd project until a more robust structure/format is found.


harshad16 (approved):
/lgtm
/approve

PR #2245, opened 2022-08-10
hive "ingest" schema on CL2

erikerlandson:
Declares the new catalog: `osc_datacommons_hive_ingest`, and a dedicated schema `ingest` intended for use with the osc-ingest-tools method `fast_pandas_ingest_via_hive`

also declares a privileged user group `hive_ingest` that has access to this schema.


erikerlandson:
xref:  https://github.com/os-climate/os_c_data_commons/issues/189

MichaelTiemannOSC:
Much appreciated if this can be done today...it strongly impacts the pipeline rewrites I need/want to do.

PR #2248, opened 2022-08-11
Add Pep to the odh-admin group

codificat:
This is to add myself to the `odh-admin` group, with the goal of getting access to the osc cluster to use it for CD for the ODH CNBi functionality.

Also taking the opportunity to sort the user list in that group alphabetically.

### DIFF
--- a/cluster-scope/base/user.openshift.io/groups/odh-admin/group.yaml
+++ b/cluster-scope/base/user.openshift.io/groups/odh-admin/group.yaml
@@ -3,17 +3,18 @@ kind: Group
 metadata:
   name: odh-admin
 users:
-  - anishasthana
+  - 4n4nd
   - accorvin
-  - maulikjs
-  - vpavlin
-  - lavlas
-  - LaVLaS
+  - anishasthana
+  - codificat
   - crobby
-  - nakfour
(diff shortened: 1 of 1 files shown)


codificat:
/cc @tumido 
as discussed in the services call earlier today

cluster-scope/base/user.openshift.io/groups/odh-admin/group.yaml
```diff
@@ -3,17 +3,18 @@ kind: Group
...
-  - crobby
+  - maulikjs
   - nakfour
   - rimolive
-  - humairak
-  - HumairAK
-  - 4n4nd
   - tumido
-  - hemajv
+  - vpavlin
```
HumairAK, line 20: @codificat can I ninja your edit and request you remove the following: 

- vpavlin
- maulikjs
- nakfour
- 4n4nd

HumairAK: if you don't mind - feel free to add as a separate commit

codificat: Sure! done (new commit added)

PR #2259, opened 2022-08-17
Add pep to osc-cl1 odh-admin group

codificat:
This is to add myself to the `odh-admin` group, with the goal of getting access to the osc cluster 1 to use it for CD for the ODH CNBi functionality.

Also taking the opportunity to sort the user list in that group alphabetically.

It seems that #2248 did not have the effect I expected - I modified a different odh-admin group there. Let me know if I should revert that change, thanks.

### DIFF
--- a/cluster-scope/overlays/prod/osc/osc-cl1/groups/odh-admin.yaml
+++ b/cluster-scope/overlays/prod/osc/osc-cl1/groups/odh-admin.yaml
@@ -3,13 +3,14 @@ kind: Group
 metadata:
   name: odh-admin
 users:
-  - redmikhail
-  - HumairAK
-  - erikerlandson
   - 4n4nd
-  - rimolive
   - caldeirav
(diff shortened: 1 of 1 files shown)


HumairAK:
@codificat sorry for the confusion, as the groups for os-climate are handled separate than the rest of the community cloud, if you do not require access to smaug odh then feel free to remove yourself. But we always welcome additional help/support in that area if you are ever interested so I personally am OK with it. 

/lgtm
/approve

codificat:
> if you do not require access to smaug odh then feel free to remove yourself. But we always welcome additional help/support in that area if you are ever interested so I personally am OK with it.

Ok, thanks! I do not currently require access, but I will keep myself there and I will try to help if I can (and will do my best not to break anything).

> /lgtm /approve

thanks - it seems that prow/tide is not working for some reason, I'm looking at that...

PR #2287, opened 2022-08-24
updating documentation for getting acm acess

Gregory-Pereira:
Came up in the process of resolving ACM access support issue.

/cc @HumairAK


HumairAK:
@Gregory-Pereira can you also update the docs for argocd with the same req: https://www.operate-first.cloud/apps/content/argocd-gitops/get_access_to_argocd.html

PR #2286, opened 2022-08-24
feat: Install ODF in external mode to Infra cluster

tumido:
Part of: https://github.com/operate-first/operations/issues/475
Depends on: https://github.com/operate-first/apps/pull/2296 https://github.com/operate-first/apps/pull/2360

/hold

Requires credentials to be created in Vault first

/cc @larsks @HumairAK

PR #2296, opened 2022-08-25
feat: Update vault admins

tumido:
Required for: https://github.com/operate-first/apps/pull/2286#discussion_r954052743

/cc @HumairAK


tumido:
`kustomize-build` due to deleted `ceph` folder - this still counts as a change for the script, hence it tries to check it. This should be improved in the script. For this PR we can safely override it

/override kustomize-build

cluster-scope/overlays/prod/moc/smaug/groups/vault-admins.yaml
```diff
@@ -7,5 +7,6 @@ users:
   - dystewart
   - Gregory-Pereira
   - harshad16
-  - 4n4nd
   - SamoKopecky
+  - tumido
+  - durandom
```
durandom, line 12: 😍

PR #2303, opened 2022-08-25
Conver osc-cl1 secrets to use ESO.

HumairAK:
In this change apps superset and trino dev in osc-cl1 have their secrets
added to vault and removed from git, with external secrets taking their
place instead.

Furthers: https://github.com/operate-first/apps/issues/1996


kfdefs/overlays/osc/osc-cl1/trino-dev/kustomization.yaml
```diff
@@ -3,10 +3,10 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
 resources:
+  - configs
+  - externalsecrets
+  - hive-metastores
   - kfdef.yaml
   - secure-route.yaml
-  - secrets
-  - hive-metastores
-  - configs
 generatorOptions:
```
Gregory-Pereira, line 11: I thought generatorOptions pertained to secretGenerators. Does it also extend to configurations for configmapGenerators? If so it makes sense to keep this, but if not it should be deleted, no?

kfdefs/overlays/osc/osc-cl1/superset/externalsecrets/superset.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: external-secrets.io/v1beta1
```
Gregory-Pereira, line 1: Should this fine not be named `superset-custom.yaml` to keep the filename and secret names the same?

Gregory-Pereira (commented):
/lgtm, just two questions

kfdefs/overlays/osc/osc-cl1/trino-dev/kustomization.yaml
HumairAK: yes, you'll see configmapgenerator add a hash (config-maps are technically supposed to be immutable, as initially intended anyway), we use this option to disable it

erikerlandson (approved):

kfdefs/overlays/osc/osc-cl1/superset/externalsecrets/superset.yaml
HumairAK: yep yep tyty

kfdefs/overlays/osc/osc-cl1/trino-dev/kustomization.yaml
Gregory-Pereira: Thanks for the explanation! In that case ...
/lgtm
/approve

PR #2311, opened 2022-08-26
set uniq table loc and increase max partitions

erikerlandson:
- increase maximum partitions from 100 to 300
- set `iceberg.unique-table-location=true`

note: `unique-table-location` defaults to `true` in newer releases, but it does no harm.


erikerlandson:
cc @MichaelTiemannOSC

PR #2313, opened 2022-08-26
riskthinking catalog

erikerlandson:
Defines a catalog using trino BigQuery connector to the RiskThinking.ai BigQuery data.


erikerlandson:
@HumairAK  @rimolive 
We are going to need a new secret for injecting `${ENV:OSC_RISKTHINKING_BQ_SA}`

I am not sure exactly what deployment needs to see this env var.  Possibly just the trino coordinator and workers.
I am assuming we are not going to be creating a hive metadata-store for this catalog since it is federating a BigQuery db.

I have set a `/hold` on this PR until we resolve where to inject.

I have the json file that will constitute the value of this secret. I can send it to you securely if you need to do create the secret contents.  note, these contents are expected to be base64 encoded:
https://trino.io/docs/current/connector/bigquery.html#requirements

erikerlandson:
xref:
https://github.com/operate-first/apps/pull/2325
https://github.com/operate-first/odh-manifests/pull/37

These should be merged and verified before merging this

Issue #2316, opened 2022-08-28
MOC Infra Web terminal failure


Gregory-Pereira:
After MOC-Infra upgrade some deployments are failing. One example of this is the web-terminal, specifically the 'devworkspace-controller` pod in the `openshift operators` namespace. Its continually crashlooping; I pulled this from the error logs: ```{"level":"error","ts":1661711251.0421686,"logger":"cmd","msg":"","error":"no matches for kind \"MutatingWebhookConfiguration\" in version \"admissionregistration.k8s.io/v1beta1\"","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/remote-source/app/vendor/github.com/go-logr/zapr/zapr.go:128\nmain.main\n\t/remote-source/app/cmd/manager/main.go:136\nruntime.main\n\t/opt/rh/go-toolset-1.13/root/usr/lib/go-toolset-1.13-golang/src/runtime/proc.go:203"} ```

Gregory-Pereira:
Seems this API version does not exist on infra: 
``` 
$  oc api-versions | grep "admissionregistration.k8s.io/"
admissionregistration.k8s.io/v1
``` 
We should expect to see both API versions listed:
```
admissionregistration.k8s.io/v1
admissionregistration.k8s.io/v1beta1
```

tumido:
I think this is due to missed upgrade window on the operator available upgrade path. We're attempting upgrade from 1.1.1 to 1.5.1 after all... In these cases there's no harm in simply deleting the subscription resource and recreating it. It will pull the latest CSV and run from there. There's no running web terminal at this point so it causes no outage on the cluster. It seems the problem has resolved itself by that.

PR #2317, opened 2022-08-28
updating verison of k8s webhook configurations to v1beta1

Gregory-Pereira:

Think this should address #2316. Can I get @tumido or @HumairAK to take a look and verify this is the correct solution for enabling that api version in infra?


tumido:
Resolved by removing the operator's subscription and resyncing it. This PR is no longer needed.

PR #2320, opened 2022-08-29
Pre commit & hooks and image version updates

VannTen:
This follow the discussion in https://github.com/thoth-station/kebechet/issues/588
See also https://docs.google.com/document/d/16EIdTs12apkjuNlgBCMa0gQ2Gd0CFu9wn-N9GQmwTdw/edit#heading=h.pf1xbw314osl

@harshad16

This is a rough draft.

Can we use tekton pipeline in prowjobs ? It seems it's possible from reading
some of Prow documentation, but I'm not sure of our instance capabilities ?

### DIFF
--- a/prow/overlays/smaug/config.yaml
+++ b/prow/overlays/smaug/config.yaml
@@ -372,7 +372,67 @@ tide:
           probot-template:
             skip-unknown-contexts: false
 
+presets:
+- labels:
+  preset-xdg-cache: "true"
+  env:
+    - name: XDG_CACHE_HOME
+      value: /var/cache/xdg
+  volumesMount:
+    - name: cache
+      mountPath: /var/cache/xdg
+  volumes:
+    - name: cache
+      persistentVolumeClaim:
+        claimName: ci-xdg-cache
+
+presubmits:
+  thoth-station:
+  - name: pre-commit # This should be overridable(by in_repo_config). Check https://github.com/kubernetes-sigs/prow/issues/18
+    labels:
+      preset-xdg-cache: "true"
+    decorate: true
+    skip_report: false
+    always_run: true
+    context: aicoe-ci/prow/pre-commit
+    spec:
+      containers:
+        - image: quay.io/thoth-station/thoth-precommit-py38:latest
+          command:
+            - "pre-commit"
+            - "run"
+            - "--all-files"
+      resources:
+        requests:
+          memory: "150Mi"
+          cpu: 0.5 # We don't set a CPU limit since it's a compressible resource
+        limits:
+          memory: "300Mi"
+# overrides would look like
+#presubmits:
+#- name: pre-commit
+#  labels:
+#    preset-xdg-cache: "true"
+#  decorate: true
+#  skip_report: false
+#  always_run: true
+#  context: aicoe-ci/prow/pre-commit
+#  spec:
+#    containers:
+#      - image: quay.io/thoth-station/thoth-precommit-py38:latest
+#        command:
+#          - "pre-commit"
+#          - "run"
+#          - "--from-ref origin/HEAD"
+#          - "--to-ref HEAD"
+
 periodics:
+  - name: pre-commit-hooks-update
+    interval: 7d
+    decorate: true
+    annotations:
+      description: Updates pre-commit image version and pre-commit hooks
+    spec: {} # TODO. Ideally a tekton pipeline if that's possible (gh query for all concerned repo + one updater per repo ?)
   - name: periodic-close-rotten
     interval: 3h
     decorate: true
--- a/prow/overlays/smaug/config.yaml
+++ b/prow/overlays/smaug/config.yaml
@@ -432,7 +432,11 @@ periodics:
     decorate: true
     annotations:
       description: Updates pre-commit image version and pre-commit hooks
-    spec: {} # TODO. Ideally a tekton pipeline if that's possible (gh query for all concerned repo + one updater per repo ?)
+    spec: # TODO. Ideally a tekton pipeline if that's possible (gh query for all concerned repo + one updater per repo ?)
+    initContainers: #PROTO
+      - name: repo-lister
+        image: github-cli
+        command:  gh search repos --owner thoth-station --update ">$(date -Idate -d '2 months ago')"
   - name: periodic-close-rotten
     interval: 3h
     decorate: true
--- a/prow/overlays/smaug/cache-storage.yaml
+++ b/prow/overlays/smaug/cache-storage.yaml
@@ -1,0 +1,9 @@
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: ci-xdg-cache
+spec:
+  capacity: 20Gi
+  accessModes:
+    - ReadWriteMany


Gregory-Pereira:
/cc @Gregory-Pereira 
will take a more in depth look later and see if I can test some of this stuff, but I dont immediately see anything wrong with these manifests.

VannTen:
Do you know if I can use tekton pipeline with that prow install ? I can
hack together the update-hook periodics as a single shell script, but
it would be cleaner with reusable tektons tasks (gh querying and pr
notably)

VannTen:
Also, if we **can** use a tekton pipeline, I just found out https://github.com/tektoncd/catalog
(TL;DR: a library of tasks and pipelines for common usecases) which provides:

- github pr creation
- git repo pulling
That would leave only github api query to get the repos to update, a pre-commit
update task and the plumbing to run a matrix for each repo.

prow/overlays/smaug/config.yaml
```diff
@@ -372,7 +372,71 @@ tide:
...
+  volumesMount:
+    - name: cache
+      mountPath: /var/cache/xdg
+  volumes:
+    - name: cache
+      persistentVolumeClaim:
+        claimName: ci-xdg-cache
+
+presubmits:
+  thoth-station:
```
harshad16, line 390: we dont need this function here,
we tend to use the in_repo config with .prow.yaml for this.

prow/overlays/smaug/config.yaml
```diff
@@ -372,7 +372,71 @@ tide:
           probot-template:
             skip-unknown-contexts: false
 
+presets:
+- labels:
```
harshad16, line 376: what would this be used for ?

prow/overlays/smaug/config.yaml
```diff
@@ -372,7 +372,71 @@ tide:
...
+#          - "--from-ref origin/HEAD"
+#          - "--to-ref HEAD"
+
 periodics:
+  - name: pre-commit-hooks-update
+    interval: 7d
+    decorate: true
+    annotations:
+      description: Updates pre-commit image version and pre-commit hooks
+    spec: # TODO. Ideally a tekton pipeline if that's possible (gh query for all concerned repo + one updater per repo ?)
```
harshad16, line 435: how about we reuse the script in pipeline-helper with more modification
https://github.com/thoth-station/pipeline-helpers/blob/master/bump_base_image_version.py
and use it here with the image?

harshad16 (changes requested):
thanks for the initial work

VannTen:
On Wed, Aug 31, 2022 at 07:59:25AM -0700, Harshad Reddy Nalla wrote:
> +presubmits:
> +  thoth-station:
>
> we dont need this function here,
> we tend to use the in_repo config with .prow.yaml for this.

The idea was to use a default baseline, which will suffice for most of
the repos, and only deviates if necessary. (This relies on the idea that
we can override this config from inside .prow.yaml files, I haven't been
able to find documentation to see if it can work or not yet)

> +presets:
> +- labels:
> what would this be used for ?

That's for defining common stuff (namely, volume mounts and env
vars)which will be applied to every ProwJob adding the label
`preset-xdg-cache: "true"`. This is for using the precommit cache
instead of reinstalling the hooks on each time. Hooks are shareable
between runs and codebases, so we need to share the cache.
The preset just makes it more easy to use.

This is intended in combination with the "override" method I alluded to
in the previous point. For override to be painless, it's better to only
add a label than the whole thing (volume, mount, env)

> >  periodics:
> +  - name: pre-commit-hooks-update
> +    interval: 7d
> +    decorate: true
> +    annotations:
> +      description: Updates pre-commit image version and pre-commit hooks
> +    spec: # TODO. Ideally a tekton pipeline if that's possible (gh query for all concerned repo + one updater per repo ?)
> 
> how about we reuse the script in pipeline-helper with more modification
> https://github.com/thoth-station/pipeline-helpers/blob/master/bump_base_image_version.py
> and use it here with the image?

What for ? AFAICT this basically regex over a yaml file, and for this
part we can just use the pre-commit image with `precommit
autoupdate`.

The tricky part is more getting all the repos we want ('updated since'
if we want "active" repos) and feed that into the 'clone, precommit
update, pr'). Not that hard, it's just going to be a bit more ugly
without using tekton pipelines and doing it all in a pod.

VannTen:
notes to self (and anyone interested):

- gh cli dev team does not support an official docker image and is unwilling to
  do so (see https://github.com/cli/cli/discussions/3698#discussioncomment-785838 for details). Additionnaly, the non-interactive usability is not super great, in my experience.
  Another client to the github api would be probably easier on the long term.
- Alternative to the last commit: use all-repos with the pre-commit autofixer.
  Basically what I did manually last time. We can feed it a json files for the
  source.

harshad16:
> The idea was to use a default baseline, which will suffice for most of the repos, and only deviates if necessary. (This relies on the idea that we can override this config from inside .prow.yaml files, I haven't been able to find documentation to see if it can work or not yet)

The prow CI is serving multiple org, and not all require the pre-commit checks for their PRs.
So that is the reason, the in_repo_config is switched on, so user can pick their CI checks. 
we weren't planning for this change, right?
we were looking for ways to do: 
1. auto update the `pre-commit autoupdate` 
2. auto update the base image for the repo that has pre-commit image in .prow.yaml.

> That's for defining common stuff (namely, volume mounts and env vars)which will be applied to every ProwJob adding the label `preset-xdg-cache: "true"`. This is for using the precommit cache instead of reinstalling the hooks on each time. Hooks are shareable between runs and codebases, so we need to share the cache. The preset just makes it more easy to use. This is intended in combination with the "override" method I alluded to in the previous point. For override to be painless, it's better to only add a label than the whole thing (volume, mount, env)

ack, we wouldn't need this as we were not planning on execute pre-commit checks through this way.
we can plan for this in a separate task, if this help in larger scale.

> What for ? AFAICT this basically regex over a yaml file, and for this part we can just use the pre-commit image with `precommit autoupdate`. The tricky part is more getting all the repos we want ('updated since' if we want "active" repos) and feed that into the 'clone, precommit update, pr'). Not that hard, it's just going to be a bit more ugly without using tekton pipelines and doing it all in a pod.

It was in reference to the autoudpate of the baste image for the precommit images.
we dont need to look for the active repo anymore, that was only suggestion tobe done in case this feature was to be in aicoe-ci.

VannTen:
Follow-up/summary of the [sig-devsecops call yesterday on this](https://docs.google.com/document/d/16EIdTs12apkjuNlgBCMa0gQ2Gd0CFu9wn-N9GQmwTdw/edit?skip_itp2_check=true&pli=1#heading=h.m7sbzvfi1672): 

1. For image updating/hooks updating:
    - how to clone the repo ? (extrarefs/clonerefs ?)
    - do periodic jobs run per repo ?
      (Two things makes me thinks that they don't:
        - [the documentation](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables)
          state that there is no REPO_* env vars for periodics jobs.
        - our own periodics for the issue lifecycle stuff include github API
          queries specifying the org. Would'nt not be needed if this was per
          repo.
    - if we do it ourselves, we should use a Github API client (not gh)
      (@harshad16 if you can add a link to the stuff we use ?)

2. For image updating: 
  - let's see if we can write something  https://github/thoth-station/pipelines-helpers
  - reuse .aicoe-ci.yaml updater
  - (also: we had thought about simply using `latest` as we control the image,
    let's not forget that as an alternative)

3. For hooks updating:
  - versions of hooks depends on pre-commit version ?
    No, it just gets the latest git tag on the hook repository. [doc](https://pre-commit.com/#updating-hooks-automatically)


Overall the first item (basically: logic to run *stuff* on a bunch of repo)
would be cleaner using tekton (failure granularity, in particular). I'm still
gonna try to take a look, we'll see.

VannTen:
yq one-line for replacing the image version, if we want:
```bash
yq -y '(.presubmits[].spec.containers[].image | select(contains("precommit"))) |= "new_image_ref"' < .prow.yaml | sponge .prow.yaml
```

VannTen:
> The prow CI is serving multiple org, and not all require the pre-commit checks for their PRs.

If I understand the syntax right, the presubmit job would only be
defined for thoth-station org. Since there is already reference to
specific orgs (and repos) in the rest of the config I thought this was
acceptable that was acceptable

```yaml
presubmits:
 thoth-station:
    # defs of presubmits
```

> So that is the reason, the in_repo_config is switched on, so user can pick their CI checks. 

That would still be used. In particular, I was trying to determine if
when the same presubmits is defined in central config (`config.yaml`) an
in repo (`.prow.yaml`) the in repo would "win". But I'm unable to find
and answer in the prow docs, and had no response yet in the issue I
opened on their github repo :<.
If anyone has answers for that please speak up !

The whole "defining a baseline pre-submits job" depends on that,
otherwise we won't have a choice anyway and will use the .prow.yaml
everywhere. 

> we weren't planning for this change, right?
> we were looking for ways to do: 
> 1. auto update the `pre-commit autoupdate` 
> 2. auto update the base image for the repo that has pre-commit image in .prow.yaml.

Regarding 2, we had considered just using the `latest` tag, which would
render unnecessary updating the image version in .prow.yaml itself
(well, just one time, but no need for a periodic job for that).

Defining the job in the prow config was part of the same idea of
reducing the maintenance at the repo level. Having a central baseline
facilitates some things (like adding the cache stuff for example,
instead of needing opt-in from all the repos).

Most of the repo have a nearly identical pre-commit job, I
think. Quick metric:

$ find thoth-station/.cache -name .prow.yaml -exec sh -c 'yq .presubmits[0] < $1 | md5sum' _ {} \; | cut -d ' ' -f 1 | sort | uniq -c
      1 0be42889ed5615d64062faf463e5d8ee
      3 13d3d0b95a79acc92d744cab11239f19
      3 287a36059a1d32f42bf7ca196c609af1
      1 2f9ad4f38c41a6112cf635e3548f2664
      1 3bc26a4cdaf6b30b978b11030aa08ae5
      2 54fae42a48626431b42102395f98a230
      1 5a7fe382a49e5f1a83be8f82d5d79b01
      1 5af43a65775593758add8aeba2366f17
     10 660cc76415a85cbde34858331a8b54d2
      1 6d232559b3289fe75eae45891c1e5510
     22 78aa744ccc3415d1137ff347fa133521
      1 85d39dbdf0244e974d42da70818af3f6
      2 8a7165cc1b08a68038dbec97661f2edf
      4 a3495d6101b61fa72417ac4752adb836
      5 a3d686aaa080b7224397b9f1a2eb58b2
      1 a4c8461200cf4a3677d1ea724fffc1fe
      9 a5b27936847d0ea718e5530018da7482
      1 aafdef93868876091ec487a5c7f5a1a8
      1 c747b9f88feeb5297db442d60884bf82
      1 ca810815db713a42658afe57a9b13699
      1 d7fd96e649d65c9727d4dbbd9be689d3
      1 de4ff2292b869cc37f7396aba824879a
      2 f0170efa7ef99b42da5725aa85cb09d2
      5 f4b0e27443a5492dc75128e18b149e0f
      1 fa463c7dc990df8c82df5ed765b3a290


> ack, we wouldn't need this as we were not planning on execute pre-commit checks through this way.
> we can plan for this in a separate task, if this help in larger scale.

Ok, I'll do that in another PR.
I think it can be a big improvement and it's quite easy.

***@***.*** ~/redhat/thoth-station/metrics-exporter] (master =)$ time pre-commit run --all-files
[Hooks installing ...]
[Hooks running ...]

real	0m24.788s
user	0m17.345s
sys	0m1.890s

***@***.*** ~/redhat/thoth-station/metrics-exporter] (master =)$ time pre-commit run --all-files
[Hooks running ...]

real    0m1.161s
user    0m3.713s
sys     0m0.441s

> It was in reference to the autoudpate of the baste image for the precommit images.
> we dont need to look for the active repo anymore, that was only suggestion tobe done in case this feature was to be in aicoe-ci. 

That's great, that cranks the complexity down :)

VannTen:
/remove-lifecycle stale
/lifecycle frozen
Will wait until we have something for #2462

goern:
I'll close this, as it seems to have run dry. Feel free to re-open

/close

@VannTen @harshad16 @Gregory-Pereira Thanks for the energy! :1st_place_medal:

PR #2333, opened 2022-08-30
docs: Remove getting started page

### LINKED ISSUE #1766: Suggestion about Inexact information on the GitOps Docs Getting Started page
## Inexact information on the Getting Started page of the GitOps Documentation

In the section ["There are 3 parts to Operating First"](https://www.operate-first.cloud/apps/content/about/intro.html#there-are-3-parts-to-operating-first), the third point states that "The Mass Open Cloud (MOC) is a production cloud resource where projects are run.". While it remains somewhat true since smaug within MOC is the cluster for almost all user workloads, I think it seems outdated to just mention MOC and not the rest of the clusters (some of these also run experimental workloads) in this explanation/description of what Operate First is.

SamoKopecky:
Resolves #1766 

We decided to remove this page since it is outdated. I put a link to the websites getting started page until we recreate this page on our own.

/cc @HumairAK

### DIFF
--- a/docs/content/about/intro.md
+++ b/docs/content/about/intro.md
@@ -1,53 +1,5 @@
 # Getting started with Operate First
 
-If you are an open source developer interested in bringing your project to an Operate First environment, you have come to the right place.
-Below is important information and instructions on how to get started.
+Visit the [Getting Started][1] page on out website.
 
-## There are 3 parts to Operating First
-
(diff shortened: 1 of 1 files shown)


durandom:
@schwesig @quaid you might want to check if all the information we just removed here is available at https://www.operate-first.cloud/getting-started/ if it's still relevant. If it's not relevant anymore, great, but if so, maybe add it to the website

PR #2344, opened 2022-08-31
feat(catalog): Create CronJob to upload SA tokens to vault

SamoKopecky:
Part of https://github.com/operate-first/service-catalog/issues/59

## Changes
Add following resources to all managed clusters:
- `CronJob` that uploads token of `service-catalog-k8s-plugin` service account to vault
- `ConfigMap` `service-catalog-k8s-plugin` that tells the `CronJob` which env/cluster it is in
- `ServiceAccount` `service-catalog-k8s-plugin` that will be used for fetching pods, deployments etc. for the service-catalog k8s plugin
- `Secret` `service-catalog-k8s-plugin-token` for the `service-catalog-k8s-plugin` SA
- `Namespace` `service-catalog-k8s-plugin` where the `CronJob` will run

Add `secret-mgmt` base to `service-catalog-k8s-plugin` namespace in all clusters
The tokens are stored only in smaug vault endpoint.


cluster-scope/overlays/prod/emea/balrog/configmaps/service-catalog-k8s-plugin.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: service-catalog-k8s-plugin
+data:
+  env: emea
+  cluster: balrog
```
tumido, line 7: Maybe a nit, however can we capitalize these? `$ENV` and `$CLUSTER` is easier to recognize in the script above than `$env` and `$CLUSTER`. Since capitals are convention for constants and env variables, it's easier to pick up these values come from a mounted configmap...

cluster-scope/components/project-admin-rolebindings/service-catalog-k8s-plugin/rbac.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+    name: namespace-admin-service-catalog-k8s-plugin
+roleRef:
+    apiGroup: rbac.authorization.k8s.io
+    kind: ClusterRole
+    name: admin
+subjects:
+    - apiGroup: rbac.authorization.k8s.io
+      kind: Group
+      name: operate-first
```
tumido, line 12: I do not understand this cluster role, how is it different from https://github.com/operate-first/apps/blob/master/cluster-scope/components/project-admin-rolebindings/operate-first/rbac.yaml ?

You are not creating new user group here. The `service-catalog-k8s-plugin` namespace is owned by `operate-first` group, hence you should reuse the `project-admin-rolebindings/operate-first` component.

cluster-scope/base/batch/cronjobs/service-catalog-k8s-plugin/cronjob.yaml
```diff
@@ -0,0 +1,35 @@
...
+metadata:
+  name: upload-token-to-vault
+spec:
+  schedule: "0 * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: upload-token-to-vault
```
tumido, line 12: Can you please specify some `.resources.requests` and `resources.limits` here?

cluster-scope/base/core/namespaces/service-catalog-k8s-plugin/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: service-catalog-k8s-plugin
+resources:
+    - namespace.yaml
+components:
+    - ../../../../components/project-admin-rolebindings/service-catalog-k8s-plugin
+    - ../../../../components/limitranges/default
+    - ../../../../components/resourcequotas/small
```
tumido, line 9: can we go `x-small` here?

cluster-scope/base/batch/cronjobs/service-catalog-k8s-plugin/cronjob.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: upload-token-to-vault
+spec:
+  schedule: "0 * * * *"
```
tumido, line 6: I think you should also set:
- https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/#concurrency-policy
- https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/#jobs-history-limits
- https://kubernetes.io/docs/concepts/workloads/controllers/job/#clean-up-finished-jobs-automatically

tumido (changes requested):
I have few nits, otherwise this looks great to me!

cluster-scope/base/batch/cronjobs/service-catalog-k8s-plugin/cronjob.yaml
SamoKopecky: I've set it so that can view the last 3 failed Jobs and if the job is successful it gets deleted immediately. Let me know if thats fine.

tumido (changes requested):
Job history and other changes looks good, I think we still need to change the concurrency policy. However overall this looks great! :+1:

cluster-scope/base/batch/cronjobs/service-catalog-k8s-plugin/cronjob.yaml
```diff
@@ -0,0 +1,44 @@
...
+            - name: VAULT_ADDR
+              value: https://vault-ui-vault.apps.smaug.na.operate-first.cloud
+            envFrom:
+            - secretRef:
+                name: service-catalog-k8s-plugin-token
+            - configMapRef:
+                name: service-catalog-k8s-plugin
+            resources:
+              limits:
+                cpu: 250m
```
tumido, line 40: Missing memory request/limit here.

tumido (changes requested):

SamoKopecky:
Also noticed that namespace and labels were missing in the cronjob, so I added those too.

tumido:
Btw, I think you can lower the memory and cpu request, I don't expect this to need more than 50 milicores and 128Mi RAM...

tumido (approved):
/lgtm
/approve

PR #2361, opened 2022-09-01
feat: Add missing Resource entities, add metadata.title to match ADR

tumido:
Adapt changes to existing systems in service catalog so it matches [ADR 0022](https://github.com/operate-first/blueprint/pull/109), namely:

- Each provided service has a system, which is featured on the homepage (if user facing)
- Each system has gateway/access links for users
- Operators are components providing other components or resources direcly
- Namespaced operators are duplicit for each deployment, they are part of their respective systems now
- Resources map both clusters, externally provided resources and internally provided resources like databases etc.


service-catalog/systems/grafana.yaml
```diff
@@ -2,7 +2,15 @@ apiVersion: backstage.io/v1alpha1
...
 metadata:
   name: grafana
-  description: Grafana system
+  title: Grafana (private)
+  description: Grafana is a multi-platform open source analytics and interactive visualization web application. This instance is available to community members only.
+  annotations:
+    operate-first.cloud/featured: "true"
+    operate-first.cloud/logo-url: https://raw.githubusercontent.com/grafana/grafana/main/public/img/grafana_icon.svg
+  links:
+    - url: https://grafana.operate-first.cloud/
```
SamoKopecky, line 11: Maybe we should remove this link since its for the public instance?

SamoKopecky (commented):
Like it a lot! Learned a lot about how things are connected just from reading these files :smile: 

Just have one small suggestion bellow.

service-catalog/systems/grafana.yaml
tumido: No, this is link to the private instance. Public instance has https://grafana-public.operate-first.cloud/ url

SamoKopecky: Sorry about that, I should check more carefully next time!

tumido:
rebased please review again

/cc @durandom @SamoKopecky @mimotej @Gregory-Pereira @schwesig

SamoKopecky (approved):
/lgtm

service-catalog/resources/service-catalog-bucket-cla-b58cf604-019a-467c-8cae-50e3a47c12b1.yaml
```diff
@@ -1,15 +1,16 @@
 apiVersion: backstage.io/v1alpha1
 kind: Resource
 metadata:
-  name: s3-bucket
-  description: AWS S3 bucket deployed using OpenShift
+  name: service-catalog-bucket-cla-b58cf604-019a-467c-8cae-50e3a47c12b1
```
Gregory-Pereira, line 4: Does it make sense to keep the bucketname as the resource / filename? I dont see any links to the resource itself (like the s3 endpoint for instance) so shouldnt we just use a shortend name for it and put the full bucket name in the description if we really need it?

service-catalog/components/service-catalog-postgres-cluster.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: service-catalog-pg-cluster
```
Gregory-Pereira, line 4: I think this name and the file name should be the same, it makes it easier when reviewing and keeping track of what resources are referenced where

service-catalog/resources/service-catalog-postgres-cluster.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: backstage.io/v1alpha1
+kind: Resource
+metadata:
+  name: service-catalog-pg-cluster
```
Gregory-Pereira, line 4: same thing as above, consider renaming resource or file

Gregory-Pereira (commented):
/lgtm 
I just had a few comments about resource / filenames which are all stylistic. As such I will approve the PR but hold until you see my comments and can decide what to do with it.
/hold
/approve

service-catalog/components/service-catalog-postgres-cluster.yaml
tumido: good catch, will fix that :slightly_smiling_face:

durandom (commented):
I'm not really sold on the grafana system/service/operator relationships.

Wouldnt this suffice:

System: 
 * grafana
 
Components: 
 * grafana-operator (clusterwide service)
 * grafana-public (website)
 * grafana (website)

API: 
 * grafana
 * grafana-public

service-catalog/resources/service-catalog-bucket-cla-b58cf604-019a-467c-8cae-50e3a47c12b1.yaml
tumido: Not sure what would you like to see here. Bucket name is the only unique identifier we have for this particular bucket. Keep in mind that this entity describes a specific bucket for Service Catalog. It maps a resource, not a resource type (s3 buckets) in general. 

Yes, there are no links, however there's a relation listed below - this bucket is provided by `odf-external-cluster-smaug` Ceph cluster resource. This cluster is not intended to be accessed from outside of the cluster, so linking `s3.openshift-storage.svc` host doesn't make any sense.

service-catalog/components/opf-grafana-public-grafana-operator.yaml
```diff
@@ -1,14 +1,15 @@
 apiVersion: backstage.io/v1alpha1
 kind: Component
 metadata:
-  name: grafana-public
-  description: The public instance of grafana in operate first
+  name: opf-grafana-public-grafana-operator
```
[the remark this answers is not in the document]
durandom: ok. 🤦 wasnt aware that the grafana operator is namespaced....

service-catalog/components/opf-monitoring-grafana-operator.yaml
```diff
@@ -0,0 +1,33 @@
...
+  description: A Kubernetes Operator based on the Operator SDK for creating and managing Grafana instances
+  annotations:
+    operate-first.cloud/logo-url: https://raw.githubusercontent.com/grafana/grafana/main/public/img/grafana_icon.svg
+    argocd/app-name: grafana-smaug
+    github.com/project-slug: operate-first/apps
+    backstage.io/kubernetes-label-selector: "control-plane=controller-manager"
+    backstage.io/kubernetes-namespace: opf-monitoring
+    backstage.io/techdocs-ref: url:https://github.com/operate-first/apps/tree/master/docs/content/grafana
+  links:
+    - url: https://github.com/operate-first/apps/issues/new
```
durandom, line 15: to avoid duplication and maybe links going stale, should we add most links at the `systems` level?
probably just a detail and nitpick

durandom:
/lgtm
/approve

@tumido just unhold if you want to get it in

PR #2365, opened 2022-09-02
enable debug for bq plugin, per request from trino devs

erikerlandson:
Trying to debug why the riskthinking catalog is failing on reading data from tables with "table not found" errors


Gregory-Pereira (approved):
/lgtm

rynofinn (approved):

PR #2367, opened 2022-09-02
add grafana-public dashboard

schwesig:
solves https://github.com/open-services-group/scrum/issues/40

### DIFF
--- a/grafana/overlays/moc/smaug-public/dashboards/operate-first/opf-projects-overview.yaml
+++ b/grafana/overlays/moc/smaug-public/dashboards/operate-first/opf-projects-overview.yaml
@@ -1,0 +1,874 @@
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata:
+  labels:
+    app: grafana
+    public: "true"
+  name: opf-projects-overview
+spec:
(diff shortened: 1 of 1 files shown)


schwesig:
dashboard
snapshot
[grafana.operate-first.cloud/dashboard/snapshot/2I4k71DpSz3wvIWvvnT2eSi08XyVH71N](https://grafana.operate-first.cloud/dashboard/snapshot/2I4k71DpSz3wvIWvvnT2eSi08XyVH71N)

Gregory-Pereira:
I haven't used Grafana snapshots before but Im pretty sure it implies that successfully created this resource in a Grafana instance which is pretty much all the validation we need. Just in case I am going to create it as a dry run and then delete it so argocd can apply the changes.

Gregory-Pereira:
@schwesig, when loading it in, it says "not valid json" are we sure that this is constructed properly? I think you may be missing a closing parenthasy or something.

Gregory-Pereira:
Never mind, it looks good.
/lgtm
/approve

Gregory-Pereira:
This will fail being synced and argocd as smaug grafana and public smaug grafana are competing over who owns the admin credentials resource. I will introduce a pr to fix this.

PR #2378, opened 2022-09-05
Added roastdeer to open-aiops

roastdeer:
Added myself as per the steps described in a reply to my issue https://github.com/operate-first/support/issues/629


durandom:
@roastdeer can we get a +1 from your team @davidohana @ronensc @drbwa or @eranra ?

durandom:
I think this group is not the right place, it was a different project. Let's continue the discussion in https://github.com/operate-first/support/issues/629 to find a right place

PR #2380, opened 2022-09-05
[WIP] POC create namespace resource with templated patch file

durandom:
based on https://docs.google.com/document/d/1K7-lBRvYCrEbjKG0-G74vbE9xVlV1-v1HJncAotHMD4/edit#

cc @tumido


durandom:
https://github.com/operate-first/robozome/pull/10

scripts/patches/group.diff.mustache
```diff
@@ -0,0 +1,13 @@
+diff --git a/cluster-scope/base/user.openshift.io/groups/{{name}}/group.yaml b/cluster-scope/base/user.openshift.io/groups/{{name}}/group.yaml
+new file mode 100644
+--- /dev/null
++++ b/cluster-scope/base/user.openshift.io/groups/{{name}}/group.yaml
+@@ -0,0 +1,{{users.length}}@@
```
durandom, line 5: @tumido this is where it breaks already, at least with the golang implementation of mustache. There's no `.length` implemented.
Removing the `@@` line results in an empty applied patch, so the number has to be accurate.

I'm investigating some of the jsonnet/ksonnet tooling mentioned at https://jsonnet.org/articles/kubernetes.html

durandom:
@HumairAK @mimotej PTAL at https://github.com/operate-first/apps/pull/2491
That PR was created by running the commands in scripts/README_patches.md

1. please review in #2491 if I'm missing something
2. I think the jinja method is the most readable and straightforward way of doing things.

scripts/jinja/group.env
```diff
@@ -0,0 +1,2 @@
+GROUP=gingersnap
+USERS="['ryanemerson', 'pminz', 'tristantarrant', 'karesti', 'wburns', 'fax4ever', 'rigazilla', 'jabolina', 'pruivo', 'dpanshug']"
```
durandom, line 2: @mimotej @HumairAK this is where https://github.com/mattrobenolt/jinja2-cli falls apart when parsing from ENV variables. I think yaml configs would work. 
Did you start looking into passing parameters from issues to the bot? Is there an example yaml config file?

PR #2382, opened 2022-09-06
updated readme

durandom:
* vault login
* cluster envs
* argocd cli


Gregory-Pereira (commented):
One thought, it strikes me that with [Tom's Pr to change the pre-submit triggers](https://github.com/operate-first/apps/pull/2321) we only run pre-commit on changes to the docs, but we should also have a pre-submit to ensure the docs themselves build properly. Thoughts?

/lgtm
/approve

durandom:
> One thought, it strikes me that with [Tom's Pr to change the pre-submit triggers](https://github.com/operate-first/apps/pull/2321) we only run pre-commit on changes to the docs, but we should also have a pre-submit to ensure the docs themselves build properly. Thoughts?

I ran the pre-commit locally. IIUC it was disabled because it breaks the prow pipeline. I wouldn't spend too much time on it unless the issue is solved already

PR #2385, opened 2022-09-06
feat: Enable ODF console plugin on Infra

tumido:
OCP 4.9 introduced dynamic console plugins, this is a first attempt to track which ones are enabled on Infra cluster (right now ODF one is disabled, this PR enables it)


Gregory-Pereira:
/retest kubeval-validation
This should be fixed with storage being moved

PR #2388, opened 2022-09-06
feat(catalog): Allow CronJob to write secrets to vault

SamoKopecky:
Changes to `CronJob`:
* Save the final vault token to an env variable instead of a file by default
* Change role to one that allows writing of secrets
* Switch from `jq` to `yq` which is installed in toolbox image
* Switch from running as `vault-secret-fetcher` to `vault-secret-writer`

Changes to manifests:
* Add `service-catalog-k8s-plugin` config map for moc/smaug to kustomization file
* Remove the addition of `vault-secret-fetcher` SA from clusters other then smaug
* Create `vault-secret-writer` SA for all clusters

### DIFF
--- a/cluster-scope/base/batch/cronjobs/service-catalog-k8s-plugin/cronjob.yaml
+++ b/cluster-scope/base/batch/cronjobs/service-catalog-k8s-plugin/cronjob.yaml
@@ -45,5 +45,5 @@ spec:
               requests:
                 cpu: 50m
                 memory: 128Mi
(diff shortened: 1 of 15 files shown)


SamoKopecky:
/hold 
Need to add `$ENV-ops-rw` roles to vault.

SamoKopecky:
/unhold

`$ENV-ops-rw` roles were added to vault.

tumido (approved):
/lgtm
/approve

PR #2389, opened 2022-09-06
Deploy stackrox via a helm chart using argocd

traytorous:
This is an alternative way to deploy Stackrox onto the smaug cluster. This is because using the other method (getting the templates from helm directly and managing it) could lead to some problems down the road from an upgrading standpoint. This also seems to remove the need for external secrets because the chart itself generates the secrets and place the secrets on the open-shift cluster.

### DIFF
--- a/stackrox/base/deployments/stackrox-central-deployment.yaml
+++ b/stackrox/base/deployments/stackrox-central-deployment.yaml
@@ -1,0 +1,16 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: stackrox-central-deployment
+  namespace: stackrox
+spec:
+  project: stackrox
+  source:
+    chart: stackrox/stackrox-central-services  
+    repoURL: https://raw.githubusercontent.com/stackrox/helm-charts/main/opensource/
+    targetRevision: 71.0.0
+    helm:
+      releaseName: stackrox-central-services
+  destination:
+    server: " "
+    namespace: stackrox
--- a/stackrox/base/deployments/stackrox-secured-cluster-services-deployment.yaml
+++ b/stackrox/base/deployments/stackrox-secured-cluster-services-deployment.yaml
@@ -1,0 +1,16 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: stackrox-secured-cluster-services-deployment
+  namespace: stackrox
+spec:
+  project: stackrox
+  source:
+    chart: stackrox/stackrox-secured-cluster-services
+    repoURL: https://raw.githubusercontent.com/stackrox/helm-charts/main/opensource/
+    targetRevision: 71.0.0
+    helm:
+      releaseName: stackrox-secured-cluster-services
+  destination:
+    server: " "
+    namespace: stackrox
--- a/stackrox/overlays/moc/smaug/kustomization.yaml
+++ b/stackrox/overlays/moc/smaug/kustomization.yaml
@@ -1,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../../../base
+namespace: stackrox
+patchesStrategicMerge:
+- set-server.yaml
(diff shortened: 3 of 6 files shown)


stackrox/base/deployments/stackrox-central-deployment.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: stackrox-central-deployment
+  namespace: stackrox
+spec:
+  project: stackrox
+  source:
+    chart: stackrox/stackrox-central-services  
+    repoURL: https://raw.githubusercontent.com/stackrox/helm-charts/main/opensource/
```
durandom, line 10: This gives a `400`. Is there a way to view the chart?

stackrox/overlays/moc/smaug/set-server.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: stackrox-central-deployment
+  namespace: stackrox
+spec:
+  destination:
+    server: "smaug"
```
durandom, line 8: can we try `jerry` first?

durandom (commented):
@traytorous dont you need to supply any values for the chart? like https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#values-files

also let's wait for @HumairAK to review.

stackrox/base/deployments/stackrox-central-deployment.yaml
traytorous: Hi @durandom I had the same question as well. However for some reason it does work when I use helm pull. I don't know what type of magic they are using but you can only get the chart  using helm once you add the repo to your local helm list. Instructions on how to do it is below.
- https://github.com/stackrox/helm-charts

stackrox/overlays/moc/smaug/set-server.yaml
traytorous: I believe we can? I have never heard of the Jerry cluster before. Also I don't know if the server argument that I put down (smaug) is correct. I was going to wait for @HumairAK to look at it and review it with me.

stackrox/base/deployments/stackrox-secured-cluster-services-deployment.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
```
durandom, line 2: All argocd applications go into https://github.com/operate-first/apps/tree/master/argocd/overlays/moc-infra/applications/envs/emea/jerry

stackrox/base/deployments/stackrox-secured-cluster-services-deployment.yaml
```diff
@@ -0,0 +1,16 @@
...
+  namespace: stackrox
+spec:
+  project: stackrox
+  source:
+    chart: stackrox/stackrox-secured-cluster-services
+    repoURL: https://raw.githubusercontent.com/stackrox/helm-charts/main/opensource/
+    targetRevision: 71.0.0
+    helm:
+      releaseName: stackrox-secured-cluster-services
+  destination:
```
durandom, line 14: destination would look like here: https://github.com/operate-first/apps/blob/master/argocd/overlays/moc-infra/applications/envs/emea/jerry/cluster-management/cluster-resources.yaml#L6-L7

stackrox/overlays/moc/smaug/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
durandom, line 1: I dont think we need all this overlay stuff here. Should be enough to have the `Kind: Application` to let argocd pick it up

durandom (changes requested):
Looking at https://github.com/stackrox/stackrox#quick-installation-using-helm and inspecting https://raw.githubusercontent.com/stackrox/stackrox/master/scripts/quick-helm-install.sh it looks like you need to install the central services first

```
helm install -n stackrox --create-namespace stackrox-central-services stackrox/stackrox-central-services \
 --set central.adminPassword.value="${STACKROX_ADMIN_PASSWORD}" \
 "${installflags[@]+"${installflags[@]}"}"
```

You'll also want to provide a values files as per https://argo-cd.readthedocs.io/en/stable/user-guide/helm/ to set e.g. the adminPassword

PR #2391, opened 2022-09-07
Fake metrics endpoint to make ODF happy

tumido:
Related to: https://github.com/operate-first/support/issues/652

Cherrypicking from @larsks https://github.com/OCP-on-NERC/nerc-ocp-config/pull/18

> While the "monitoring-endpoint" configuration in OCS 4.8 was optional,
it appears to be required in 4.10, and ODF refuses to deploy if it
can't reach the endpoint.
> 
> This commit provides a fake metrics endpoint to make the ODF operator
happy.


yehudasa:
@oritwas @jdurgin maybe there should be a better way to override ODF?

yehudasa:
@durandom @larsk @tumido
According to @oritwas there are a couple of issues here. First, you're supposed to read the ceph-mgr export endpoint and use that one instead of the fake one that you provided.
The other issue, is that this should actually be optional according to the external mode docs. So will either need to fix the docs or fix the behavior. Can you open a jira ticket re this?

tumido:
@yehudasa the problem here is that the external Ceph solution we use doesn't provide a metrics endpoint to users so we don't have access to it.

Is there a settings for `StorageCluster` where we can disable ODF contacting the metrics endpoint?

https://github.com/operate-first/apps/blob/e9bf61f7debcd4478e963a7eb13b2b87fe1cf6fd/cluster-scope/base/ocs.openshift.io/storageclusters/ocs-external-storagecluster/storagecluster.yaml#L9-L11

ODF refuses to create the storage cluster unless it can contact metrics endpoint in the external mode.

tumido:
@yehudasa Which project in Jira should I use for the ticket?

durandom:
pinged @yehudasa via chat

tumido:
@yehudasa I think it's not optional, however the docs are a bit confusing about this:

image


Anyways, opened a BZ here: https://bugzilla.redhat.com/show_bug.cgi?id=2129117

PR #2392, opened 2022-09-07
feat(catalog): Move manifests to a bundle, fix cronjob

SamoKopecky:
Changes to cronjob:
* export `VAULT_TOKEN` variable so that vault can see it
* fix secret creation

Changes to manifests:
* Create a bundle from manifests so that it is grouped in one logical part


tumido (approved):
/lgtm
/approve

PR #2393, opened 2022-09-08
prow metrics

durandom:
- fix prow servicemonitor, remove public metrics routes

removed the public routes and changed the scheme to http

https://docs.openshift.com/container-platform/4.9/monitoring/troubleshooting-monitoring-issues.html helped troubleshooting

`oc port-forward -n openshift-user-workload-monitoring pod/prometheus-user-workload-0 9090`
then go to Targets and see if the prow targets are up


durandom:
@harshad16 are those public routes needed? and what about https://github.com/operate-first/apps/blob/master/prow/overlays/smaug/route-ghproxy-metrics.yaml ?

The idea would be to collect all metrics internally and make them avail in the central prom/grafana

harshad16:
> are those public routes needed? 

It was public just for debugging purposes.
The metrics routes don't need to be public.

> and what about https://github.com/operate-first/apps/blob/master/prow/overlays/smaug/route-ghproxy-metrics.yaml?

same for ghproxy metrics, this route is just for exposing metrics.

durandom:
@Gregory-Pereira please review

prow/overlays/smaug/service-monitor.yaml
```diff
@@ -6,10 +6,12 @@ metadata:
     monitor-component: operate-first
 spec:
   endpoints:
-    - interval: 30s
+    - path: /metrics
       port: metrics
-      scheme: https
+      scheme: http
```
Gregory-Pereira, line 11: Should scheme be included here, and should it be `http`? Our other service monitors [see search for examples](https://github.com/operate-first/apps/search?q=serviceMonitor) don't specify a scheme at all.
```suggestion
```

prow/overlays/smaug/service-monitor.yaml
```diff
@@ -6,10 +6,12 @@ metadata:
...
   namespaceSelector:
     matchNames:
       - opf-ci-prow
-  selector: {}
+  selector:
+    matchLabels:
+      app.kubernetes.io/component: prow
```
Gregory-Pereira, line 17: Do we need this selector? We already have specified that we want to monitor things in the `opf-ci-prow` namespace, what are we hoping to filter out by using this additional label selector?

Gregory-Pereira (commented):
/lgtm, just two questions

prow/overlays/smaug/service-monitor.yaml
durandom: Maybe it defaults to `http`, but before it was `https` and would not work with the exposed services. So I would rather keep it explicit

prow/overlays/smaug/service-monitor.yaml
durandom: yes, you're right. removing the selector works

Gregory-Pereira (approved):
/lgtm

PR #2395, opened 2022-09-08
whitelist grafana resources for b4mad

durandom:
otherwise, I get 
`Resource integreatly.org:Grafana is not permitted in project b4mad.`

image


Gregory-Pereira (approved):
/lgtm

PR #2396, opened 2022-09-08
moving prow storage to ionos s3 bucket

Gregory-Pereira:

Since the current prow configuration seems to work to some extent, I have commented out the OBC manifest in the `kustomization.yaml` and other references to it, so we may use it again whenever we wish.

/cc @harshad16


harshad16 (commented):
/lgtm
:+1:

PR #2403, opened 2022-09-13
Remove old smaug trino catalogs.

HumairAK:
Primarily the ones backed by s3 object buckets that no longer exist (from clusters rick and morty). Also convert secrets to leverage vault.

Furthers: https://github.com/operate-first/apps/issues/1996


Gregory-Pereira (commented):
/lgtm
/approve

PR #2221, opened 2022-08-03
feat(catalog): Redefine grafana component following a new ADR

SamoKopecky:
Part of [operate-first/service-catalog#55](https://github.com/operate-first/service-catalog/issues/55)

~~Blocked by: [operate-first/service-catalog#78](https://github.com/operate-first/service-catalog/pull/78)~~

## Changed
Redefine grafana as a fully defined entity following the new [ADR 0022](https://github.com/operate-first/blueprint/pull/109), which includes adding:
- `grafana`, `dex` APIs
- `smaug`, `sqlite` resources
- `grafana-public`, `grafana-oauth` components
- `dex`, `grafana` systems

Add labels to grafana operator deployments since they are needed for the k8s plugin.

Move grafana documentation to `docs` folder because `MkDocs` doesn't like when the source `.md` files are in the same directory as the `mkdocs.yaml` file. See the error here:

```
ERROR    -  Config value: 'docs_dir'. Error: The 'docs_dir' should not be the parent directory of the config file. Use a child directory instead so that the 'docs_dir' is a sibling of the config file.
ERROR    -  Config value: 'site_dir'. Error: The 'site_dir' should not be within the 'docs_dir' as this leads to the build directory being copied into itself and duplicate nested files in the 'site_dir'.(site_dir:
            '/home/skopecky/git/apps/docs/content/grafana/site', docs_dir: '/home/skopecky/git/apps/docs/content/grafana')
Aborted with 2 Configuration Errors!

```

PR #2405, opened 2022-09-13
Correct grafana label.

HumairAK:
SSIA

### DIFF
--- a/grafana/overlays/moc/smaug-public/grafana-public.yaml
+++ b/grafana/overlays/moc/smaug-public/grafana-public.yaml
@@ -6,7 +6,7 @@ metadata:
 spec:
   deployment:
     labels:
-      app.kubernetes.io/name=grafana-public
+      app.kubernetes.io/name: grafana-public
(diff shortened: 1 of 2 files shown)


HumairAK:
Curious why this wasn't caught by kubeval here https://github.com/operate-first/apps/pull/2221 

@mimotej any clue?

mimotej:
> Curious why this wasn't caught by kubeval here #2221
> 
> @mimotej any clue?

@HumairAK Grafana resource seems to be listed in [skip-kinds file](https://raw.githubusercontent.com/operate-first/schema-store/main/skip_kinds.txt). That means it was ignored. Interesting thing tho is that when I ran that manually without using skip_kinds file everything was ok. I am considering whether or not is `skip_kinds` file necessary since we are using `--ignore-missing-schemas` anyway. Maybe we should remove this flag and not use it in current setup (main reason behind using it was that there was an issue that some manifests couldn't be generated from openapi spec and this allowed us to ignore these schemas without using `--ignore-missing-schemas`, but since we are using it anyway this is just another link in the chain which can cause false negatives...)

PR #2407, opened 2022-09-14
chore: Update group/user in service catalog

tumido:
Update Group and User with some profile details in service catalog

/cc @SamoKopecky


SamoKopecky (approved):
/lgtm

PR #2413, opened 2022-09-15
Drop prow autoclosing for thoth-station

VannTen:
Resolve https://github.com/thoth-station/core/issues/456

See [Thoth tech talk](https://docs.google.com/document/d/1jz4j07Q8gkv9H-23BgOp_4qN3uI29JsuP-_kTKEj-e8/edit#heading=h.5eiqu7cdu6o2) for context.

PR #2414, opened 2022-09-15
thoth-station: Close unactive issues needing reporter input

VannTen:
Possible follow-up to #2413

See https://github.com/thoth-station/core/issues/456#issuecomment-1247240211

(needs discussion)

@KPostOffice
@goern


KPostOffice:
My only question is what happens after the reporter makes a comment? Is there any way to automatically remove the triage labels so that it gets taken off this timer after a user input?

VannTen:
I see two ways to do it:

- remove the triage labels automatically when reporter responds, adding
  it back if necessary (information not sufficient, other needed)
- let the triager to the removing, not assuming the information added is
  sufficient until it is validated.

The former require more automation (detecting new comment from
particular poster).

I have a slight preference for the removing the triage label
automatically, I think it works better when we can't invest time on some
issues immediately.

KPostOffice:
> I see two ways to do it: - remove the triage labels automatically when reporter responds, adding it back if necessary (information not sufficient, other needed) - let the triager to the removing, not assuming the information added is sufficient until it is validated. The former require more automation (detecting new comment from particular poster). I have a slight preference for the removing the triage label automatically, I think it works better when we can't invest time on some issues immediately.

I do too. I think it better represents the work that needs to be done. I'm fine leaving it for a future issue for now though. WDYT?

VannTen:
> I do too. I think it better represents the work that needs to be done.
> I'm fine leaving it for a future issue for now though. WDYT?

Fine from me as well.

-

I thought of something else: sometimes we put `triage/needs-information`
because we ourselves need to do some research. Do we have something more
specific like `needs-reporter-feedback` or something like that ?

KPostOffice:
> I do too. I think it better represents the work that needs to be done. I'm fine leaving it for a future issue for now though. WDYT?
> Fine from me as well. - I thought of something else: sometimes we put `triage/needs-information` because we ourselves need to do some research. Do we have something more specific like `needs-reporter-feedback` or something like that ?

I don't think so. The closest thing we have is `priority/awaiting-more-evidence`

VannTen:
/remove-lifecycle stale

From the discussion, I think we're fine with the current state ?
We might want to add a more precise label or use `priority/awaiting-more-evidence`, but that can wait a follow-up.

wdyt ?
@goern @KPostOffice

PR #2415, opened 2022-09-16
Add docs on updating jwt for vault runbook.

HumairAK:
Documenting here as we encountered this earlier today.


tumido (approved):
/lgtm
/approve

PR #2419, opened 2022-09-19
fix(catalog): Add bundle and change configmap

SamoKopecky:
The `service-catalog-k8s-plugin` bundle is missing from the curator overlay since it doesn't import the `common` overlay.
ConfigMaps for OSC clusters contain a typo.

/cc @tumido


tumido (commented):
/lgtm
/approve

PR #2424, opened 2022-09-19
externalizing alertreceiver secrets and removing emea overlay

Gregory-Pereira:
Part one of externalizing all secrets in apps repo.

/cc @HumairAK 


alertreceiver/overlays/moc/smaug/kustomization.yaml
```diff
@@ -3,7 +3,8 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
-  - ../../common
+  - ../../base
```
HumairAK, line 6: build will fail, should be: `../../../base`

alertreceiver/overlays/moc/smaug/kustomization.yaml
```diff
@@ -3,7 +3,8 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
```
HumairAK, line 5: can you also add the namespace to the `kustomization.yaml` : `namespace: opf-alertreceiver`

PR #2425, opened 2022-09-19
externalizing triage-party secret

Gregory-Pereira:
Part of migrating away from sops in Operate-first/apps.

/cc @HumairAK 


triage-party/overlays/moc/smaug/kustomization.yaml
```diff
@@ -3,6 +3,7 @@ kind: Kustomization
 resources:
   - ../../../base
   - pcache.yaml
+  - secret.yaml
```
HumairAK, line 6: can you rename it to externalsecret.yaml?

PR #2429, opened 2022-09-20
bumping resource requests for prow jobs

durandom:
pre-commit is heavily parallelized
the other jobs exceed 1cpu a bit

I tested locally with `docker stats` and just looking at mem and CPU usage

cc @Gregory-Pereira

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -13,10 +13,10 @@ presubmits:
           resources:
             requests:
               memory: "256Mi"
-              cpu: "200m"
+              cpu: "1500m"
             limits:
               memory: "500Mi"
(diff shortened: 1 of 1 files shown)


.prow.yaml
```diff
@@ -32,10 +32,10 @@ presubmits:
           resources:
             requests:
               memory: "512Mi"
-              cpu: "200m"
+              cpu: "1.5"
```
HumairAK, line 35: "1.5" will get converted to "1500m" when deployed, this may result in argocd complaining (not sure), but let's just set it to "1500m" to be consistent with the api spec, same with the other decimal values.

PR #2430, opened 2022-09-20
feat: Add apicurio project to vault secret managament

mimotej:
Part of: https://github.com/operate-first/support/issues/620

### DIFF
--- a/cluster-scope/overlays/prod/moc/smaug/secret-mgmt/api-designer/kustomization.yaml
+++ b/cluster-scope/overlays/prod/moc/smaug/secret-mgmt/api-designer/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: api-designer
+resources:
+  - ../base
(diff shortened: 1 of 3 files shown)


tumido (approved):
/lgtm
/approve

HumairAK:
I'm not sure this is the way we want to do this. This secretstore uses the `moc-ops` role which would give this team the ability to create external secrets that can read from any namespace listed here: https://vault-ui-vault.apps.smaug.na.operate-first.cloud/ui/vault/access/smaug-k8s/item/role/show/moc-ops 

We probably want to create a separate role for each team and create a secret store for them that uses only this role that is isolated to the team's namespace.

riprasad:
@mimotej Thanks for onboading our projects on vault. I followed the documentataion [Add External Secrets to a namespace](https://www.operate-first.cloud/apps/content/vault_eso/add_external_secrets_to_ns.html) to proceed forward, but I see `Not authorized - permission denied instead of a populated window`.

I am not sure if I am following the right steps to integrate vault to fetch secrets from my namespaces, but if I am on the right path, I am anticipating membership needs to be configured for us here `apps/cluster-scope/overlays/prod/moc/smaug/groups/vault-* group(s).`

PR #2443, opened 2022-09-20
removing keycloack deployment

Gregory-Pereira:
/cc @HumairAK 
removing entire keycloak directory as we are no longer deploying it on any cluster.
removing keycloack argocd app.
removing rhsso operator and where it is included in the infra cluster.


cluster-scope/overlays/prod/moc/infra/kustomization.yaml
```diff
@@ -24,7 +24,6 @@ resources:
   - ../../../../base/operators.coreos.com/subscriptions/cert-manager
   - ../../../../base/operators.coreos.com/subscriptions/cluster-logging
   - ../../../../base/operators.coreos.com/subscriptions/kubernetes-nmstate-operator
-  - ../../../../base/operators.coreos.com/subscriptions/rhsso-operator
```
tumido, line 27: Please remove `operatorgroup/keycloak` and `namespaces/keycloak` as well. Those can be also removed from `cluster-scope/base`

tumido (changes requested):
Looks good! :+1: Some additional resources needs to be cleaned up...

PR #2449, opened 2022-09-21
externalizing acm secrets

Gregory-Pereira:
/cc @HumairAK


HumairAK:
the obvious issue with this is that we use ACM to "manage" smaug, and vault is deployed on smaug, so ACM will depend on a cluster that it manages. The circular dependency is not ideal, but I think we should go ahead with this for now as this is better than having these secrets in git. 

In the worst case if smaug + vault goes down, the secrets will continue to persist on the managed cluster (infra) anyway. So not a big deal.

PR #2478, opened 2022-09-22
Add ingress/api certs to ingress-controller & apiserver config.

HumairAK:
Should resolve the issue of certs that are added by cert-manager not being picked up by jerry.


durandom:
@HumairAK CI fails, because you forgot some kustomization.yaml or include the resources directly?!

PR #2491, opened 2022-09-28
onboard kaoto

durandom:
fixes https://github.com/operate-first/support/issues/969

### DIFF
--- a/cluster-scope/base/core/namespaces/kaoto/namespace.yaml
+++ b/cluster-scope/base/core/namespaces/kaoto/namespace.yaml
@@ -1,0 +1,10 @@
+apiVersion: user.openshift.io/v1
+kind: Namespace
+metadata:
+  annotations:
+    op1st/docs: https://kaoto.io/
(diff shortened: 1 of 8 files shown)


cluster-scope/base/core/namespaces/kaoto/namespace.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: user.openshift.io/v1
+kind: Namespace
+metadata:
+  annotations:
+    op1st/docs: https://kaoto.io/
+    op1st/onboarding-issue: https://github.com/operate-first/support/issues/969
+    op1st/project-owner: kaoto
+    openshift.io/display-name: Kaoto Integrations
+    openshift.io/requester: kaoto
```
durandom, line 9: @mmelko please check if you're happy with the links here

mmelko: I'd change openshift.io/display-name to `Kaoto` too. `Kaoto Integrations` would suggest that Kaoto is actually an integration framework which is not.

PR #2494, opened 2022-09-30
fix kaoto namespace

durandom:
wrong apiversion

/assign @mimotej 

@mimotej do you know why this hasn't been caught by the schema validator?

### DIFF
--- a/cluster-scope/base/core/namespaces/kaoto/kustomization.yaml
+++ b/cluster-scope/base/core/namespaces/kaoto/kustomization.yaml
@@ -6,3 +6,4 @@ resources:
 components:
 - ../../../../components/limitranges/default
 - ../../../../components/resourcequotas/small
+namespace: kaoto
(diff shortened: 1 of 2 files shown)


mimotej:
> wrong apiversion
> 
> /assign @mimotej
> 
> @mimotej do you know why this hasn't been caught by the schema validator?
@durandom This is expected behaviour. Actual line that show what happened is here in the #2491 
image
Names of the schemas are mostly created using `kind` and `apiVersion` as can be seen [here](https://github.com/operate-first/schema-store/blob/main/schemas/master-standalone-strict/namespace-v1.json). So actual file it tried to find was propably `namespace-user.openshift.io/v1` (or something similliar). Therefore both of these things are not validated in our pipeline because we are using `--ignore-missing-schemas` which when it is unable to find schema throws only warning not error. This comes from the limitation of `kubeval` where we have to either accept that it will be failing in situation when we are adding new CRDs or that it will not check for these kinds of errors.

PR #2502, opened 2022-09-30
Updated configuration stack to move it to infra nodes

redmikhail:
Moving instances of OCP monitoring stack to infra nodes and adding pvc to eliminate issues with local filesystems being overrun. @HumairAK can you please take a look?


redmikhail:
@HumairAK

PR #2504, opened 2022-10-03
feat(catalog): Add ConfigMap for jerry in service catalog

SamoKopecky:
Part of https://github.com/operate-first/service-catalog/issues/113

Add a ConfigMap that is needed for the cronjob on jerry cluster for uploading secrets to vault so that the service catalog can query this cluster.


tumido (approved):
/lgtm
/approve

PR #2515, opened 2022-10-04
feat: Add Prow to service catalog

tumido:
SSIA

/cc @SamoKopecky

### DIFF
--- a/service-catalog/components/prow-hook.yaml
+++ b/service-catalog/components/prow-hook.yaml
@@ -1,0 +1,24 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: prow-hook
+  title: Hook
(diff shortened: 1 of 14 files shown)


service-catalog/components/prow-hook.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: prow-hook
+  title: Hook
+  description: Receives all the web hooks from GitHub. It is a stateless server that listens for GitHub webhooks and dispatches them to the appropriate plugins.
```
HumairAK, line 6:
```suggestion
  description: Receives all the webhooks from GitHub. It is a stateless server that listens for GitHub webhooks and dispatches them to the appropriate plugins.
```

tumido: this is all taken from upstream.. but :shrug: why not...

PR #2525, opened 2022-10-05
fix(catalog): Build grafana docs for the right components

SamoKopecky:
- Reference the correct grafana components 
- Remove documentation link from public grafana components

/cc @tumido


tumido (approved):
/lgtm
/approve

PR #2528, opened 2022-10-05
add image patch for airflow that uses trino-0.317.0 client

erikerlandson:
an attempt at upgrading the trino python client in the image used by airflow (via open-metadata).


erikerlandson:
Unsure how best to test this.

HumairAK:
let's go with this instead: https://github.com/operate-first/apps/pull/2534 
as we're coupling this with OM upgrade

PR #2545, opened 2022-10-10
feat(catalog): Add new resource permissions for k8s plugin

SamoKopecky:
Part of https://github.com/operate-first/service-catalog/issues/86

New resource permissions are required for an updated k8s plugin in the service catalog.

/cc @tumido


tumido (approved):
/lgtm
/approve

PR #2551, opened 2022-10-12
pv creation test

sostrades-radouanbouadel:


HumairAK:
/hold 

@sostrades-radouanbouadel -- we do not create PV's manually, they are dynamically created by creating `PersistentVolumeClaims`, on the cluster. Users of the cluster do not need to store these PVCs in the `apps` repo, but are welcome to store them in their own respective repositories.

PR #2552, opened 2022-10-13
Adding myself for odh-admin (osc-cl1)

VannTen:
@harshad16: that's the correct file right ? (for access working on CNBi)


harshad16 (approved):
yes this should be enough.

/lgtm
/approve

PR #2583, opened 2022-10-26
fix ray-bu namespace rolebindings

durandom:


Gregory-Pereira:
thanks for catching this.
/lgtm
/approve

PR #2587, opened 2022-10-30
fix(catalog): RHACM plugin API endpoint

SamoKopecky:
We are renaming `cluster-status` plugin to `rhacm` so the API endpoints need to updated.

/cc @tumido


tumido (approved):
/lgtm
/approve

PR #2589, opened 2022-10-31
Remoing all old usage of SOPs

Gregory-Pereira:

/cc @HumairAK 

Addresses: https://github.com/operate-first/apps/issues/1996#issuecomment-1290671136


cluster-logging/overlays/moc/infra/externalsecrets/clo-kafka.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: clo-kafka
+spec:
+  secretStoreRef:
+    name: opf-vault-store
+    kind: SecretStore
+  target:
+    name: clo-kafka
+  dataFrom:
+    - extract:
+        key: moc/infra/openshift-logging/clo-kafka
```
HumairAK, line 13: This won't work, since the crt content has been stored as binaries, you need to do something like this: 

https://github.com/operate-first/apps/blob/master/observatorium/overlays/moc/smaug/vector/clo-kafka.yaml#L14-L18
which gets data from: https://vault-ui-vault.apps.smaug.na.operate-first.cloud/ui/vault/secrets/k8s_secrets/show/moc/smaug/opf-observatorium/clo-kafka

HumairAK: same goes for smaug/externalsecrets

PR #2594, opened 2022-11-06
Add robozome ns

HumairAK:
This PR adds robozome namespace, and some starter scripts for onboarding/offboarding projects. The scripts are an alternative implementation to: https://github.com/operate-first/apps/pull/2380 
But still use jsonnet for templating. 

The scripts here are simplified to one script per task (onboarding & offboarding), with documents included. I've removed the dependency on jinja since it's not needed, and the need to create temp folders/files (also not needed). 

I've also added detailed comments in the scripts so it will make it easier to convert them to docs later. 

Scripts consume from a payload file that will be created by the robozome app, when running manually you can create your own payload file by basing it off the examples provided in this PR.

### DIFF
--- a/scripts/onboarding.sh
+++ b/scripts/onboarding.sh
@@ -1,0 +1,90 @@
+set -o allexport -o pipefail -ex
+
+# Pre-condition:
+# - Create the config file at location ${PAYLOAD_PATH}
+# - Clone the github.com/operate-first/apps repo at location ${WORKING_DIR}
+# - Working branch should be clean, checked out of upstream default branch.
+# - Set Environment variables: ORG_NAME (=operate-first), SOURCE_REPO (=support), ISSUE_NUMBER
+
+CONFIG=${PAYLOAD_PATH}
+REPO=${WORKING_DIR}/apps
+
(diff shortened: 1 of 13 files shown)


scripts/onboarding.sh
```diff
@@ -0,0 +1,90 @@
...
+kustomize edit set namespace ${NAMESPACE}
+kustomize edit add component ../../../../components/limitranges/default
+kustomize edit add component ../../../../components/resourcequotas/${QUOTA}
+
+# "Give the team being onboarded access to the Namespace via OCP rbac"
+
+cd ${REPO}
+RBAC_PATH="${REPO}/cluster-scope/components/project-admin-rolebindings/${GROUP}"
+mkdir ${RBAC_PATH}
+jsonnet --ext-str GROUP --ext-code USERS scripts/jsonnet/rbac.jsonnet | yq -P > ${RBAC_PATH}/rbac.yaml
```
durandom, line 59: ❤️  this. Way cleaner than creating json files and converting the files via yet another tool. Cool that this works with `yq`

PR #2595, opened 2022-11-07
feat(catalog): Create exeternal secret for service-catalog

SamoKopecky:
Part of https://github.com/operate-first/service-catalog/issues/128

Create an external secret for a secret in vault. The SA `vault-secret-fetcher` has already been created. Namespace has been added to the `smaug-k8s/moc-ops` role.

/cc @tumido


tumido (approved):
/lgtm
/approve

PR #2597, opened 2022-11-09
[prow] Remove legacy presubmit check for from tide config

codificat:
`prow` (`tide`) is not merging pull requests for the `sefkhet-abwy` repository. See for example https://github.com/AICoE/sefkhet-abwy/pull/203

The pre-commit check driven by the aicoe-ci pipelines was removed a while ago, and this is now covered by a prow pre-submit job, so this PR removes that check from the tide query list for that repo.


harshad16 (approved):
/lgtm
/approve
thanks :100:

PR #2604, opened 2022-11-14
feat(service catalog): Add service account with role for prometheus plugin

mimotej:
Adds SA, Clusterrole and binding for prometheus plugin that this token to access rbac-query-proxy of MCO. More information can be found [here](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.3/html-single/observability/index#external-metric-query)
Part of https://github.com/operate-first/service-catalog/issues/16
Unblocks https://github.com/operate-first/service-catalog/pull/189

### DIFF
--- a/cluster-scope/base/core/serviceaccounts/service-catalog-prometheus-plugin/kustomization.yaml
+++ b/cluster-scope/base/core/serviceaccounts/service-catalog-prometheus-plugin/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- serviceaccount.yaml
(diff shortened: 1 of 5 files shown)


cluster-scope/base/rbac.authorization.k8s.io/clusterroles/service-catalog-prometheus-plugin/clusterrole.yaml
```diff
@@ -0,0 +1,11 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: service-catalog-prometheus-plugin
```
tumido, line 4: Why do we need a new cluster role here? Isn't `cluster-monitoring-view` enough?

tumido (changes requested):

cluster-scope/base/rbac.authorization.k8s.io/clusterroles/service-catalog-prometheus-plugin/clusterrole.yaml
mimotej: Didn't notice that it exists thanks :+1:. Should be changed now.

tumido (approved):
/lgtm
/approve

PR #2611, opened 2022-11-15
allow robozome-op1st for prow trigger plugin

durandom:
Allow robozome-op1st app to trigger prow tests.

already discussed at
https://github.com/thoth-station/support/issues/122

@tumido you mentioned a prow upgrade, but it looks like we're pulling always the master images and the dashboard reports version v20220812-9414716697

so I guess this would work?


tumido (approved):
I guess... We never know unless we try it. 🙂

/lgtm
/approve

PR #2614, opened 2022-11-15
Update scripts to have environment and cluster in one

mimotej:
Related to: https://github.com/operate-first/support/pull/1144
This PR adjusts scripts to work with a new change in support templates


scripts/payload_examples/offboarding.yaml
```diff
@@ -1,7 +1,5 @@
 cluster:
-  - Smaug
-environment:
-  - moc
+  - moc/Smaug
```
tumido, line 2: why don't we lowercase it as `moc/smaug`? Or write it as `MOC/Smaug`. `moc/Smaug` looks like a weird hybrid to me :laughing:

mimotej: we can go with the first one :+1:. I have tried to keep lower and upper case as it was before that, but this will be better :).

tumido (approved):
/lgtm
/approve

PR #2615, opened 2022-11-15
Update Pulp CR

fao89:
Signed-off-by: Fabricio Aguiar <fabricio.aguiar@gmail.com>


fao89:
@tumido @harshad16 could you please take a look?

fao89:
We've rewritten the operator in go,
you can learn more about the CRs here:
https://docs.pulpproject.org/pulp_operator/pulp/

harshad16 (commented):
Thank for the updates.

/ok-to-test

pulp/overlays/moc/smaug/opf-pulp-route.yaml
```diff
@@ -1,15 +0,0 @@
----
```
fao89, line 1: we don't need the pulp-web anymore, the operator will build the routes automatically

harshad16: Will the certs be pulled?

fao89: good point, I'm not sure, @git-hyagi do you know?

harshad16: ack, shall we keep the route till we get that pr in.
in meantime, we can get the rest of the changes w.r.t to the CRs into cluster
it would balance the current deployment.

fao89: we don't have a single route now, the operator creates multiple roles for different paths, so even if we keep this file it won't work

harshad16 (approved):
/lgtm
/approve

Thanks for the updates.
Lets get this in.

PR #2618, opened 2022-11-16
Increase WildFly quota

kabir:
Also fixes name of the service

Fixes https://github.com/operate-first/support/issues/1135


durandom:
@kabir looks good, except for the `.idea` files :)

PR #2619, opened 2022-11-16
fix: Increase ghproxy resources to avoid OOM kills

tumido:
ghproxy is being OOMKilled and is raising TLS handshake errors/timeouts.

OOM kills should be solved by increasing memory request/limit from the original really restrictive `64Mi` (upstream [has none](https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/ghproxy.yaml)).

I think TLS handshakes errors can be caused by delays due to really low CPU limit - 2 milicores (upstream [has no limit](https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/ghproxy.yaml)).

### DIFF
--- a/prow/overlays/smaug/ghproxy.yaml
+++ b/prow/overlays/smaug/ghproxy.yaml
@@ -61,11 +61,11 @@ spec:
               mountPath: /cache
           resources:
             requests:
-              memory: "64Mi"
-              cpu: "2m"
+              memory: "128Mi"
+              cpu: "10m"
             limits:
-              memory: "64Mi"
-              cpu: "2m"
+              memory: "256Mi"
+              cpu: "50m"
       volumes:
         - name: cache
           persistentVolumeClaim:


prow/overlays/smaug/ghproxy.yaml
```diff
@@ -61,11 +61,11 @@ spec:
...
           resources:
             requests:
-              memory: "64Mi"
-              cpu: "2m"
+              memory: "128Mi"
+              cpu: "10m"
             limits:
-              memory: "64Mi"
-              cpu: "2m"
+              memory: "256Mi"
```
codificat, line 67: This switches from guaranteed to burstable QoS. Do we want that? Or should we keep limits == requests so that the pod doesn't get unexpectedly killed?

codificat (commented):
/lgtm

I have a comment/question about the change, but I'm good with any answer (just a clarification)

prow/overlays/smaug/ghproxy.yaml
tumido: I agree with you, however I'd say let's keep it this way for now and see the actual consumption. Once we know it, we can match limits == requests afterwards to achieve guaranteed QoS. WDYT?

codificat: Makes sense. :shipit:

PR #2624, opened 2022-11-16
Install RH SSO operator into `apex` namespace

tumido:
This partially reverts commit 006a2082e0f54867f80e7a9a9960e6beaf02e86c.

Resolves: https://github.com/operate-first/support/issues/1145


durandom:
/hold

LGTM, but if somebody needs another keycloak instance, they would need to create another subscription, which might lead to multiple versions of operators. You might use the bundle approach and set the namespace in the bundle. Up to you. This is also good

durandom:
/lgtm
/approve

if you want it merged, remove the hold

PR #2625, opened 2022-11-16
fix: Match ghproxy resource request and limit

tumido:
After #2619  it seems `ghproxy` has stabilized. Utilization is below request so let's match limits to request.

Memory utilization averages around `50Mi` and CPU around `2m`. I think `128Mi` and `10m` (matching the limit to request values) is quite generous with a large enough buffer.


codificat:
That's great, thanks!

/lgtm
/approve

PR #2627, opened 2022-11-16
Move the OSC cl1 staging ODH deployment to the meteor namespace

codificat:
The Custom Notebook Image functionality at this point needs the meteor operator and the ODH dashboard in the same namespace.

This PR moves the staging ODH deployment to the aicoe-meteor namespace.

### DIFF
--- a/kfdefs/overlays/osc/osc-cl1/cnbi/kfdef.yaml
+++ b/kfdefs/overlays/osc/osc-cl1/cnbi/kfdef.yaml
@@ -9,10 +9,43 @@ spec:
   applications:
     - kustomizeConfig:
         repoRef:
-          name: manifests
+          name: meteor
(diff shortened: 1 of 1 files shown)


harshad16 (changes requested):
i would suggest moving the content kfdef to the kfdef of the cnbi 
instead of doing this.

codificat:
> i would suggest moving the content kfdef to the kfdef of the cnbi instead of doing this.

Makes sense. Updated.

Note: I did not add the profiles configmap from the jupyterhub-stage namespace into the cnbi namespace because the current ODH versions work differently in configuring this (`notebookSizes` in the dashboard config)

harshad16 (approved):
/unhold
/lgtm
/approve 

:+1:

PR #2637, opened 2022-11-18
make crunchy postgres PGO available cluster wide

durandom:
@tumido @SamoKopecky I need the crunchy operator in another namespace and I thought this would reuse the subscription

- ~~rename service-catalog subscription to crunchy-postgres~~
- ~~make crunchy-postgres available in b4mad-racing~~
- remove service-catalog operator group and bundle
- install PGO into openshift-operators


tumido (approved):
/lgtm
/approve

PR #2651, opened 2022-11-23
Add ryanaslett (Linux Foundation) as cluster-admin

ryanaslett:
To cluster 2


tumido (approved):
@ryanaslett can you please add yourself to the org as well?

Just add your name to the list here: https://github.com/operate-first/common/blob/main/github-config.yaml#L18

/lgtm
/approve

PR #2657, opened 2022-11-25
feat: Add pulp to service catalog

fao89:
Signed-off-by: Fabricio Aguiar <fabricio.aguiar@gmail.com>

### DIFF
--- a/service-catalog/apis/pulp.yaml
+++ b/service-catalog/apis/pulp.yaml
@@ -1,0 +1,12 @@
+apiVersion: backstage.io/v1alpha1
+kind: API
+metadata:
+  name: pulp
+  description: The pulp api of our operate first instance
+spec:
+  type: openapi
+  lifecycle: production
+  owner: group:pulp
+  system: pulp
+  definition:
+    $text: https://docs.pulpproject.org/pulpcore/api.json
--- a/service-catalog/components/pulp-operator.yaml
+++ b/service-catalog/components/pulp-operator.yaml
@@ -1,0 +1,33 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: service-catalog-pulp-operator
+  title: Pulp Operator
+  description: Pulp is a platform for managing repositories of software packages and making them available to a large number of consumers
+  annotations:
+    operate-first.cloud/logo-url: https://docs.pulpproject.org/pulpcore/en/pulpproject.org_images/pulp_logo_big.png
+    backstage.io/kubernetes-label-selector: "control-plane=controller-manager,app.kubernetes.io/name=pulp-operator,olm.owner.namespace=opf-pulp"
+    argocd/app-name: pulp-smaug
+    github.com/project-slug: operate-first/apps
+  links:
+    - url: https://github.com/pulp/pulp-operator/issues/new
+      title: Support path
+      icon: user
+    - url: https://docs.pulpproject.org
+      title: Pulpcore documentation
+      icon: web
+    - url: https://pulp.operate-first.cloud/pulp/api/v3/
+      title: Instance of Pulp
+      icon: web
+    - url: https://github.com/operate-first/apps/tree/master/pulp/overlays/moc/smaug
+      title: Manifests
+      icon: web
+spec:
+  type: operator
+  lifecycle: production
+  owner: group:pulp
+  system: pulp
+  dependsOn:
+    - resource:smaug
+  providesApis:
+    - pulp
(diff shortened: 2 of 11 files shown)


harshad16 (commented):
/ok-to-test

service-catalog/apis/pulp.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: backstage.io/v1alpha1
+kind: API
+metadata:
+  name: pulp
+  description: The pulp api of our operate first instance
+spec:
+  type: openapi
+  lifecycle: production
+  owner: group:operate-first
```
tumido, line 9: @fao89 while this is not an issue, maybe we can consider creating a new user group for the Pulp maintainers? It may be different point of contact than the generic Operate First SRE group. :slightly_smiling_face: (I'm fine with leaving this as is, just noting that for your considerations)

service-catalog/components/pulp.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - url: https://docs.pulpproject.org
+      title: Pulpcore documentation
+      icon: web
+    - url: https://pulp.operate-first.cloud/pulp/api/v3/
+      title: Instance of Pulp
+      icon: web
+    - url: https://github.com/operate-first/apps/tree/master/pulp/overlays/moc/smaug
+      title: Manifests
+      icon: web
+spec:
```
tumido, line 26: This `.spec` should feature additional relation to the API defined above, like this:

https://github.com/operate-first/apps/blob/d3127104ae4f6266c30282b598064c221ee25bb6/service-catalog/components/backstage.yaml#L34-L35

tumido (commented):
Looks great! :slightly_smiling_face: 

I think you can extend this with a Pulp resource - the pulp "cluster" provided by the operator, similarly as we do with Postgres operator (defined as a component, like you did with the pulp operator) which provides a PG cluster - modeled as a `kind: Resource` here:

https://github.com/operate-first/apps/blob/master/service-catalog/resources/service-catalog-pg-cluster.yaml

This is great! Thank you for joining Backstage! :+1: :100:

service-catalog/apis/pulp.yaml
fao89: how can we create a new group?

fao89:
> Looks great! slightly_smiling_face
> 
> I think you can extend this with a Pulp resource - the pulp "cluster" provided by the operator, similarly as we do with Postgres operator (defined as a component, like you did with the pulp operator) which provides a PG cluster - modeled as a `kind: Resource` here:
> 
> https://github.com/operate-first/apps/blob/master/service-catalog/resources/service-catalog-pg-cluster.yaml
> 
> This is great! Thank you for joining Backstage! +1 100

I'm not sure if I understood well, but I added one more yaml

service-catalog/components/pulp-operator.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: service-catalog-pulp-operator
+  title: Pulp Operator
+  description: Pulp is a platform for managing repositories of software packages and making them available to a large number of consumers
+  annotations:
+    operate-first.cloud/logo-url: https://docs.pulpproject.org/pulpcore/en/pulpproject.org_images/pulp_logo_big.png
+    backstage.io/kubernetes-label-selector: "control-plane=controller-manager"
```
tumido, line 9: This would show all pods matching this label in the cluster. I think you want to limit this to the Pulp operator only. I think following selector should work:

```suggestion
    backstage.io/kubernetes-label-selector: "control-plane=controller-manager,app.kubernetes.io/name=pulp-operator,olm.owner.namespace=opf-pulp"
```

service-catalog/components/pulp.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: service-catalog-pulp
+  title: Pulp Cluster
+  description: Pulp is a platform for managing repositories of software packages and making them available to a large number of consumers
+  annotations:
+    operate-first.cloud/logo-url: https://docs.pulpproject.org/pulpcore/en/pulpproject.org_images/pulp_logo_big.png
+    backstage.io/kubernetes-label-selector: "control-plane=controller-manager"
```
tumido, line 9: Same as above, you want to match all the resources created by the operator here:

```suggestion
    backstage.io/kubernetes-label-selector: "app.kubernetes.io/part-of=pulp"
```

tumido (commented):
The only thing I'd change is the pod selector in Components. Otherwise LGTM.

tumido (approved):
/lgtm
/approve

PR #2662, opened 2022-11-29
Adding route TLS secret to pulp

fao89:
trying ingress to see if the cert works

### DIFF
--- a/pulp/overlays/moc/smaug/opf-pulp.yaml
+++ b/pulp/overlays/moc/smaug/opf-pulp.yaml
@@ -63,8 +63,7 @@ spec:
       - /tmp
     allowed_import_paths:
       - /tmp
-  route_annotations:
-    kubernetes.io/tls-acme: "true"
+  route_tls_secret: pulp-letsencrypt
   route_tls_termination_mechanism: Edge
   route_host: pulp.operate-first.cloud
   storage_type: File


pulp/overlays/moc/smaug/opf-pulp.yaml
```diff
@@ -63,10 +64,9 @@ spec:
       - /tmp
     allowed_import_paths:
       - /tmp
-  route_annotations:
+  ingress_annotations:
     kubernetes.io/tls-acme: "true"
```
fao89, line 68: do I need to do another thing to enable the cert?

harshad16: I think it would be enough.

harshad16:
will this create multiple ingress sets for each path with the same host.(same as the routes) ?

harshad16 (approved):
/lgtm
/approve

Lets give this a try to fix the certs issue.

pulp/overlays/moc/smaug/opf-pulp.yaml
tumido: What do you want to achieve with this annotation? To generate a cert? I don't think that can work... That would assume the `acme-operator` can act upon Ingres.. but you want cert manager to solve this.. since you need a single cert, the same one, shared for all the routes.

harshad16: Correct, i guess that why the idea was to use ingress instead of multiple routes, to have one ingress with all the paths, as the host is the same.
and acme seems to be able to handle ingress : https://github.com/tnozicka/openshift-acme/blob/master/docs/design/architecture.adoc#managed-objects

fao89:
> This will create ingress sets for all the paths, right?

yes, one ingress with all paths

tumido:
So.. let me explain this whole thing, I'll start with the maybe very obvious stuff, so feel free to ignore it if it's too boring and skip ahead. :slightly_smiling_face: 

Exposing things via Routes as you did previously created a `Route` resource per each path. This works fine until you want to secure this via certs. We have a [`acme-operator`](https://github.com/tnozicka/openshift-acme) deployed on our clusters which can help with cert provisioning. However there are few catches of this operator:

1. It acts per each `Route` object with `kubernetes.io/tls-acme: "true"` annotation. That means the operator requests a new certificate per each `Route` object, no matter if it's targeting the same host or not. In this case, when Pulp generates many `Route`s, `acme-operator` tries to order a new certificate for each. This results in us hitting [rate limits on Let's Encrypt side](https://letsencrypt.org/docs/rate-limits/). The limit for the same domain is **5 per week**, which is essentially blocking us from properly resolving this for few more days anyways. Since all the `Route`s generated by Pulp use the same hostname and differ in path, it can also result in interesting troubles when endpoint `/a` would use different cert than endpoint `/b`.
2. `acme-operator` can act upon `Route` object only. There's no support for Ingress in the controller (despite outlining this in their arch docs). Their readme states it may support it eventually, but there's no active development of this controller. So, adding the  `kubernetes.io/tls-acme: "true"` annotation would not result in any certificates being requested.

However, we also provide one additional and different method for requesting a certificate - [`cert-manager`](https://cert-manager.io/). This operator is more kubernetes native than the openshift specific `acme-operator` and takes a more sensible approach to certificates. It can help you generate a `Secret` with the certificate using the following pattern:

1. Since you don't need wildcard certs, we can use the ACME http01 challenge, that simplifies things for us.
2. We either create a cluster-scoped `ClusterIssuer` for everybody to consume or use just a namespace scoped `Issuer`. This resource would look like this:

   ```
   apiVersion: cert-manager.io/v1
   kind: Issuer
   metadata:
     name: letsencrypt
   spec:
     acme:
       email: ops-team@operate-first.cloud
       privateKeySecretRef:
         name: letsencrypt-key
       server: 'https://acme-v02.api.letsencrypt.org/directory'
     solvers:
       - http01:
           ingress:
             serviceType: ClusterIP
   ```

3. Then we can simply request a `Certificate` from Let's Encrypt like this:

   ```
   apiVersion: cert-manager.io/v1
   kind: Certificate
   metadata:
     name: pulp
   spec:
     dnsNames:
       - pulp.operate-first.cloud
     issuerRef:
       name: letsencrypt
     secretName: pulp-letsencrypt   
   ```

4. This should result in `pulp-letsencrypt` Secret being created (or whatever `.spec.secretName` is set in the `Certificate` resource.
5. Then you should be able to mount this Secret into your `Ingress`:
   ```
   apiVersion: networking.k8s.io/v1
   kind: Ingress
   metadata:
     name: pulp
   spec:
     ...
     tls:
     - hosts:
       - pulp.operate-first.cloud
       secretName: pulp-letsencrypt
   ``` 


However as I'm trying this our right now, it seems the challenge pods are not spawning because of:
https://cert-manager.io/docs/release-notes/release-notes-1.10/#on-openshift-the-cert-manager-pods-may-fail-until-you-modify-security-context-constraints

Therefore we may need to use dns01 challenge instead. I'll try setting that up tomorrow (requires a service account from DNS provider). In that scenario you'd still need to do the steps 3.-5. outlined above (with slight modifications).


There is [some development on the OCP side going on](https://docs.openshift.com/container-platform/4.10/security/cert_manager_operator/cert-manager-operator-release-notes.html) in order to better integrate Cert Manamager. We should look into that later... It may enable us to use ingress shims (provision certs directly from the `Ingress` resource via annotations. This doesn't work for me currently).

harshad16:
Thank you @tumido , for taking the time and providing a thorough explanation.

/remove-approve
/remove-lgtm

fao89:
@tumido thanks for the explanation!
I'll modify this PR to mount the letsencrypt secret into the routes then
https://github.com/pulp/pulp-operator/pull/801

PR #2671, opened 2022-12-02
Add Pep to ArgoCD admins

codificat:
This PR is to add myself to the ArgoCD admins group. From time to time I keep finding that I can't access X or Y resource in ArgoCD, and this is an attempt at fixing that. Let me know if this is overkill.

Also taking the opportunity to remove Anand, who I believe is not active anymore.

### DIFF
--- a/cluster-scope/base/user.openshift.io/groups/argocd-admins/group.yaml
+++ b/cluster-scope/base/user.openshift.io/groups/argocd-admins/group.yaml
@@ -3,8 +3,8 @@ kind: Group
 metadata:
   name: argocd-admins
 users:
-  - 4n4nd
   - anishasthana
+  - codificat
   - dystewart
   - durandom
   - gregory-pereira


tumido (commented):
While I'm not against this change, can you please let me know which projects you'd like to have access to and can't? I think we should adopt much broader read access - public read should become the default.

Do you need to be able to perform manual changes to those apps?

codificat:
> Can you please let me know which projects you'd like to have access to and can't? 

The trigger for the request was the ODH applications in OSC's cluster 1, which we use to contribute to the ODH custom notebook image functionality.

One example from earlier today: after https://github.com/operate-first/odh-manifests/pull/52 merged I was expecting the deployment of the dashboard in the aicoe-meteor namespace to get updated; this did not happen.

I ended up manually updating it in the cluster (I do have access to the cluster), but I can not even see the related resources in ArgoCD (https://github.com/operate-first/apps/tree/master/kfdefs/overlays/osc/osc-cl1).

> I think we should adopt much broader read access - public read should become the default.

That would be nice :smiley: 

> Do you need to be able to perform manual changes to those apps?

Not really manual changes, as in theory these should all go via the `apps` repo... but eventually a manual sync might come handy. In the example below, the repos were up to date but the changes were not reflected in the cluster; not sure what was missing in the argocd side (as I can't see it) or what actions there might have been needed.

tumido:
@codificat FFR: we're doing some policy changes around ArgoCD now (see @durandom's PR). I think all you want to achieve should be possible already without being an admin (without write access). If there are apps not synced, I think we should rather turn on autosync declaratively in the app manifest instead of manual interventions... That being said, I'm approving this PR.

tumido (approved):
/lgtm
/approve

PR #2675, opened 2022-12-06
feat: Create SCC with seccomp runtime/default

tumido:
This change will allow cert-manager to use http01 challenge. This SCC is equivalent to `runtime-v2` available since OCP 4.11.

https://docs.openshift.com/container-platform/4.10/security/seccomp-profiles.html#configuring-default-seccomp-profile_configuring-seccomp-profiles
https://docs.openshift.com/container-platform/4.11/release_notes/ocp-4-11-release-notes.html#ocp-4-11-auth-pod-security-admission
https://cert-manager.io/docs/release-notes/release-notes-1.10/#on-openshift-the-cert-manager-pods-may-fail-until-you-modify-security-context-constraints


durandom (commented):
/approve
/lgtm

PR #2676, opened 2022-12-06
feat: Create certificate for pulp.operate-first.cloud

tumido:
Depends on: #2675
Related to: #2662

This will create a `pulp-letsencrypt` Secret in the `opf-pulp` namespace. This secret will contain `tls.crt` and `tls.key` keys.

/hold (until requirements are merged)
/cc @fao89 @harshad16


fao89 (approved):
Thank you!

PR #2678, opened 2022-12-07
feat: Add secret store to apex

tumido:
Resolves: https://github.com/operate-first/support/issues/1154


tumido:
/cc @durandom @mimotej

please review

PR #2679, opened 2022-12-07
Rename pulp cert secret

fao89:


pulp/overlays/moc/smaug/issuer.yaml
```diff
@@ -6,7 +6,7 @@ spec:
   acme:
     email: ops-team@operate-first.cloud
     privateKeySecretRef:
-      name: letsencrypt-key
+      name: pulp-letsencrypt
```
tumido, line 9: This is the private key for the ACME account (see docs [here](https://cert-manager.io/docs/configuration/acme/#creating-a-basic-acme-issuer). I don't think you want the change this. The actual cert secret is defined in the `Certificate` resource here:

https://github.com/fao89/apps/blob/53d86f722b119e4aaf70e4d22a318878584e14d3/pulp/overlays/moc/smaug/certificate.yaml#L10

tumido (changes requested):

PR #2680, opened 2022-12-08
Create kuberay operator deployment

saadullah01:
We want to create kuberay operator deployment. We have created an issue here: https://github.com/operate-first/support/issues/1149

### DIFF
--- a/kfdefs/overlays/moc/smaug/ray-bu/manager.yaml
+++ b/kfdefs/overlays/moc/smaug/ray-bu/manager.yaml
@@ -26,8 +26,6 @@ spec:
       containers:
       - command:
         - /manager
-#        args:
-#        - --enable-leader-election
         image: kuberay/operator
         ports:
         - name: http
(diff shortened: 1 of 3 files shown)


tumido (changes requested):
Hey! Is there any reason why these manifests are required to live within the `kfdefs` top-level folder and in this repository at all? Is this expected to be owned by the Operate First Ops team? I think these manifests can be freely owned by the Ray BU team in their own repo, or am I missing something?

/cc @erikerlandson @durandom

kfdefs/overlays/moc/smaug/ray-bu/manager.yaml
```diff
@@ -0,0 +1,60 @@
...
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: kuberay
+        app.kubernetes.io/component: kuberay-operator
+    spec:
+      securityContext:
+        runAsNonRoot: true
+      serviceAccountName: kuberay-operator
```
tumido, line 25: Where is this Service account defined? I think you're missing a manifest in the PR.😉

tumido (changes requested):

goern:
closing this, ODH upstream seems to have done a lot of work and moved on..
thanks @saadullah01 
/close

PR #2683, opened 2022-12-12
docs: GitOps JupyterBook -> Backstage [part 1]

tumido:
Part of: https://github.com/operate-first/service-catalog/issues/196

`mkdocs` doesn't allow building documentation from the same directory as `mkdocs.yaml` is located, therefore all the files had to be moved into `docs/` folder.


durandom (commented):
/approve
/lgtm

PR #2687, opened 2022-12-13
make argocd free again

durandom:
- readonly role
- allow anonymous access

As per https://argo-cd.readthedocs.io/en/stable/operator-manual/rbac/ the role `readonly-public` does not exist. I think we want the policy `readonly` 
And then we want anonymous users to see whats going on.


tumido (approved):
If we want to remove `readonly-public` please remove also:
https://github.com/operate-first/apps/pull/2687/files#diff-3be23888dff2f10b0a59d94e86efd7b0acf16da098e6217055f97d34649aa336R19

https://github.com/operate-first/apps/tree/master/cluster-scope/base/user.openshift.io/groups/argocd-readonly

https://github.com/operate-first/apps/blob/master/cluster-scope/overlays/prod/common/kustomization.yaml#L18

tumido (changes requested):
meant to hit the other button :facepalm:

tumido (approved):
/lgtm
/approve

PR #2688, opened 2022-12-13
fix: Add missing aggregate role to ray bundle

tumido:
SSIA

/hold

waiting for confirmation from @erikerlandson in this thread: https://operatefirst.slack.com/archives/C01RMPVUUK1/p1670946452014479?thread_ts=1669660577.038729&cid=C01RMPVUUK1


goern:
@tumido is this still a thing? @erikerlandson @codificat something you need?

erikerlandson:
IIUC, the clusters it would apply to are being decommissioned. Also, codeflare operators and sdk is likely to be the way to interact with ray going forward, so this can probably be closed.

PR #2689, opened 2022-12-14
let dex map any group to argocd

durandom:


argocd/overlays/moc-infra/configs/argo_cm/dex.config
```diff
@@ -7,21 +7,4 @@ connectors:
...
-        - kepler-admins
-        - lab-cicd
-        - octo-ushift-dev
-        - operate-first
-        - osc-admins
-        - rekor
-        - thoth
-        - tremor-demo
-        - workshops
+      insecureCA: true
```
tumido, line 10: Please fix the EOF here.

tumido (changes requested):
one small nit, otherwise LGTM

argocd/overlays/moc-infra/configs/argo_cm/dex.config
durandom: meh, too reliant on my precommit checks

PR #2691, opened 2022-12-15
Adding Kubescape workflow

dwertent:
[Kubescape](https://github.com/kubescape/kubescape) is a tool for detecting Kubernetes manifests misconfigurations based on security, best practices, and compliance.

In this PR, I added [Kubescape's GitHub actions](https://github.com/kubescape/github-action) so you will see in the security tab the misconfigurations detected in the submitted YAML files.

**This will not break/block the current CI!** 

If the maintainers wish to block critical severities from being pushed, they will need to set the parameter as demonstrated [here](https://github.com/kubescape/github-action#fail-kubescape-scanning-based-on-maximum-severity-of-a-failed-control).
 
Here is a screenshot of how the security tab will look like:
image

### DIFF
--- a/.github/workflows/kubescape.yaml
+++ b/.github/workflows/kubescape.yaml
@@ -1,0 +1,16 @@
+name: Kubescape scanning for misconfigurations
+on: [push, pull_request]
+jobs:
+  kubescape:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: kubescape/github-action@main
+      continue-on-error: true
+      with:
+        format: sarif
+        outputFile: results.sarif
+    - name: Upload Kubescape scan results to Github Code Scanning
+      uses: github/codeql-action/upload-sarif@v2
+      with:
+        sarif_file: results.sarif


tumido:
Hm.. seeing those alerts I don't know this is helpful much in the current configuration... All I went through seems to be false positives. We store manifest partials in this repo and compose them via `kustomize`. Therefore things like *Don't use default namespace* really doesn't make sense if the namespace is set via the `kustomization.yaml`... etc. How can we configure what kind of alerts we want to see? 

https://github.com/operate-first/apps/pull/2691/checks?check_run_id=10116787811 

Other family of false positives can be attributed to OpenShift vs Kubernetes - most of these errors won't get accepted or deploy on OpenShift unless permitted by a special SCC for example.

durandom:
@dwertent does kubescape support silencing checks/warnings and how does it handle kustomize?

dwertent:
@tumido Thank you for your detailed reply, I will add here a custom framework so we do not run tests that are not relevant to YAML files.
@durandom kustomize should be supported, the GitHub actions do not support ignoring checks, I will see what I can do about this :)

PR #2694, opened 2022-12-15
argocd: Fix Apex Destinations

dave-tucker:
Argo app fails with "application destination
{https://api.smaug.na.operate-first.cloud:6443 apex} is not permitted in project 'apex'"
Hoping this should fix it.

Signed-off-by: Dave Tucker <dave@dtucker.co.uk>

### DIFF
--- a/argocd/overlays/moc-infra/projects/apex.yaml
+++ b/argocd/overlays/moc-infra/projects/apex.yaml
@@ -7,9 +7,9 @@ metadata:
 spec:
   destinations:
     - namespace: apex
-      name: smaug
+      server: "https://api.smaug.na.operate-first.cloud:6443"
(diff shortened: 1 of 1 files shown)


tumido (approved):
It sometimes does that.. I think it's fixed in some ArgoCD release lately, we have migration/upgrade in the backlog... I think this should fix it for now, you are correct.

/lgtm
/approve

dave-tucker:
Sorry - re-pushed. Had to format my YAML so the pre-commit hook wouldn't fail.
Or at least I guess that's why it failed - build logs were empty :(

tumido (approved):
No worries :slightly_smiling_face: 

/lgtm
/approve

PR #2696, opened 2022-12-15
argocd: Allow apex group access to apex project

dave-tucker:
This should allow us to click sync through the web UI and use the argocd cli. Some of the permissions, `projects get` and `logs get` might already be granted to readonly users, but I'm not sure if these propagate to logged in users so adding them to be safe.

Signed-off-by: Dave Tucker <dave@dtucker.co.uk>

### DIFF
--- a/argocd/overlays/moc-infra/configs/argo_rbac_cm/policy.csv
+++ b/argocd/overlays/moc-infra/configs/argo_rbac_cm/policy.csv
@@ -1,2 +1,10 @@
 # Give Openshift group (argocd-admins) the argocd admin role with unrestricted argocd access
 g, argocd-admins, role:admin
+
+# Allow Apex users to administer the Apex Project
+p, role:apex-admins, applications, get, apex/*, allow
+p, role:apex-admins, applications, sync, apex/*, allow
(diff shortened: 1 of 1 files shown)


dave-tucker:
This assumes that this PR will also allow us to login... Attempting that now results in:
Screenshot 2022-12-15 at 20 57 25

tumido:
@dave-tucker The login error is caused by you not being a member of the `operate-first` org. Unfortunately DeX doesn't allow us to surface that error any better. Can you please add yourself to the list of members here:

https://github.com/operate-first/common/blob/d03a6778e5fe455cca7790a7dbdfe5aa302235b7/github-config.yaml#L18

That should solve the auth error I think.

The permissions should be already set for you via:

https://github.com/operate-first/apps/blob/946bc7234c068f1e5c48f0e1b4dbd9b4d5e0bb7b/argocd/overlays/moc-infra/projects/apex.yaml#L17-L23

If that doesn't work once you're able to log in, we may need to try this PR.

I think we had some ArgoCD manifest syncing troubles when I've looked right now (argo wasn't able to resolve a remote base) but it should be all synced now.

PR #2699, opened 2022-12-21
Added machinesets to osc cl1 to sync with cl2

redmikhail:
Part of the process of bringing osc-cl1 cluster to the level of control by operate-first processes as for ocs-cl2. Adding previously manually managed machinesets to the repo. Also adjustments to address https://github.com/os-climate/os_c_data_commons/issues/246

### DIFF
--- a/cluster-scope/overlays/prod/osc/osc-cl1/kustomization.yaml
+++ b/cluster-scope/overlays/prod/osc/osc-cl1/kustomization.yaml
@@ -72,5 +72,8 @@ resources:
 - configmaps/service-catalog-k8s-plugin.yaml
 - externalsecrets
 - ingresscontroller
+- machinesets
+- nodefeaturediscovery
 - machineconfigs/kepler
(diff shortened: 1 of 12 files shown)


redmikhail:
@MightyNerdEric or @ryanaslett can you please review and provide lgtm and approval if everything correct

ryanaslett:
The machinesets I checked seem to match exactly with whats defined in the cluster.   I think the only questions I have are 

1. When we define something in kustomize that already exists in the cluster, with all of the same names and data, I assume kustomize figures out that its already there, and doesnt make any changes to the cluster itself? i..e do we run any risk of the cluster removing and rebuilding all of these machines if we do this? 
2. The (https://github.com/os-climate/os_c_data_commons/issues/246) is sort of confusing to me.. looks like everything is already defined as g4dn.2xlarge .. did anything change there? 

Otherwise this all looks good to me.

PR #2700, opened 2022-12-22
Update kustomization.yaml

rbadagandi:
Adding Kruize to be part of the group for ArgoCD deployment

PR #2701, opened 2022-12-22
Added Kruize.yaml and Kustomization.yaml for argocd integration

rbadagandi:
Added Kruize.yaml and Kustomization.yaml for argocd integration

### DIFF
--- a/argocd/overlays/moc-infra/projects/kruize.yaml
+++ b/argocd/overlays/moc-infra/projects/kruize.yaml
@@ -6,7 +6,7 @@ metadata:
     project-template: global
 spec:
   destinations:
-    - namespace: 'openshift-tuning-*'
+    - namespace: 'openshift-tuning'
       name: 'smaug'
   sourceRepos:
     - '*'


argocd/overlays/moc-infra/applications/envs/moc/smaug/kruize/kruize.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: hpo
+spec:
+  destination:
+    name: smaug
+    namespace: openshift-tuning
```
tumido, line 8: This namespace is not allowed in the `kruize` ArgoCD project, see:

https://github.com/operate-first/apps/blob/4b724a89068ca34cafac695d542516f5d4b5cfc7/argocd/overlays/moc-infra/projects/kruize.yaml#L9


(`openshift-tuning-*` doesn't match `openshift-tuning` because of the `-*`)

Please add the additional namespace to the project.

tumido (changes requested):
Can you please add the change from https://github.com/operate-first/apps/pull/2700 in here? Either PR does nothing on its own. These changes need to go together. (That's the reason why checks are failing on the other PR).

PR #2705, opened 2023-01-09
fix: Fix issue with rolebinding being listed as component

mimotej:
Should resolve issue in https://github.com/operate-first/apps/pull/2697


tumido (approved):
/lgtm
/approve

PR #2707, opened 2023-01-10
chore: Update ocm API definition

SamoKopecky:
Change API definition for the OCM plugin, see https://github.com/janus-idp/backstage-plugins/pull/67.


tumido (approved):
/lgtm
/approve

PR #2708, opened 2023-01-11
chore(catalog): Update ClusterRole for OCM plugin

SamoKopecky:
Update `ClusterRole` definition for service catalog OCM plugin, see https://github.com/janus-idp/backstage-plugins/pull/67.


tumido (approved):
/lgtm
/approve

PR #2709, opened 2023-01-11
fix(smaug): Relocate image registry to PVC

tumido:
Related to: https://github.com/operate-first/support/issues/1188

Re-using Jerry's manifests for Smaug to migrate from OBC to PVC.


tumido:
self merging

PR #2710, opened 2023-01-11
fix(smaug): Relocate image registry to PVC part 2

tumido:
Further refinement of #2709 

Current PVC fails to bind with multiple replicas, adjusting according to:

https://docs.openshift.com/container-platform/4.8/registry/configuring_registry_storage/configuring-registry-storage-baremetal.html#installation-registry-storage-block-recreate-rollout-bare-metal_configuring-registry-storage-baremetal


tumido:
self merging

PR #2711, opened 2023-01-11
fix(smaug): Relocate image registry to PVC part 3

tumido:
Further refinement of  #2709 and #2710

PVC needs to be `ReadWriteOnce` as it is on Jerry


tumido:
self merging

PR #2712, opened 2023-01-12
moving prow storage to ionos s3 bucket

Gregory-Pereira:
A new secret has been created with this vault path. I used Jerry as the cluster on which to host the OBCs / S3 buckets, however before we ran into issues due to the prow service being in NA on Smaug, and the ionos s3 bucket living in EMEA, and the long connection time causing data transfer to drop. If this proves to still be an issue, I will opt to use Infra. I started with Jerry because on infra there is no backing storage class except for noobbaa, which caused the storage issues in first place.

/cc @tumido 
/cc @durandom 


tumido:
@Gregory-Pereira why the path change? can't you just update the original vault secret?

PR #2722, opened 2023-01-25
Upgrade openmetadata

MightyNerdEric:


MightyNerdEric:
@erikerlandson Looking for a merge on this. The failed test comment above was based on an old version of this that needed to be rebased from master. After rebasing and force-pushing my feature branch, the test passed as expected.

tumido (approved):
/assign @mightynerderic

/approve

Erik can you please `lgtm` this once you take a look?

tumido:
> /assign @erikerlandson

Yeah, sorry @MightyNerdEric that's what I've meant.

openmetadata/base/openmetadata-dependencies/deployments/airflow-db-migrations.yaml
```diff
@@ -70,6 +73,9 @@ spec:
         - name: db-migrations
           image: quay.io/os-climate/ingestion:0.13.1
           imagePullPolicy: IfNotPresent
+          securityContext:
```
redmikhail, line 76: I wonder if we can avoid explicitly setting UID/GID allowing openshift automatically assign gid. We would need to  change scc associated with airflow service account to enable it anyways.

MightyNerdEric: Ok, I've removed the securityContext changes. These may be required, but we can merge the config changes separately and then troubleshoot if we're still seeing problems with the images not being able to properly write data.

PR #2723, opened 2023-01-26
Update path to iris-notebook, and add to cl2

MightyNerdEric:
Testing and validation are positive, so we're ready to move this over to the primary os-climate accounts and add it to cl2.

Signed-off-by: Eric Ball <eball@linuxfoundation.org>


tumido (approved):
/lgtm
/approve

PR #2728, opened 2023-02-03
increased cpu limits

oindrillac:
Jupyterhub is unable to spawn up because the cpu limits are reached. Increased the resource limits and request

Screenshot 2023-02-03 at 7 00 32 PM


tumido (approved):
/lgtm
/approve

PR #2731, opened 2023-02-07
change rollout strategy

durandom:
otherwise upgrading the openshift-image-registry will not work, since our storage in jerry is only block storage


tumido (approved):
/lgtm
/approve

PR #2732, opened 2023-02-07
set hive.timestamp-precision to MICROSECONDS

erikerlandson:
The goal is to align the timestamp format between hive and iceberg, and greatly reduce friction on data ingestion due to timestamp format mismatches.
cc @MichaelTiemannOSC


MichaelTiemannOSC:
The OWNERS file should be edited to remove ryanfinn.

tumido (approved):
/lgtm
/approve

PR #2733, opened 2023-02-13
Add additionnal Janus Idp users

christophe-f:


cluster-scope/base/user.openshift.io/groups/janus-idp/group.yaml
```diff
@@ -1,8 +1,10 @@
-apiVersion: user.openshift.io/v1
+\apiVersion: user.openshift.io/v1
```
tumido, line 1: @christophe-f can you please fix this? Leading backslash would cause issues.

tumido (changes requested):

tumido (approved):
/lgtm
/approve

christophe-f:
oops. Typo fixed

Issue #1142, opened 2021-10-08
Add missing limitranges to all namespaces with quota


tumido:
If a namespace has a resource quota while it doesn't have a limit range may cause issues if a pod has no specific resource requests/limits. In such case the pod tries to default to the default limit range. And since there's no limit range, there's no value to be compared against the resource quota. As a result the pod fails to schedule.

Please check on all namespaces in `cluster-scope/base/core/namespaces/*/kustomization.yaml` and add:

```yaml
components:
...
- ../../../../components/limitranges/default
```

Only if the `components` already contain `- ../../../../components/resourcequotas/*` entry or if there's a `resourcequota.yaml` in `resources` field.

Example:

1. This resource is fine, no change is needed, since both limit range and quota are present:
https://github.com/operate-first/apps/blob/40cdeb8b3011ba053ba5a00c514148df02387d8d/cluster-scope/base/core/namespaces/acme-operator/kustomization.yaml#L2-L10
2. This resource lacks limit range and it needs to be added since there's resource quota present in `resources`:
https://github.com/operate-first/apps/blob/40cdeb8b3011ba053ba5a00c514148df02387d8d/cluster-scope/base/core/namespaces/opf-jupyterhub/kustomization.yaml#L1-L13
3. This resource is fine, since it doesn't contain resource quota, so no limit range is required
https://github.com/operate-first/apps/blob/40cdeb8b3011ba053ba5a00c514148df02387d8d/cluster-scope/base/core/namespaces/openshift-logging/kustomization.yaml#L1-L7

PR #2734, opened 2023-02-13
Adding new namespace for to OS-Climate cl2

redmikhail:
/cc @MightyNerdEric

### DIFF
--- a/cluster-scope/base/core/namespaces/aicos-osc-demo/kustomization.yaml
+++ b/cluster-scope/base/core/namespaces/aicos-osc-demo/kustomization.yaml
@@ -1,0 +1,7 @@
+kind: Kustomization
+apiVersion: kustomize.config.k8s.io/v1beta1
+namespace: aicos-osc-demo
+resources:
+    - namespace.yaml
+components:
+    - ../../../../components/project-admin-rolebindings/aicos-osc-demo
--- a/cluster-scope/base/core/namespaces/aicos-osc-demo/namespace.yaml
+++ b/cluster-scope/base/core/namespaces/aicos-osc-demo/namespace.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    openshift.io/display-name: "OS Climate: aicos-osc-demo"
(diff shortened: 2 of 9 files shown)


cluster-scope/base/core/namespaces/aicos-osc-demo/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Kustomization
+apiVersion: kustomize.config.k8s.io/v1beta1
+namespace: aicos-osc-demo
+resources:
+    - namespace.yaml
+components:
+    - ../../../../components/project-admin-rolebindings/aicos-osc-demo
```
tumido, line 7: Limit ranges and quotas are missing here. While you can have a namespace without quotas, I discourage it. Limit ranges are really good to have as well. 

See other namespaces for reference:
https://github.com/operate-first/apps/blob/master/cluster-scope/base/core/namespaces/apex/kustomization.yaml#L8-L9

redmikhail: @tumido it make sense, however in this case we want to give developers bit of flexibility and potentially apply quotas subsequently. Also there is additional factor that that they will need GPU resources as well

tumido: Ok then. If you as a cluster admin are fine with that, than sure.

Just keep in mind that once you decide to add `ResourceQuota`, you need to apply the `LimitRange` with pod defaults as well. Otherwise the quota calculations for pods without set `resources` (either request or limit for both CPU and memory) would fail and the pod would subsequently fail to schedule. Which is an interesting problem, imo.

https://github.com/operate-first/apps/issues/1142#issue-1020981783

Btw, GPUs can be specified in the quota as well (and if they are not, they are not restricted).

tumido (commented):
@MightyNerdEric please review

redmikhail:
@tumido Can you please lgtm and approve this request if it looks good to you

tumido (approved):
Sure thing 🙂

/lgtm
/approve

PR #2742, opened 2023-02-21
feat: Add k8s plugin tokens to janus-idp

SamoKopecky:
Add k8s plugin tokens to the `janus-idp` namespace since we need them for the `ocm` plugin installation. I also updated the `janus-idp` policy to allow reads to the referenced vault secret.


tumido (approved):
/lgtm
/approve

PR #2746, opened 2023-02-23
feat: Add Open Metadata to OSC-CL1

MightyNerdEric:


MightyNerdEric:
/assign @redmikhail 
Mikhail, if you could give this a look and also tag anyone else you think should check it out before merging, I'd appreciate it. Thanks!

redmikhail:
@MightyNerdEric can you try to do rebase

PR #2749, opened 2023-02-27
chore: Update physrisk-resilience resource quotas

MightyNerdEric:
This team will need cpu and ram limits similar to a "large" resource quota, but need to have storage and objectbucketclaims unlimited for now.

### DIFF
--- a/cluster-scope/base/core/namespaces/physrisk-resilience/kustomization.yaml
+++ b/cluster-scope/base/core/namespaces/physrisk-resilience/kustomization.yaml
@@ -1,9 +1,9 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-- namespace.yaml
+  - namespace.yaml
+  - resourcequota.yaml
 namespace: physrisk-resilience
 components:
-- ../../../../components/limitranges/default
-- ../../../../components/resourcequotas/small
-- ../../../../components/project-admin-rolebindings/physrisk
+  - ../../../../components/limitranges/default
+  - ../../../../components/project-admin-rolebindings/physrisk
--- a/cluster-scope/base/core/namespaces/physrisk-resilience/resourcequota.yaml
+++ b/cluster-scope/base/core/namespaces/physrisk-resilience/resourcequota.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: ResourceQuota
+metadata:
+  name: physrisk-resilience-custom
+spec:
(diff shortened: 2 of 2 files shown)


cluster-scope/base/core/namespaces/physrisk-resilience/kustomization.yaml
```diff
@@ -1,9 +1,8 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-- namespace.yaml
+  - namespace.yaml
```
tumido, line 4: The `resourcequota.yaml` file needs to be listed here, otherwise the quota won't apply.

tumido (changes requested):

cluster-scope/base/core/namespaces/physrisk-resilience/kustomization.yaml
```diff
@@ -1,9 +1,8 @@
...
 namespace: physrisk-resilience
 components:
-- ../../../../components/limitranges/default
-- ../../../../components/resourcequotas/small
-- ../../../../components/project-admin-rolebindings/physrisk
+  - ../../../../components/limitranges/default
```
redmikhail, line 7: I also think we should remove limitrange as memory/cpu for individual containers for this app probably will be higher

tumido: Our `limitranges/default` component assigns defaults only. It gets applied/assigned only if there are no `resources.limits` specified in the pod. If you assign those values in pod spec, nothing changes for you. If you have quota on a namespace you **are required** to have limitranges defaults, otherwise the pod without resource restriction fails to schedule. Removing this will cause issues, please don't do that.

https://github.com/operate-first/apps/blob/e446586e1a60c6dea17fd7b0baca85fc821d9819/cluster-scope/components/limitranges/default/limitrange.yaml#L6-L13

MightyNerdEric: @tumido Roger that.

cluster-scope/base/core/namespaces/physrisk-resilience/kustomization.yaml
MightyNerdEric: Done.

tumido (approved):
/approve

@redmikhail please lgtm :slightly_smiling_face:

PR #2752, opened 2023-03-02
feat: Add namespace for users service-catalog

mimotej:
Adds a namespace for service-catalog webterminal because by design only kubeadmin can create DevWorkspaces in `openshift-terminal` namespace. I went with small quota because each workspace uses roughly 200m so this means we can run roughly 5 terminals at one time which seems to me like good starting point. For now I have also allowed to users who are in `service-catalog` rolebinding to use this namespace.


tumido (changes requested):
Please update the `kustomization.yaml` for appropriate clusters to propagate this namespace.

tumido:
@mimotej please rebase

mimotej:
> @mimotej please rebase

@tumido done

PR #2753, opened 2023-03-02
Temporary sync change

PatAKnight:
Janus-idp is having some issues and we are wanting to temporarily disable the auto sync while we work on them.


PatAKnight:
/assign @redmikhail 
Hello Mikhail, do you mind giving this a look right quick. We are trying to solve some issues with our showcase app and looking to temporarily disable ArgoCD sync

PatAKnight:
Nevermind, we managed to get it working after all

PR #2765, opened 2023-03-16
feat: Add MachineWithoutValidNode alert

mimotej:
This PR adds a MachineWithoutValidNode alert.
The metrics it uses do not include `managed_cluster_id`
 label by default but they have `clusterID` instead. In this case `clusterID` is the ID of the cluster the alert originated in (therefore we can define `managed_cluster_id` label ourselves). For now I used following [guide](https://github.com/openshift/machine-api-operator/blob/master/docs/user/Alerts.md) If you have ideas for better guide feel free to point them out.


tumido (approved):
/lgtm
/approve
/ok-to-test

PR #2769, opened 2023-03-21
fix: Add by clause to PromQL expression custom rule

mimotej:
The custom alert rule `MachineWithoutValidNode` does not seem to display proper labels. The PromQL expression is missing aggregation operator which would add proper labels to the alert. More about aggregation can be seen [here](https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators) 
Here is the result of query without aggregation
image
Here is the result of query with aggregation
image


tumido (approved):
/lgtm
/approve

PR #2770, opened 2023-03-21
feat: Increase quota for web-terminal-service-catalog namespace

mimotej:
I found out that `small` resource quotas are not enough for running at least two webterminals at once (one devworkspace has limit 800m) therefore we cant spawn two devworkspaces at once


tumido (approved):
/lgtm
/approve

PR #2780, opened 2023-04-15
removing argocd autosync

Gregory-Pereira:
all applications (with the exception of janus-idp and dependent apps) have been removed from autosync so their deployments can be scaled down.

/cc @tumido 
/cc @cooktheryan


goern:
@Gregory-Pereira this change introduced a lot of `warning  comment not indented like content  (comments-indentation)` errors...

Gregory-Pereira:
yes the pre-commit doesn't like comments and their formatting. I can delete the commented out sync configurations if we want.

goern:
I think you just need to indent them on the correct level, https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.comments_indentation

PR #2783, opened 2023-04-16
script to descale workloads

Gregory-Pereira:
decomissioning script
/cc @cooktheryan

### DIFF
--- a/scripts/the-big-red-button.sh
+++ b/scripts/the-big-red-button.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+
+# Function to scale down a workload to zero replicas
+scale_down_workload() {
+  local namespace="$1"
+  local workload_type="$2"
+  local workload_name="$3"
+
+  echo "Scaling down $workload_type $workload_name in namespace $namespace to 0 replicas..."
+  oc scale $workload_type $workload_name --replicas=0 -n $namespace
+}
+
+# Check if 'oc' command is available
+if ! command -v oc &> /dev/null; then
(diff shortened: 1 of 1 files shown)


goern:
why dont we do this via PR that change the argocd configuration and have argocd do the job? @Gregory-Pereira

Gregory-Pereira:
@goern , already have a PR (https://github.com/operate-first/apps/pull/2780) to remove auto-sync. We wanted to approach it this way because we do not want to delete all manifests in case a team complains and OCTO approves them to spin up their workloads again, so this method removes compute while leaving the structure in place. What we do plan to remove: routes, services, deployments, statefulsets, replicasets, daemonsets, etc. What we don't plan to remove: configmaps, secrets, etc. We have a post spindown meeting to evaluate if this strategy is effective enough to reduce our costs with management.

goern:
I don't get that. The plan communicated was to decommission the smaug cluster. Is that still the plan? So cleaning up the apps repo would make most sense, as it is reflecting the desired state. Removing objects from a cluster manually and turning off Gitops feels suboptimal and leaves the gitops repo in a state that needs to be considered unclean, so restarting or deploying on a different infrastructure would involve a cleanup of the repo anyway.

cooktheryan:
@goern the strategy is hopefully one last safety net for people to realize the environment is powering down. It's a "soft" last warning before shutdown. This way if the project is something incredibly important to RH, a university or etc that it can be synced again and an immediate migration plan can be established.

goern:
:evilgrin: `revert commit` you said? :)

PR #2792, opened 2023-05-08
add sabrycito to user

sabrycito:
Hello ! I'm the new U-Reg Data Engineer intern and need to work with OS-C Trino data for my project


goern:
@cooktheryan Who is our OSC liaison to handle this? From my point of view operate fist is no longer providing support for OS Climate, is that correct?

cooktheryan:
@redmikhail can you take a look at this? I thought we had migrated most of the items over to the new repo. cc @Gregory-Pereira

